PR #157, opened 2013-06-12
Hex

jonasdn:
Hi,

This pull request could be considered as a RFC.

It implements parsing of Intel HEX files in a new file stlink-hex.c.
It adds handling of flashing the parsed hex file to stlink_fwrite_flash in stlink-common.c
and also to the stlink-gui.

It does nothing with the flash/main.c  ui. To use it from the command line you just pass the
hex file and specify the base address. Even tho that information is discarded and the stlink_fwrite_flash function uses the start_addr from the hex file. 
Could this be improved?

The branch also contains update to the gui, to make it re-read the device memory after
flashing. Should I Submit this patch later on?

Regards
Jonas

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,6 +16,7 @@ CFILES = \
 	src/stlink-common.c \
 	src/stlink-usb.c \
 	src/stlink-sg.c \
+	src/stlink-hex.c \
 	src/uglylogging.c \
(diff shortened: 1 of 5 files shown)


Nightwalker-87:
Closed in favour of #459.

PR #232, opened 2014-05-10
Calling stlink_reset() causes the target to reset after disconnecting.

mrdunk:
While trying to read/write directly from/to RAM i noticed that the STM32 would always reset after stmlink disconnects.
This is not a good thing when using stlink to access RAM on running programs.

Unfortunately writing to Flash fails if this reset does not occur.

My hardware is STM32VLDISCOVER where this change works as intended
however i am not able to verify this patch on other platforms.

### DIFF
--- a/flash/main.c
+++ b/flash/main.c
@@ -120,13 +120,13 @@ int main(int ac, char** av)
 
   if (o.devname != NULL) /* stlinkv1 */
   {
-    sl = stlink_v1_open(50, 1);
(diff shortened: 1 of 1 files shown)


xor-gate:
You did resolve a issue 2 years ago with this PR. I'm closing this for now because it should be fixed in with tools in master and nobody did comment on this. When the problem still occurs rebase with master and open a issue, and after this new a squashed pull request. Thanks!

PR #328, opened 2015-08-15
Make _stlink_usb_reset use hardreset

mlu:
This change makes my board, STM32F7 Discovery actually reset

(final patch for today :) )

### DIFF
--- a/src/stlink-common.h
+++ b/src/stlink-common.h
@@ -68,6 +68,7 @@ extern "C" {
 #define STLINK_DEBUG_WRITEDEBUGREG	0x0f
 #define STLINK_DEBUG_ENTER_SWD		0xa3
 #define STLINK_DEBUG_ENTER_JTAG	0x00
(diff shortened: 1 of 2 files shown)


texane:
Thanks for the patches :)

I am not sure the last one will work for all the boards supported
so far, but I merged it. If people have issues, we will revert and
do something more generic.

2015-08-15 19:45 GMT+02:00 mlu <notifications@github.com>:

mlu:
I think the patch should be reverted, for now.

We should keep the define for hardware reset and add a hardware reset function so both are available. 
The do a bit of careful testing of how much the is reset, and also consider if we want reset+debughalt 
or simply reset run.

/Magnus
 

15 aug 2015 kl. 19:49 skrev texane <notifications@github.com>:

texane:
Hi,

could you create a request to revert it ?

Cheers,

2015-08-21 15:36 GMT+02:00 mlu <notifications@github.com>:

Issue #308, opened 2015-04-27
Flash data gets corrupted after verify passed (STM32F429I-Discovery)


ambrop72:
When I flash my program to the board using st-flash, the flashing will succeeds with verification passing, but after some time or power cycling, some data in a particular part becomes corrupted. I have checked this by reading back data using st-flash and diff'ing the hexdumps - I can see a few bytes different.

**When the chip is flashed using OpenOCD, the data does not evaporate and is flashed reliably.**

xor-gate:
I'm digging up old issues, this should be fixed by PR #371. Please open a new detailed issue when this problem still occurs with latest stlink tools from master and latest stlink firmware (2.27).

PR #371, opened 2016-02-19
remove empty byte truncation from stlink_fread

gtalusan:
I'm trying to backup the flash memory on an EMW3165 with an eBay ST-Link v2 using the following command:
```
st-flash read out.bin 0x8000000
```
I expect the output to be 0x80000 bytes in size but I get something quite shorter.  Appending the length to the command line also produces the same short output.  Looking at the code, I see that the number of FF's or 00's is counted on each iteration of the read, but the file is truncated after all of the reads are done.

I think the writes should be shortened inside the loop, but it looks like honouring `sl->q_len` is the way to go.

This patch removes the erased byte counter, bumps up the read size to min(0x1800, flash page size) similar to the verify flash data function.

### DIFF
--- a/src/stlink-common.c
+++ b/src/stlink-common.c
@@ -1032,9 +1032,6 @@ int stlink_fread(stlink_t* sl, const char* path, stm32_addr_t addr, size_t size)
     int error = -1;
     size_t off;
     int num_empty = 0;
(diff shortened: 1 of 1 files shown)


gtalusan:
Also wonder if #308 is related.

PR #391, opened 2016-04-06
st-probe: Initial working stlink_probe_* API and CLI tool

xor-gate:
Hi all,

I'm creating a continues integration system where a machine has multiple stlinks connected. There is currently no good auto-probe tool/library to find the current connected usb stlinks. With this autoprobe I generate openocd configurations to attach to all the programmers.

The following changes have been made:

* Added `stlink_probe_usb`
* Added `stlink_probe_usb_free`
* Extended `stlink_t with serial field`
* Added example CLI tool `st-probe`

Let me know what you think, there is a little code duplication with this addition but the current `stlink_open_usb` is way to big for me to refactor. I tried to refactor the `stlink_open_usb` function but failed and broke the whole thing.

Here is some example output:
```
jjacobs@workbench ~/repos/github/stlink/build $ ./st-probe
Found 2 stlink programmers
563f6706677256563113066700
         flash: 262144 (pagesize: 2048)
          sram: 65536
        chipid: 0x0414
         descr: F1 High-density device
493f6d06667048513615106700
         flash: 1048576 (pagesize: 16384)
          sram: 196608
        chipid: 0x0413
         descr: F4 device
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,3 +16,4 @@ marpe@mimuw.edu.pl
 marco.cassinerio@gmail.com
 jserv@0xlab.org
 michael@pratt.im
+jerry.jacobs@xor-gate.org
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,10 @@ target_link_libraries(st-util stlink)
 add_executable(st-info src/st-info.c)
 target_link_libraries(st-info stlink)
 
-install(TARGETS  stlink st-flash st-util st-info
+add_executable(st-probe src/st-probe.c)
+target_link_libraries(st-probe stlink)
+
+install(TARGETS  stlink st-flash st-util st-info st-probe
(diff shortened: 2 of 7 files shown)


rikvdh:
:+1: for st-probe

xor-gate:
@texane thanks for the merge awesomeness, i hope other people will like it to. My ideas to expand the st-probe tool with csv,json or openocd configuration outputs when the argument is given. What do you think? The idea to emit openocd configuration is because the gdb-stub is much more complete. It fully supports ARM semihosting which is very usefull in CI/Endurance environments. And `st-util` doesn't support this at all, but this is a different story.

texane:
Thanks for your work. The features you propose look useful ... let know if
you
work on these, I will be glad to merge them.

2016-04-06 18:16 GMT+02:00 Jerry Jacobs <notifications@github.com>:

PR #395, opened 2016-04-08
Styling

halt-hammerzeit:
In case you find it appropriate


halt-hammerzeit:
Also, just in case you know, is it wrong when it says 
```
ERROR gdbserver/gdb-server.c: cannot send: -2
```
When I stop the execution of the program?
(stutil exits then)

Nightwalker-87:
@catamphetamine: Yes, that looks somehow reasonable to me as the running gdb process is interrupted.

PR #400, opened 2016-04-29
st-info: Add support for OpenOCD hla_serial printing

xor-gate:
Initial work on OpenOCD hla serial info, which is a followup of work not done in PR  https://github.com/texane/stlink/pull/391#issuecomment-206467687

This also fixes the link/build image of the README.md to texane/stlink. Please would you enable the build and add support for building pull requests under Travis ?

### DIFF
--- a/README
+++ b/README
@@ -1,7 +1,7 @@
 STMicroelectronics Stlink Tools
 ===============================
 
-[![Build Status](https://travis-ci.org/xor-gate/stlink.svg?branch=travis)](https://travis-ci.org/xor-gate/stlink)
+[![Build Status](https://travis-ci.org/texane/stlink.svg?branch=travis)](https://travis-ci.org/texane/stlink)
 
 ## HOWTO
 
--- a/src/st-info.c
+++ b/src/st-info.c
@@ -13,7 +13,28 @@ static void usage(void)
     puts("st-info --pagesize");
     puts("st-info --chipid");
     puts("st-info --serial");
+    puts("st-info --hla-serial");
     puts("st-info --probe");
+}
+
+/* Print normal or OpenOCD hla_serial with newline */
+static void stlink_print_serial(stlink_t *sl, bool openocd)
+{
+    const char *fmt;
+
+    if (openocd) {
+       printf("\"");
+       fmt = "\\x%02x";
+    } else {
+       fmt = "%02x";
+    }
+
+    for (int n = 0; n < sl->serial_size; n++)
+        printf(fmt, sl->serial[n]);
+
+    if (openocd)
+       printf("\"");
+    printf("\n");
 }
 
 static void stlink_print_info(stlink_t *sl)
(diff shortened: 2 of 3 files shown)


texane:
Sorry, I do not know about Travis. I will investigate that asap.

xor-gate:
@texane I understand, it is very simple. You link your github account with https://travis-ci.org and enable the stlink repository for building. Simple as that!

texane:
OK. But I do not want to register blindly to (another ...) web service
before understanding the added value ... Thanks for the suggestion,
I will have a look.

2016-04-29 10:48 GMT+02:00 Jerry Jacobs <notifications@github.com>:

xor-gate:
@texane yes I completely understand. You just have to authorized the 3thparty app (travis) to pull sources from github. Continues integration is highly necessary to develop a good quality software which people actually use... and during development things break, people change things, make things and break things. CI resolves all the chaos and checks if it still builds (and runs). 

You really should have a read on that topic: https://en.wikipedia.org/wiki/Continuous_integration

texane:
Yes, I know about CI.

To me, the major problem we have with the STLINK project is to test changes
on all the boards known so far. CI does not solve that particular point,
unless
a farm of physical boards is available somewhere to test every case.
Distributed
validation by the user themselves is a working and acceptable solution.

2016-04-29 11:29 GMT+02:00 Jerry Jacobs <notifications@github.com>:

xor-gate:
Yes I totaly agree with your points, but before the software actually hits the user it should be checked automaticly if it builds. Running on "all" hardware is indeed a different story, but this should be no reason to not enable Public CI.

csdexter:
That is to say there are two levels of testing:
1. does the entire code base still make sense w.r.t. the language used? (Which is checked by compiling, linking and checking for success -- something that Travis was born to do)
2. does the code base still run on platform X and Y on which it ran yesterday? (Which is covered by your distributed testing)

xor-gate:
Exactly @csdexter, this is correctly summarized.

csdexter:
It's worth also mentioning that in a perfect world, you'd have an emulator which can be scripted (both in terms of "pretend to be this particular STM32 chip" and in terms of "run this code, expect the SPI to spit out exactly those bytes, exit non-zero if it didn't happen") so that you could do automated end-to-end testing (which could then be configured in Travis too).

I know this is a moderate to considerable pain-in-the-neck for AVR and I don't think it's doable for ARM at all today (especially as things like QEMU focus on application-class ARM chips, not controller-class ones).

xor-gate:
What you can do is write a libusb stub for uniting testing but then you are far from home :=)

csdexter:
It should actually be simpler than it sounds. I'm not privy to the ST-Link comms internals, but I'm guessing it's a matter of request/response and so very basic unit testing could be written along the lines of "pretend to report value X in register Y" and then assert that "st-flash attempts to write a Flash page at X" or similar.

That is, to test that the behaviour you expect st-flash to have given certain input data is actually displayed as designed.

Same goes for invalid/bogus data (see long stream of "I have this new chip and st-flash was totally confused when it connected to it" situations).

xor-gate:
Yeah it is simple as you mention, but writing it is a different story.

csdexter:
Well, I guess I should open my editor and start whipping something up, then :-)

texane:
Agree with everything that has been said so far in this disucssion. Trust
you on
the added value, so I enabled Travis CI. Wont have time to investigate it
in near
future though ...

2016-04-29 13:17 GMT+02:00 Radu - Eosif Mihailescu <notifications@github.com

xor-gate:
@texane thanks for your trust and effort to enable travis

Issue #379, opened 2016-03-01
STM32L152RB chip id not found in SWD mode


pengshulin:
I connected a standalone st-linkv2(not the board integrated one) with STM32L152RBT6 embedded in Discovery board in SWD mode (RST+CLK+DAT, I've disconnect the three pins with the integrated stlink), but failed to find the chip.  The integrated stlink works well.
In the debugging info, core_id is identified (I think the hardware connection no problem), but reads 0 from DBGMCU_IDCODE(@0xE0042000).
The stlink is upgraded to the latest firmware and st-flash tool is built from the master branch.
Can anyone help me?

Here's the debug error info:
`10:26:57 DEBUG src/stlink-common.c: stlink current mode: debug (jtag or swd)`
`10:26:57 DEBUG src/stlink-common.c: stlink current mode: debug (jtag or swd)`
`10:26:57 DEBUG src/stlink-common.c: *** stlink_reset ***`
`10:26:57 DEBUG src/stlink-common.c: *** looking up stlink version`
`10:26:57 DEBUG src/stlink-common.c: st vid         = 0x0483 (expect 0x0483)`
`10:26:57 DEBUG src/stlink-common.c: stlink pid     = 0x3748`
`10:26:57 DEBUG src/stlink-common.c: stlink version = 0x2`
`10:26:57 DEBUG src/stlink-common.c: jtag version   = 0x18`
`10:26:57 DEBUG src/stlink-common.c: swim version   = 0x4`
`10:26:57 INFO src/stlink-common.c: Loading device parameters....`
`10:26:57 DEBUG src/stlink-common.c: *** stlink_core_id ***`
`10:26:57 DEBUG src/stlink-common.c: core_id = 0x2ba01477`
`10:26:57 DEBUG src/stlink-common.c: *** stlink_read_debug32 0 from 0xe0042000`
`10:26:57 DEBUG src/stlink-common.c: *** stlink_read_debug32 0 from 0x40015800`
`10:26:57 WARN src/stlink-common.c: unknown chip id! 0`
`10:26:57 DEBUG src/stlink-common.c: *** stlink_close ***`

If I use the integrated stlink, it works well, here's the right info:
`16:36:23 DEBUG src/stlink-common.c: stlink current mode: debug (jtag or swd)`
`16:36:23 DEBUG src/stlink-common.c: stlink current mode: debug (jtag or swd)`
`16:36:23 DEBUG src/stlink-common.c: *** stlink_reset ***`
[...]

PR #413, opened 2016-05-14
Fix the 'st-flash erase' crash in Cygwin issue

gerryqd:
In this case, function 'get_opts()' will crash, as it will
access memory outside of 'av' array.

This fix add the protection for this case.


xor-gate:
Closing this after discussion of optparse rewrite in #379.

Godsentp:
Hello

Nightwalker-87:
Obsolete with commit 7c99eb050300b76bfd64b380d7c8693476d540e2 that mered with #459.

PR #414, opened 2016-05-16
Add L0 Category 2 device (chip id: 0x425)

raron:
I tried it with: [NUCLEO-L031K6](http://www.st.com/web/catalog/tools/FM116/CL1620/SC959/SS1532/LN1847/PF262547)
I'm not sure about the sram_size, but based on [this](http://www.st.com/web/catalog/mmc/FM141/SC1544/SS1833/LN1862) and [RM0377](http://www.st.com/st-web-ui/static/active/en/resource/technical/document/reference_manual/DM00108282.pdf) every category 2 device has 8 kB SRAM (now).

### DIFF
--- a/include/stlink.h
+++ b/include/stlink.h
@@ -138,6 +138,7 @@ extern "C" {
 
 #define STM32_CHIPID_F303_HIGH      0x446
 #define STM32_CHIPID_L0_CAT5        0x447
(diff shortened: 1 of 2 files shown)


xor-gate:
Thanks for your testing and contribution!

Issue #402, opened 2016-04-29
ironing out compiler warnings and code smells


xor-gate:
For all compiler warnings you should look at Travis CI: https://travis-ci.org/texane/stlink

When developing the probe and openocd hla_serial functionality I extended the CMake build with extra compiler flags which detect code smells (see https://sourcemaking.com/refactoring/smells). Here is the build info with mostly variable shadowing (different scoped variables with same name), precision lost (shorten, on clang compiler), missing variable declarations (written in header file or defined with static scope), :

```
-- Build files have been written to: /Users/travis/build/xor-gate/stlink/build
Scanning dependencies of target stlink
[ 11%] Building C object CMakeFiles/stlink.dir/src/stlink-common.c.o
/Users/travis/build/xor-gate/stlink/src/stlink-common.c:1020:36: warning: implicit conversion loses integer precision: 'unsigned long' to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        stlink_read_mem32(sl, addr + off, aligned_size);
        ~~~~~~~~~~~~~~~~~     ~~~~~^~~~~
/Users/travis/build/xor-gate/stlink/src/stlink-common.c:1079:37: warning: implicit conversion loses integer precision: 'unsigned long' to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        stlink_write_mem32(sl, addr + off, size);
        ~~~~~~~~~~~~~~~~~~     ~~~~~^~~~~
/Users/travis/build/xor-gate/stlink/src/stlink-common.c:1084:36: warning: implicit conversion loses integer precision: 'unsigned long' to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        stlink_write_mem8(sl, addr + len, mf.len - len);
        ~~~~~~~~~~~~~~~~~     ~~~~~^~~~~
/Users/travis/build/xor-gate/stlink/src/stlink-common.c:1138:36: warning: implicit conversion loses integer precision: 'unsigned long' to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        stlink_read_mem32(sl, addr + off, aligned_size);
        ~~~~~~~~~~~~~~~~~     ~~~~~^~~~~
/Users/travis/build/xor-gate/stlink/src/stlink-common.c:1165:45: warning: implicit conversion loses integer precision: 'unsigned long' to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
[...]

PR #419, opened 2016-05-19
Fix const warning on struct stlink_chipid_params

jmlemetayer:
Just fix the 2 warnings:

    warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]

### DIFF
--- a/src/chipid.c
+++ b/src/chipid.c
@@ -389,7 +389,7 @@ static const struct stlink_chipid_params devices[] = {
 
 const struct stlink_chipid_params *stlink_chipid_get_params(uint32_t chipid)
 {
-	struct stlink_chipid_params *params = NULL;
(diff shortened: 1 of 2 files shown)


xor-gate:
Thanks for fixing! I have overlooked this. And also have a related issue open (when compiling with CMake):
https://github.com/texane/stlink/issues/402

jmlemetayer:
No problem. Thanks for your reactivity.

xor-gate:
If you have time and would like to contribute some more maybe you can look into issue #402?

jmlemetayer:
I will try to check this issue

xor-gate:
Great, when you work on it, and need some discussion you should write at #402. Thanks!

PR #420, opened 2016-05-20
WIP: Fix compiler warnings

jmlemetayer:
Trying to fix all warnings noticed by #402 for the different builds:

- [x] linux / gcc / autotools
- [x] linux / clang / autotools
- [x] linux / gcc-5 / cmake
- [x] linux / clang-3.8 / cmake
- [x] osx / clang / cmake
- [x] osx / clang / autotools

### DIFF
--- a/include/stlink.h
+++ b/include/stlink.h
@@ -67,7 +67,7 @@ extern "C" {
         STLINK_FLASH_TYPE_L4
     };
 
-    typedef struct {
+    struct stlink_reg {
         uint32_t r[16];
         uint32_t s[32];
         uint32_t xpsr;
@@ -80,7 +80,7 @@ extern "C" {
         uint8_t basepri;
         uint8_t primask;
         uint32_t fpscr;
-    } reg;
+    };
 
     typedef uint32_t stm32_addr_t;
 
@@ -166,11 +166,11 @@ typedef struct flash_loader {
     int stlink_write_debug32(stlink_t *sl, uint32_t addr, uint32_t data);
     int stlink_write_mem32(stlink_t *sl, uint32_t addr, uint16_t len);
     int stlink_write_mem8(stlink_t *sl, uint32_t addr, uint16_t len);
-    int stlink_read_all_regs(stlink_t *sl, reg *regp);
-    int stlink_read_all_unsupported_regs(stlink_t *sl, reg *regp);
-    int stlink_read_reg(stlink_t *sl, int r_idx, reg *regp);
-    int stlink_read_unsupported_reg(stlink_t *sl, int r_idx, reg *regp);
-    int stlink_write_unsupported_reg(stlink_t *sl, uint32_t value, int r_idx, reg *regp);
+    int stlink_read_all_regs(stlink_t *sl, struct stlink_reg *regp);
+    int stlink_read_all_unsupported_regs(stlink_t *sl, struct stlink_reg *regp);
+    int stlink_read_reg(stlink_t *sl, int r_idx, struct stlink_reg *regp);
+    int stlink_read_unsupported_reg(stlink_t *sl, int r_idx, struct stlink_reg *regp);
+    int stlink_write_unsupported_reg(stlink_t *sl, uint32_t value, int r_idx, struct stlink_reg *regp);
     int stlink_write_reg(stlink_t *sl, uint32_t reg, int idx);
     int stlink_step(stlink_t *sl);
     int stlink_current_mode(stlink_t *sl);
--- a/src/common.c
+++ b/src/common.c
@@ -170,18 +170,6 @@ static uint32_t __attribute__((unused)) read_flash_rdp(stlink_t *sl) {
     return rdp & 0xff;
 }
 
-static inline uint32_t read_flash_wrpr(stlink_t *sl) {
-    uint32_t wrpr;
-    stlink_read_debug32(sl, FLASH_WRPR, &wrpr);
-    return wrpr;
-}
-
-static inline uint32_t read_flash_obr(stlink_t *sl) {
-    uint32_t obr;
-    stlink_read_debug32(sl, FLASH_OBR, &obr);
-    return obr;
-}
-
 static inline uint32_t read_flash_cr(stlink_t *sl) {
     uint32_t reg, res;
 
@@ -364,12 +352,6 @@ static void set_flash_cr_strt(stlink_t *sl) {
     stlink_read_debug32(sl, cr_reg, &val);
     val |= cr_strt;
     stlink_write_debug32(sl, cr_reg, val);
-}
-
-static inline uint32_t read_flash_acr(stlink_t *sl) {
-    uint32_t acr;
-    stlink_read_debug32(sl, FLASH_ACR, &acr);
-    return acr;
 }
 
 static inline uint32_t read_flash_sr(stlink_t *sl) {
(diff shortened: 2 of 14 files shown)


xor-gate:
looks good, but if we are going to move to a enforced codestyle with tooling (checkpatch + linux codestyle) typedefs are forbidden so we should change this to something like: `struct stlink_reg`.

jmlemetayer:
I just wanted not to break the "code spirit". But if it's ok for you I will change this :smiley:

I have set this pull request as WIP to check the travis builds and also to have feedbacks. Thanks for your support!

xor-gate:
Yeah I know what you mean, but in the upcoming time -> Milestone 2.0.0 there will some cleanup going on for easier understanding of the codebase and contribution.

jmlemetayer:
Sounds interresting. If I have extra time I will think about it.

xor-gate:
See https://github.com/texane/stlink/milestones -> 2.0.0

xor-gate:
@jmlemetayer make sure you rebase with latest master because i have moved the flash loader functions into a seperate file.

jmlemetayer:
The remainings warnings `typedef ‘_GStaticAssertCompileTimeAssertion_0’ locally defined but not used` are a known glib bug which has been fixed (I don't have this issue on my ubuntu 16.04).

@xor-gate Is that a problem for you?

xor-gate:
We have to see how the glibc compiler warning will behave because I want to enable `-Werror` then all warnings are upgraded as errors. As for travis we run ubuntu 14 indeed. But there is a way to have newer VMs.

src/tools/gui/stlink-gui.c
```diff
@@ -518,7 +516,7 @@ stlink_gui_set_connected (STlinkGUI *gui)
 }
 
 static void
-connect_button_cb (GtkWidget *widget, gpointer data)
+connect_button_cb (GtkWidget * UNUSED(widget), gpointer data)
```
xor-gate: This is way ugly. You can just do it in a more portable way (pseudo-C):

<function> (param1, param2) {
     (void)param1;
     param2++;
}

jmlemetayer:
Yeah, this is what I expected. I see 2 solutions:

 - Upgrade the GLib version in travis
 - Do not build GUI in travis (as the GLib is only used by GTK+ in the GUI)

include/stlink.h
```diff
@@ -15,6 +15,15 @@
...
 
+#ifdef UNUSED
+#elif defined(__GNUC__)
+#define UNUSED(x)	UNUSED_ ## x __attribute__((unused))
+#elif defined(__LCLINT__)
+#define UNUSED(x)	/*@unused@*/ x
+#else
+#define UNUSED(x)	x
+#endif
+
```
xor-gate: We should not do this, see my other comment. As this is highly compiler depended and I want in some time also support native visual studio compilation. Also when we really need to do this, sure not in the public scope.

src/common.c
```diff
@@ -181,6 +182,7 @@ static inline uint32_t read_flash_obr(stlink_t *sl) {
     stlink_read_debug32(sl, FLASH_OBR, &obr);
     return obr;
 }
+#endif
```
xor-gate: We need to throw this away when unused.

src/common.c
```diff
@@ -366,11 +368,13 @@ static void set_flash_cr_strt(stlink_t *sl) {
     stlink_write_debug32(sl, cr_reg, val);
 }
 
+#if 0
 static inline uint32_t read_flash_acr(stlink_t *sl) {
     uint32_t acr;
     stlink_read_debug32(sl, FLASH_ACR, &acr);
     return acr;
 }
+#endif
```
xor-gate: Same comment as above, throw away when unused/dead code.

src/tools/flash.c
```diff
@@ -84,18 +84,18 @@ static int get_opts(struct opts* o, int ac, char** av)
     {
         ac--;
         av++;
-	    int i=strlen(av[0]);
-	    if(i%2 != 0){
+	    int j = (int) strlen(av[0]);
```
xor-gate: this is not really portable, as strlen really returns `size_t` we need to obey and not cast it to a signed type.

src/tools/flash.c
```diff
@@ -154,7 +154,7 @@ static int get_opts(struct opts* o, int ac, char** av)
     }
 
     o->filename = av[i + 1];
-    o->addr = strtoul(av[i + 2], NULL, 16);
+    o->addr = (uint32_t) strtoul(av[i + 2], NULL, 16);
```
xor-gate: this is a little evil as strtoul could return 0 and is of type `unsigned long int`

src/tools/term.c
```diff
@@ -41,8 +41,8 @@ struct stlinky {
                                           | buf[2] << 16   \
                                           | buf[3] << 24))
 
-static stlink_t* sl;
-sigset_t sig_mask;
+static stlink_t* gsl;
```
xor-gate: I'm not happy with a global, but prefixing it with g is not bad.

src/usb.c
```diff
@@ -737,7 +737,7 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, bool reset, char serial[16
     }
 
     libusb_device **list;
-    int cnt = libusb_get_device_list(slu->libusb_ctx, &list);
+    int cnt = (int) libusb_get_device_list(slu->libusb_ctx, &list);
```
xor-gate: We should use `ssize_t` and use it as a counter if > 0. As per libusb API:
`ssize_t 	libusb_get_device_list (libusb_context *ctx, libusb_device ***list)`

xor-gate:
To make sure people dont break things we need to build the GUI so disable is no option. Upgrading, I have no clue and needs investigation (but what about people then building with older glibc will have problems). Have you tried the cmake `SYSTEM` directive?

Most of the refactoring seems fine, but a few `brutal-casts` need to be looked at better because the compiler is actually right about the problem. Of course a `brute-force` cast then the warning will go away but some bugs could come in.

xor-gate:
Weird, no CI build is triggerd with last cmake fixup. Could you try to squash merge and push and see if it builds?

jmlemetayer:
GitHub had some issue this afternoon. I will try again tomorrow. I think you can also trigger a new build directly from travis-CI.

xor-gate:
Yeah I noticed also some bugging, you can only rebuild pushed commits. Thank you for you time and effort!

jmlemetayer:
I remove the `UNUSED` macro as requested. I was not aware that the `(void)` trick is much more portable than a macro. I will check your other comments later.

I have added the `-Werror` and the `SYSTEM` but it doesn't seems to work with the `cmake 3.8` / `clang` couple...

jmlemetayer:
@xor-gate About your comments above. I agree that it is not beautiful, but the code was already here.

For now, I don't think we should clean the code. Just fix warnings. The cleanup should be done in a second time.

xor-gate:
You are right about that, maybe we should add doxygen style comments to track todos where it needs attention/refactoring for the second step like this:

```
/** @todo refactor/review */
```

Then when I enable building doxygen documentation we can see a list of todos or grep trough the code for todo. See https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdtodo

xor-gate:
Could you try the following with the BEFORE statement, and see if it works?:

```cmake
include_directories(BEFORE SYSTEM <gtk include var>)
```

xor-gate:
Hi @jmlemetayer what do you think, should we merge without `-Werror` because if this oddness?

jmlemetayer:
Yes I think it is a good idea to remove `-Werror` for now. I did not manage to find a solution for the GLib issue.

jmlemetayer:
From http://llvm.org/apt/

> APT mirror was temporary switched off due to excess load. We are working on bringing it back. Stay tuned!

All linux build are failed ...

xor-gate:
Seems good to me, the llvm apt mirror is a different problem. We still need to refactor bits-and-pieces before hitting `v2.0.0`. Thank you for your work.

PR #430, opened 2016-06-07
Do a JTAG reset prior to reading CPU information

andyg24:
When a microcontroller is in deep sleep (stop) mode, its debug interface is turned off. Programming would therefore fail because st-flash was unable to identify the CPU. This change adds a JTAG reset prior to requesting CPU information. This change has been successfully tested with a Nucleo STM32L031 board.

### DIFF
--- a/src/usb.c
+++ b/src/usb.c
@@ -877,6 +877,7 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, bool reset, char serial[16
     }
 
     if (reset) {
+        stlink_jtag_reset(sl, 2);
         stlink_reset(sl);
         usleep(10000);
     }


xor-gate:
Sorry for the late reply, but I would like to test this more throughfully before merging if you don't mind. In the upcoming weeks I will have a continues integration setup with real hardware (nucleo and discovery boards). I will keep you posted.

xor-gate:
I'm not sure if usb.c is the correct place to add a jtag reset. As the probe/open just looks for programmers. Why is this not working: https://github.com/texane/stlink/blob/c0177ec3872244fb04fc517397059f742e2b1cc3/src/tools/flash.c#L210-L220

andyg24:
It's been a while, but I believe the reset also needs to be in usb.c
because otherwise the ID is not read and programming in flash.c is not
attempted.

xor-gate:
Yeah I think you are right, the jtag reset will toggle the NRST of the chip and everything is cleared. I'm not sure why value 2 is given, as a pin toggle should only have 0 or 1. But the protocol to the programmer is under NDA and reversed engineered so I can not tell.

Issue #439, opened 2016-06-21
Regression when programming STM32F4xx device from a STM32VL Discovery kit


dflogeras:
Hi, I often use a STM32VL Discovery as a programmer, because it was cheap and I don't really have any other use for it :)  Today I revived a project where the target is a homebrew STM32F407 board.  I tried to use the STM32VL Discovery as the USB programmer.  I know this used to work, but today I was getting a "Failed to read target voltage" and then a bunch of USB errors.  After which I would have to unplug the USB cable before I could communicate again.  This was on the current master.

I tried 1.1.0 and it worked, so I bisected and I think the problem was introduced in 71535104e47f2a96ce6a87a6da0fc4519bc64d15

If it matters, this is on Gentoo Linux, compiled with autotools, GCC 4.9.3.

The device is a  STM32F405RGT6W, and st-util -1 reports:

```
2016-06-21T17:04:33 INFO src/stlink-common.c: Device connected is: F4 device, id 0x10016413
2016-06-21T17:04:33 INFO src/stlink-common.c: SRAM size: 0x30000 bytes (192 KiB), Flash: 0x100000 bytes (1024 KiB) in pages of 16384 bytes
2016-06-21T17:04:33 INFO src/stlink-sg.c: Successfully opened a stlink v1 debugger
2016-06-21T17:04:33 INFO gdbserver/gdb-server.c: Chip ID is 00000413, Core ID is  2ba01477.
```

xor-gate:
I'm not sure, but I think stlink v1 is unable to read target voltage. Which firmware version for the program are do you use? Probably you should first try to upgrade as I have seen weird issues with broken stlink programmer firmware. Over here for v1:
http://www.st.com/content/st_com/en/products/embedded-software/development-tool-software/stsw-link001.html

dflogeras:
Well, if I plug it into the Windows STM32 ST-LINK utility, and click upgrade->device connect, a dialog pops up and says "Old firmware loader detected on ST-Link/V2.  For full operation (especially SWV support), please contect ST for upgrading."

Then it reports the Firmware Version: V0.J0.S0 STM32 Debugger

And says it wants to upgrade to "V2.J24.S0".

All of that makes me nervous.  It seems to want to upgrade to V2, and I'm not sure if this device is supposed to have V2.  Also, on previous times I've upgraded the STLink firmware, I didn't get the warning dialog, but they were already V2 devices.

Any ideas?  The link you sent shows me a version "1.12.3"

[...]

PR #440, opened 2016-06-21
Fix a regression introduced in 71535104.

dflogeras:
Since STLINK V1 cannot read voltage, and attempting to do so breaks the
USB communications, this patch will revert the behaviour to just blindly
doing 32-bit writes on these programmers.


xor-gate:
I'm not sure, the indenting seems a bit odd.

dflogeras:
I think github's diff generator makes it look weird, I followed the existing convention of 4 spaces.  Clone my repo first and have a look at it in context if you like.

dflogeras:
PS I did step through it in a debugger and it works for me as expected.

xor-gate:
Crossref for issue #439

dflogeras:
Sorry to do this, but I'm going to rebase this (used wrong email) and create another pull request.  It is identical save that.

Nightwalker-87:
Moved to #441.

PR #441, opened 2016-06-22
Fix a regression introduced in 71535104.

dflogeras:
Since STLINK V1 cannot read voltage, and attempting to do so breaks the
USB communications, this patch will revert the behaviour to just blindly
doing 32-bit writes on these programmers.


dflogeras:
Sorry, rebased the same work because I used the wrong email

texane:
Thanks for the fix

PR #447, opened 2016-07-22
Update udev path in README

toniokroeger:


xor-gate:
@toniokroeger thanks for spotting this!

PR #452, opened 2016-08-09
Unhexify GDB monitor commands (qRcmd)

Fabien-Chouteau:
It improves readability and will allow parsing of more complex monitor
commands.

### DIFF
--- a/src/gdbserver/gdb-server.c
+++ b/src/gdbserver/gdb-server.c
@@ -927,6 +927,21 @@ static void cache_sync(stlink_t *sl)
     cache_flush(sl, ccr);
 }
 
+static size_t unhexify(const char *in, char *out, size_t out_count)
+{
+    size_t i;
+    unsigned int c;
+
+    for (i = 0; i < out_count; i++) {
+        if (sscanf(in + (2 * i), "%02x", &c) != 1) {
+            return i;
+        }
+        out[i] = (char)c;
+    }
+
+    return i;
+}
+
 int serve(stlink_t *sl, st_state_t *st) {
     int sock = socket(AF_INET, SOCK_STREAM, 0);
     if(sock < 0) {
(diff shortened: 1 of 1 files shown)


texane:
Thanks for the pull. I spotted at least one error in your code, could you
fix it
and pull again ?

You allocate char cmd[cmd_len];
but you access the cmd_len + 1 item with cmd[cmd_len] = 0;

You should thus allocate at least cmd_len + 1 item.

Also, prefer static allocations when using the stack. Otherwise,
use malloc.

Thanks

2016-08-09 16:35 GMT+02:00 Fabien Chouteau <notifications@github.com>:

xor-gate:
You should prefer `size_t` over `int` with sizes since C99. A other implementation which implements a gdb server is the blackmagic probe which can be used as a reference:

https://github.com/blacksphere/blackmagic/blob/master/src/hex_utils.c
https://github.com/blacksphere/blackmagic/blob/master/src/gdb_main.c
https://github.com/blacksphere/blackmagic/blob/master/src/gdb_packet.c

Fabien-Chouteau:
I'm not sure if this is the standard workflow but I updated the commit. (git commit --amend && git push -f)

xor-gate:
Yeah it is nice to have just one commit per fix/pull request. This is the same as rebase-squash (`git rebase -i master`)

texane:
Merged, thanks. But I would suggest, for total correctness, checking outside the unhexify() for loop that sscanf access:
in[2 * i + 0]
in[2 * i + 1]
is always valid, which may not be the case. Rare conditions (in odd size + cross page allocation), but does not cost much.

Fabien-Chouteau:
We can check if strlen(params) is a multiple of 2, which it should be.

I don't see how cross page allocation can be a problem.

texane:
'it should be' ?  What about buggy clients ? Do not suppose, esp. if the check comes for free. Truncate the size if not a 2 multiple.
Cross page -> sscanf invalid read access even on a single invalid byte may cause a crash, depending on the next page access rights.

Fabien-Chouteau:
I know it can happen, that's why I said 'it should be' not 'it will always be'. Note that the integer division (size_t alloc_size = (hex_len / 2) + 1;) will handle the case where strlen(params) is odd.

PR #456, opened 2016-08-15
Fix memory map for STM32F7xxx

zulusw:
Also:
- fixed a typo in an if statement condition in make_memory_map
- cleaned up usage of '4096' in favor of const size_t sz = 4096.


texane:
Nice, merged. Thanks.

Issue #110, opened 2012-10-05
Reading application from hexfile for use with st-flash


Shaun1:
Is there any support currently for flashing directly from a .hex file using st-flash?

xor-gate:
Implemented in PR #459.

Issue #239, opened 2014-05-28
'addr too high' ERROR - maybe CCM? stm32f407


rusefi:
We are failing to program our .hex file using stlink. Same .hex file works OK if flashed using the official "st link utility" on Windows. The only thing maybe special about our hex file is that we are using CCM memory region - we've tried and it seem to work if CCM is not used.

The failing to flash binary is available at http://rusefi.com/build_server/rusefi_firmware.zip

the error message is:
build red$ ~/Development/embedded/tools/stlink/st-flash write rusefi.bin 0x08000000
2014-05-28T16:45:48 INFO src/stlink-usb.c: -- exit_dfu_mode
2014-05-28T16:45:48 INFO src/stlink-common.c: Loading device parameters....
2014-05-28T16:45:48 INFO src/stlink-common.c: Device connected is: F4 device, id 0x10016413
2014-05-28T16:45:48 INFO src/stlink-common.c: SRAM size: 0x30000 bytes (192 KiB), Flash: 0x100000 bytes (1024 KiB) in pages of 16384 bytes
2014-05-28T16:45:50 INFO src/stlink-common.c: Attempting to write 134241388 (0x8005c6c) bytes to stm32 address: 134217728 (0x8000000)
2014-05-28T16:45:50 ERROR src/stlink-common.c: addr too high
stlink_fwrite_flash() == -1

We've tried some older source codes and current version from today (May 28, 2014)

VARSHNED:
I am trying to flash the demo software of STM32F769 discovery kit. But below is the error that i am obtaining

st-flash 1.4.0-13-g1969148
2018-01-30T17:40:14 INFO common.c: Loading device parameters....
2018-01-30T17:40:14 INFO common.c: Device connected is: F76xxx device, id 0x10016451
2018-01-30T17:40:14 INFO common.c: SRAM size: 0x80000 bytes (512 KiB), Flash: 0x200000 bytes (2048 KiB) in pages of 2048 bytes
2018-01-30T17:40:14 INFO common.c: Attempting to write 26556392 (0x19537e8) bytes to stm32 address: 134217728 (0x8000000)
2018-01-30T17:40:14 ERROR common.c: addr too high
stlink_fwrite_flash() == -1

PR #459, opened 2016-09-06
Support of Intel-HEX format for st-flash

dev26th:
Hi

here is my implementation of Intel-HEX format for st-flash (both reading and writing). The format can be selected with new command line option '--format=ihex' (one may not specify <addr> for write with the format, the address will be taken from input file).

Additionally I've re-designed the command line parsing, because it was too complicated with the new option. And added some unit-tests for parsing itself to guaranty all old options still work.

Unfortunately it's not so easy to split this pull request to single features, but I could try if you prefer.

WMBR,
dev26th

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,8 +145,11 @@ if (APPLE)
 	target_link_libraries(${PROJECT_NAME} ${CoreFoundation} ${IOKit} ${ObjC})
 endif ()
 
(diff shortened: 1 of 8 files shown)


texane:
Good one, thanks.
Merged, fixed a warning due to missing initializer (ie. empty braces {})

Nightwalker-87:
Closes #110, #157, #239.

PR #463, opened 2016-09-14
CMakeLists.txt: Fix linker command in generated .pc file

nekromant:
Signed-off-by: Andrew Andrianov <andrew@ncrmnt.org>


xor-gate:
Nice catch, thanks!

Issue #208, opened 2014-02-17
Add UNIX man pages


anatol:
I found this tool works great on my Linux Arch box. Thanks a lot for it!

One thing that worth improving is to add a UNIX man page. There is already tutorial so having the same information in man file would be very useful.

Thanks.

xor-gate:
This should be in the same category when commandline argument parsing is refactored for milestone 2.0.0. Then we can generate man pages.

xor-gate:
It seems I have broken the build on cmake 2.8. I'm currently using cmake 3.5 so I need to find out on another machine with cmake 2.8. Will fix ASAP on another machine.

xor-gate:
Unix man-pages are merged into `master`.

PR #464, opened 2016-09-15
Add Man pages

alx741:
This is the first attempt to add man pages for the utilities (#208)

Only the man page for `st-util` is ready, if this is useful I will make the pages for the other binaries as well.

The Cmake rules are modified so doing `sudo make install` installs the man pages. Then doing `man st-util` works just fine.

The makrdown man pages need to be generated first by doing `make` in the `doc/man` directory. Maybe integrate this in the Cmake rules (?)


texane:
Merged it but some modifications done. More work needed here, cf. following
commits logs.


2016-09-15 5:20 GMT+02:00 Daniel Campoverde <notifications@github.com>:

xor-gate:
@alx741  looking good, we can continue discussion in #208 if you want. I will do some work before v1.3.0 release to integrate with cmake.

alx741:
@xor-gate sure, I'll work on The other pages today

Nightwalker-87:
See also PR #467.

PR #467, opened 2016-09-15
Man pages

alx741:
- st-flash
- st-info
- st-term


xor-gate:
Looking good!

Nightwalker-87:
See also PR #464.

PR #469, opened 2016-09-16
cmake: Refactor into seperate files

xor-gate:
I have seen problems with travis infrastructure and old tools versions the git/project version gets broken. Now this is more reliable and has always a fallback to the `.version` file. When the utilities are build the version is linked into the executable and printed on startup like OpenOCD does.

@nekromant could you review this works for you, as I have moved stuff all over the place and maybe have broken debian packaging.


xor-gate:
As everybody is fearly busy and some contribution is pending for debian packaging (fixes) we need this merged. I will merge it myself for now, when there are questions feel free to place them here.

nekromant:
@xor-gate Terribly sorry, I'm currently away fishing till Wednesday-Thursday and there's literary no internet connection here (very poor cellphone coverage). I'll check everything once I'm back. (And it looks like my homeserver infrastructure along with jenkins is down due to a power outage. crap. ).

xor-gate:
Hi @nekromant yeah I just figured out a few days ago. It is no showstopper. I already have merged as someone else is improving the debian packaging. Have fun fishing!

nekromant:
@xor-gate, @OsterlaD I'm finally back. Looks like I've missed all the fun with the recent debian packaging fixes. Anyways, I've tested the current master, debian packaging and did a full ci run, everything looks fine so far. 
Regarding the recent deb packaging fixes by OsterlaD that I've missed out:
1. Do we really need to duplicate all the file installation rules in debian/rules? I suggest we do as much of the installation as possible with CMakeLists.txt, so that debian/rules would do as much as possible automatically.  
2. Is there any rationale for building both debug and release builds of stlink when packaging?

OsterlaD:
1. Use of cmake install would be better in terms of maintenance. 
    * I am not sure how to tell cmake where to install
    * adapt debian/*.install files to copy from cmake install dir, instead of from debian/tmp
    * or you can manage to install to debain/tmp with cmake
    * cmake should handle multiarch library paths

2. Just that tests depends on debug build. The build stage calls `make all` and all depends on release. The tests stage calls `make test` and test depends on debug. This is why both where build. I think automated tests where a good thing and should run, but you can disable it by add `override_dh_auto_tests:` to debian/rules, if you want.

xor-gate:
We will continue discussion in this issue:
https://github.com/texane/stlink/issues/482

PR #473, opened 2016-09-18
packaging: Improvements to debian package build.

OsterlaD:
Solution to problems with debian package build issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
 	install(FILES etc/modprobe.d/stlink_v1.conf
 	DESTINATION /etc/modprobe.d/)
 	install(FILES ${RULES_FILES}
-	DESTINATION /lib/udev/rules.d/)
+	DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/udev/rules.d/)
 endif()
 
 add_subdirectory(src/gdbserver)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.7)
 project(stlink C)
 set(PROJECT_DESCRIPTION "Open source version of the STMicroelectronics Stlink Tools")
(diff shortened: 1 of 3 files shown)


xor-gate:
You should not rely on the `build/Debug/doc/man/*.1` because this only used during building debug targets. Debian package should use only Release targets and fetch the pre-generated man-pages from `doc/man/*.1`.

xor-gate:
See here: https://github.com/texane/stlink/blob/cmake-fixes/doc/man/CMakeLists.txt#L9

xor-gate:
I have merged `cmake-fixes` into `master`, you may rebase to `master` which builds again. Travis building broke for some unknown reason and I have hardened the cmake version detection now.

OsterlaD:
Reworked patch stack on master.
Now manpages contains doc/man/st-.1 entry, corresponding to install entry usr/bin/st-.
The debian/copyright file indicates BSD-3-clause license with copyright holders from LICENSE file.

OsterlaD:
Added udev rules and modprobe installation rules.

xor-gate:
Could you squash-merge all your commits into one? Then we are ready to go!

OsterlaD:
There are still some issues.
* manpages expect to install from source, not build direcotry or debian/tmp, but the nice man pages are created from *.md files and were placed under build/Debug/doc/man
* stlink-gui requires src/tools/gui/stlink-gui.ui file and seems to expect it under /usr/local/share/stlink
* is it correct to build debug variant?

xor-gate:
I will comment inline:
> * manpages expect to install from source, not build direcotry or debian/tmp, but the nice man pages are created from *.md files and were placed under build/Debug/doc/man

Package maintainers should only use the generated manpages from `doc/man/*.1`  texane has put the missing generated roff files there in latest master. This takes away the pandoc dependency for package maintainers.

> * stlink-gui requires src/tools/gui/stlink-gui.ui file and seems to expect it under /usr/local/share/stlink

This is hardcoded into the GUI currently and it has not fixed yet.

> * is it correct to build debug variant?

The Debug target is not what you want in a package, because it contains debug symbols (GDB) and is not optimized (-O0).

Nightwalker-87:
Closed in favour of #484.

Nightwalker-87:
Closed in favour of #484.

PR #476, opened 2016-09-19
Fix broken install debian package rules.

OsterlaD:
Override dh_auto_install to use *.install files as they are.
The debian/*.install could contain source to destination format too,
but that would be much more difficult to handle with multi-arch.
Add debian copyright file.
Add stlink-tools manpages.
Insert homepage link.
Add library dependencies to binaries.
Add howto debuild to README.
Install udev rules.
Install modprobe config for usb module.
Configure cmake with /usr prefix and compile release variant.


xor-gate:
Great!, thanks for your effort.

PR #484, opened 2016-09-22
fix-482

OsterlaD:
Implement install delegate and use it in debian/rules.

Install of udev rules and modprobe config is still done in debian/rules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ MAKEFLAGS += -s
 
 all: release
 ci: lint debug release test
+install: install_release
 
 help:
 	@echo "      release: Run a release build"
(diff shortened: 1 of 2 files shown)


xor-gate:
The top level makefile should have for the install_* targets a dependency to debug/release:
When people directly type make install_debug it will fail. Also we should wrap a `install` target to `install_release`.

```
install_release: release 
...
install_debug: debug
...
```

nekromant:
@xor-gate: It looks addition of the top-level Makefile breaks automated debian, I've fixed. Please don't yet merge @OsterlaD's fix, I have a cleaner solution on the way.

OsterlaD:
Removed debian fixes.
But install delegate and documentation adaption may apply anyway.

xor-gate:
So this can be closed in favor of PR #485 ?

OsterlaD:
I think the install delegates would be useful anyway.
If you do not need them, just close the pull request.

Hopefully he gets its changes stable, it would be a clean solution.

xor-gate:
I will close this, as soon as the stuff from @nekromant has been fixed/merged. I will add install delegates to the top level makefile. Thank you for you contribution!

Nightwalker-87:
Closed in favour of #485.

PR #489, opened 2016-09-30
reset MER bit after mass erase

night-ghost:
otherwise all subsequent erases before turning off the power will be mass erases


xor-gate:
Looks good to me, but I'm wondering why the flash loader itself doesn't take care of resetting the MER flag?

night-ghost:
>why the flash loader itself doesn't take care of resetting the MER flag?

yes may be it will be better to reset this flag before ordinary erase.

xor-gate:
Sorry for the late reply, thanks!

PR #491, opened 2016-10-04
Build debian package with more use of cmake

OsterlaD:
* Make use of `dh --buildsystem cmake`, as suggested by nekromant.
* Keep install rules of module config and udev rules separate, to keep normal cmake install as documented.
* Move libstlink.so symlink to dev-package, provided by nekromant.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ MAKEFLAGS += -s
 
 all: release
 ci: lint debug release test
+install: install_release
 
 help:
 	@echo "      release: Run a release build"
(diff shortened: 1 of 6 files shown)


xor-gate:
Will this replace the work of #485?

OsterlaD:
Well, this would be my solution without changes to cmake.
If you decide that it is better to change cmake behavior, then reject this pull request.
Currently it is documented to copy module config and udev rules by hand.
If you want to change this behavior #485 is to prefer.
If you want to keep this behavior then this pull request is to prefer.

xor-gate:
I understand your concern about not touching cmake for the debian packaging. But I prefer fixing cmake for all platforms and debian packaging "at the same time". My decision is to get PR #485 ready and you can ask @nekromant if there are things you would like to see fixed if there are some thing in this PR which are not in the other.

Nightwalker-87:
Closed in favour of #485.

PR #495, opened 2016-10-08
Add option to keep RPATH in installed binaries for local installs.

dflogeras:
Useful if you don't want to install to /usr/bin or similar, so that you don't have to maintain (DY)LD_LIBRARY_PATH.  Not recommended to do this on system installs, so the default is off.  I also marked it as advanced.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,12 @@
 cmake_minimum_required(VERSION 2.8.7)
 project(stlink C)
+
+option( INSTALL_STLINK_WITH_RPATH "Keep rpath in binaries/libraries after installing.  Useful for local installs to non-system paths." OFF )
+mark_as_advanced( INSTALL_STLINK_WITH_RPATH )
+if( INSTALL_STLINK_WITH_RPATH )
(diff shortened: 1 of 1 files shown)


xor-gate:
Thank you for your contribution, I had a look at the RPATH behaviour of cmake. But what is wrong with installing to DESTDIR and adding it to PATH?

dflogeras:
PATH doesn't resolve the dynamic libraries though, you can always call binaries relatively like ./myinstall/bin/st-util, but you cannot do the equivalent for library paths without setting LD_LIBRARY_PATH.

dflogeras:
I found a simpler approach for my use case, just leave the binaries where they are in CMake's build dir, and make a couple one-time symlinks to somewhere I'd like to call them from.  That way they still have RPATH set. Simplest is best :)

xor-gate:
I must have forgotten to mention I use it this way, but my thought was you tried it and where not satisfied with this solution. I think we should document this as probably other people will use it in the same way.

Thank you for your comment, I have created issue #497 so it needs to land somewhere in the documentation when installing from source.

PR #502, opened 2016-10-16
Fix install locations for modprobe config and udev rules

c-mauderer:
I had two problems while packing stlink for Arch Linux (this package: https://aur.archlinux.org/packages/stlink-git/):

1. The file `etc/modprobe.d/stlink_v1.conf` is installed to `/etc/stlink_v1.conf`. The subdirectory `modprobe.d` is not created. I'm quite sure that this is true for other distributions too. The first patch of the patch set should solve that.

2. The udev-rules are installed to the absolute location `/lib/udev/rules.d`. In Arch Linux, `/lib` is only a link to `/usr/lib`. To create a package, I have to be able to install the files to `/usr/lib`. To fix this, I just applied the `CMAKE_INSTALL_PREFIX` to this path too. I'm not sure if this is OK for other distributions too. Would it be better to use some other variable like `CMAKE_LIBRARY_PREFIX` instead?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ install(TARGETS st-flash st-info
 if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
 	file(GLOB RULES_FILES etc/udev/rules.d/*.rules)
 	install(FILES etc/modprobe.d/stlink_v1.conf
-	DESTINATION /etc)
+	DESTINATION /etc/modprobe.d/)
 	install(FILES ${RULES_FILES}
 	DESTINATION /lib/udev/rules.d/)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.7)
 project(stlink C)
 set(PROJECT_DESCRIPTION "Open source version of the STMicroelectronics Stlink Tools")
+set(STLINK_UDEV_RULES_DIR "/etc/udev/rules.d" CACHE PATH "Udev rules directory")
+set(STLINK_MODPROBED_DIR "/etc/modprobe.d" CACHE PATH "modprobe.d directory")
 
 option(STLINK_GENERATE_MANPAGES "Generate manpages with pandoc" OFF)
 
@@ -131,9 +133,9 @@ install(TARGETS st-flash st-info
 if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
 	file(GLOB RULES_FILES etc/udev/rules.d/*.rules)
 	install(FILES etc/modprobe.d/stlink_v1.conf
-	DESTINATION /etc/modprobe.d/)
+	DESTINATION ${STLINK_MODPROBED_DIR}/)
 	install(FILES ${RULES_FILES}
-	DESTINATION /lib/udev/rules.d/)
+	DESTINATION ${STLINK_UDEV_RULES_DIR}/)
 endif()
 
 add_subdirectory(src/gdbserver)
--- a/debian/rules
+++ b/debian/rules
@@ -24,4 +24,5 @@ include /usr/share/dpkg/default.mk
 # This is example for Cmake (See http://bugs.debian.org/641051 )
 override_dh_auto_configure:
 	dh_auto_configure -- \
-    -DCMAKE_LIBRARY_PATH=$(DEB_HOST_MULTIARCH)
+    -DCMAKE_LIBRARY_PATH=$(DEB_HOST_MULTIARCH) \
+    -DSTLINK_UDEV_RULES_DIR='/lib/udev/rules.d'
--- a/doc/compiling.md
+++ b/doc/compiling.md
@@ -79,3 +79,13 @@ brew install libusb cmake
 ```
 
 Compile as described in the first section of this document.
+
+## Build using different directories for udev and modprobe
+
+To put the udev or the modprobe configuration files into a different directory
+during installation you can use the following cmake options:
+
+```
+$ cmake -DSTLINK_UDEV_RULES_DIR="/usr/lib/udev/rules.d" \
+        -DSTLINK_MODPROBED_DIR="/usr/lib/modprobe.d" ..
+```


c-mauderer:
By the way: I adapted the Arch stlink-git package to use this workaround. I'll change that back as soon as the problem is resolved here. See https://aur.archlinux.org/packages/stlink-git/

xor-gate:
I had a look, it seem legit. You are right about the first commit the path is not created. Still i'm not sure if we need to install those files, as most package maintainers will cherrypick the files themselves as this changes the system behaviour. @nekromant, @OsterlaD could you also have review?

OsterlaD:
On Ubuntu 14.04 LTS and Debian Jessie, I do not have this issue.
I guess udev will search on arch linux on /lib/udev/rules.d/ too.

Does it really not install?
I think it will install, even with the /lib -> /usr/lib link.
The link just cause that /usr/lib/udev/rules.d/ and /lib/udev/rules.d/ exist.

Maybe the path issues is a cmake issue. I use 2.8.12.2-0ubuntu3 and 3.0.2-1+deb8u1.
I acked first patch. Same behavior to me.

Second patch would cause problems, because CMAKE_INSTALL_PREFIX contains /usr prefix.
The only way I see to not influence other package builds would be a separate variable just for this purpose.

c-mauderer:
I'm not sure wether udev is patched to search in /usr/lib/udev/rules.d or if it is solved by the link.

It seems that I didn't describe my problem clearly enough. Sorry for that. I would expect that a "make install" as root would work. What I try is to create a package (or rather fix it): https://aur.archlinux.org/packages/stlink-git/

The package builds fine but can not be installed because of the location of the files. I would be able to just work around this by moving the files after installation. But it somehow feels wrong to do this instead of fixing the problem in the build system instead.

If it is OK for you, I will adapt the patch so that another name for the variable is used. I found something like `CMAKE_LIBRARY_PREFIX` in other projects. Even if the `CMAKE_` prefix would suggest this, it is not an official cmake variable: https://cmake.org/cmake/help/v3.0/manual/cmake-variables.7.html. So some other name might could be better. What do you think?

OsterlaD:
I have totally forgot about the `make install` path...
With `make install` it would be better to install udev rules to /etc/udev/rules.d and libraries to /usr/local/lib and binaries to /usr/local/bin.
We can fix up this in debian/libstlink.install.
``` patch
-lib/udev/rules.d/*.rules
+etc/udev/rules.d/*.rules lib/udev/rules.d/*.rules
```
Do you use debian rules for arch linux package generation?

nekromant:
Sorry for arriving a little bit late to join the fun ;) 
I think we should drop udev rules to /etc/udev/rules.d in all cases, including packaging. 
1. AFAIK there are a plenty of packages in debian repositories that do that already. 
2. This would work out of the box for most distros.

c-mauderer:
@OsterlaD: No I don't use the debian rules. Arch uses another package manager (pacman) and has it's own packaging rules. To create a package you more or less just write a shell script that calls `make` and `make DESTDIR=xxx install` or simmilar.

@nekromant: I would not move the rules to `/etc`. According to the description of udev (https://www.kernel.org/pub/linux/utils/kernel/hotplug/udev/udev.html chapter _Rules files_):

> The udev rules are read from the files located in the system rules directory /usr/lib/udev/rules.d, the volatile runtime directory /run/udev/rules.d and the local administration directory /etc/udev/rules.d. All rules files are collectively sorted and processed in lexical order, regardless of the directories in which they live. However, files with identical file names replace each other. Files in /etc have the highest priority, files in /run take precedence over files with the same name in /lib. This can be used to override a system-supplied rules file with a local file if needed; a symlink in /etc with the same name as a rules file in /lib, pointing to /dev/null, disables the rules file entirely.

Putting the files into `/etc` would mean that the files that are installed by the package would have the highest priority. A user would not be able to overwrite the rules (with the mentioned symlink to `/dev/null`). He would have to overwrite the rules on every reinstall.

I would prefer the mentioned variable with a sensible default value. Most distributions have their own rules where the udev rules should be put. They would be able to overwrite the variable while packing without having to track future new rules file (like for an hypothetical ST-Link version 4).

nekromant:
@c-mauderer Yep, your point is correct (I've been using Arch linux for a few years myself some time ago and I still remember some things about pacman) ;)
Since /etc/ is a user-administered dir, setting it as a default for udev rules installation would make sense for those doing 'sudo make install' installation. And for debian and arch we could override it when building package. 

But that would likely look like: 
```
set(STLINK_UDEV_RULES_DIR "/etc/udev/rules.d" CACHE PATH "Udev rules directory")
set(STLINK_MODPROBED_DIR "/etc/modprobe.d" CACHE PATH "modprobe.d directory")
...
```

c-mauderer:
I added a patch that uses the two suggested variables to the cmake configuration. Would it be necessary to adapt the debian rules too? It's been quite some time since I last touched some debian package rules.

OsterlaD:
This would be an easy way to handle it in debian package.
https://github.com/OsterlaD/stlink/commit/f2178381861bfa215f128f7c9e4d38d56c793b4b.patch

@nekromant Do you agree, or would you prefer to pass cmake variable?

nekromant:
I think passing a proper cmake variable is the proper the way to go for packaging.

xor-gate:
I agree with @nekromant we should have a cmake variable as you have patched. Probably we should have the default directory to /lib for udev and the user/sysadmin could overwrite it. I'm not sure how other distros handle this (fedora, suse et al). But probably they stick with the udev defaults.

c-mauderer:
Without any deeper Debian packaging knowledge I tried to add the necessary lines to the debian rules and install file. Could someone with deeper Debian knowledge (most likely @nekromant?) look over the patch?

OsterlaD:
Keep modprobe at etc. The udev part looks good.

c-mauderer:
Thanks for the answer @OsterlaD. I somehow was convinced that your patch moved them too. When I now look at it, I somehow have totally misread it. Sorry for that.

I reverted the modprobe configuration location and checked that the package builds. As far as I can say it should work.

nekromant:
Sorry for delays, as usual. I'll try to give it a proper spin on Monday. Currently away from my laptop.

c-mauderer:
Hello @nekromant, the delay is no problem. Take your time.

xor-gate:
@c-mauderer could you rebase with master and test. Because I needed to do some changes due to broken builds under Cygwin.

c-mauderer:
Hi @xor-gate, the rebase worked without too much of a problem. I had to move the documentation to the new file but despite that, it seems to be OK.

I used the opportunity to remove some of the intermediate stages so that a clean set of patches is left.

xor-gate:
Thank you for your work @c-mauderer. @nekromant I have squash-merged, if there are any issues we can discuss here and you are free to open a PR. Thanks all for your input on this.

c-mauderer:
@xor-gate: Thanks for merging the patches.
@all Thanks for all the help to get the patches ready to merge.

PR #505, opened 2016-10-17
flash: added easy way to reset the target

gvz:
st-flash does now have a 'reset' command
makes it easy to reset the Target

### DIFF
--- a/doc/man/st-flash.md
+++ b/doc/man/st-flash.md
@@ -32,6 +32,8 @@ read *FILE* *ADDR* *SIZE*
 erase
 :   Perform a mass erasing of the device firmware
 
+reset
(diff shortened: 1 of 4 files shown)


xor-gate:
Looks fine to me, you only forgot to update the manpage under `doc/man`. If you do that, then it can be merged.

gvz:
Thanks for your reply, I just pushed my manpage update.
I hope the shortness of the explanation is OK, but there is not much to explain in reset.

gvz:
Is there anything to be done before the PR can be merged?

xor-gate:
My excuse, I was very busy last weeks so it has been moved out of my mind. Looks good to me. Thanks!

PR #526, opened 2016-12-11
Update README.md

PeterMortensen:
Copy edited (e.g. ref. <http://en.wikipedia.org/wiki/Debian> and <http://en.wikipedia.org/wiki/Linux>)


xor-gate:
Great catch!

PR #534, opened 2016-12-29
Set SWDCLK and fix jtag_reset bug

jkhax0r:
- Fixed bug where stlink_jtag_reset in stlink_open_usb() was never being called because the STLINK version was being checked before it had been initialized
- Added support for stlink_set_swdclk() to adjust the SWD clock speed.  For example, to set at 4Mhz, 1.8Mhz, 900Khz, etc similar to in the official ST-LINK utility.  NOTE: The default when STLINK is powered appears to be 1.8Mhz (looked at scope traces) but it retains whatever was set before.


xor-gate:
Thank you for this feature and fix!

PR #538, opened 2017-01-02
add support for STM32F412

Androude:
General support for STM32F412 devices.
Added chipid, device params and fixed condition in flash_loader.c

Checked on 32F412GDISCOVERY kit board.


xor-gate:
Looks good, the build is broken (by me). Thank you for your contribution!

Issue #318, opened 2015-06-30
Probe and control multiple connected programmers


trichner:
I have multiple STM32F0Discovery Boards wired up to an USB hub and would like to specify which one to flash (I'm using `st-flash main.bin 0x800000` right now).

Is there a way to specify the USB port to use? or even an actual selection?

texane:
Hi,

you can use the STLINK_DEVICE environment variable
to specify which device to use. VID:PID format.

2015-06-30 12:15 GMT+02:00 Thomas <notifications@github.com>:

trichner:
But all stlink programmers have the same VID:PID, they only differ in Bus/Device

Example `lsusb` dump:
```
Bus 002 Device 012: ID 0483:3748 STMicroelectronics ST-LINK/V2
...
Bus 001 Device 014: ID 0483:3748 STMicroelectronics ST-LINK/V2
```

trichner:
Well, if anybody ever has the same need and is looking for a solution: `OpenOCD` can use multiple devices/programmers by using the `hla_serial` command.

shkhaliq:
Any updates on this issue?

shkhaliq:
I got a workaround. Special thanks to @xor-gate for the `st-probe` tool. Use that tool to get the serial number.
 Note the last byte is a null character, So if the serial number is `483f6e06703f53483324096700` use `483f6e06703f534833240967`.

One of the caveats of getting the serial number is that you have to make sure the programmer is connected to the board and its in a connectable state. 

After that just invoke `st-flash --serial <serial_no> <your_command_here>`.

xor-gate:
@shkhaliq could you post your st-probe output, maybe it prints a spurious 00 as last byte? It seems I also have the same which adds the extra 00 byte:

```
jjacobs@workbench ~/repos/github/stlink/build $ ./st-probe
Found 2 stlink programmers
563f6706677256563113066700
         flash: 262144 (pagesize: 2048)
          sram: 65536
        chipid: 0x0414
         descr: F1 High-density device
493f6d06667048513615106700
         flash: 1048576 (pagesize: 16384)
          sram: 196608
        chipid: 0x0413
         descr: F4 device
```

xor-gate:
For your convenience `st-probe` has been merged into `st-info` -> `st-info --probe` and the hla serial string is also printed. When only one programmer is attached you can use `st-info --hla-serial`. Have fun! Maybe this issue can be closed then @trichner ?

trichner:
[...]

PR #541, opened 2017-01-10
serialnumber support  st-util and st-flash

GagoSoft:
modified serialnumbersupport in st-flash:
  in flash_opts take care of the length of the given serialnumber (was fix 12 chars)
added serialnumber support in st-util:
  implemented like in st-flash.


xor-gate:
Thank you for your pull request, is this all the only code necessary to support multiple programmers issue #318?

GagoSoft:
I think, everything is done. I will start use (and therefore test) the Multi-ST-Link-System

xor-gate:
Thanks!

PR #546, opened 2017-01-27
Semihosting: Add support of SYS_READC

Fabien-Chouteau:


src/gdbserver/semihosting.c
```diff
@@ -457,6 +457,13 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         }
         break;
     }
+    case SYS_READC:
+    {
+        uint8_t c = getchar();
+        *ret = c;
+        return 0;
+        break;
```
potozi: will be never executed

xor-gate: Yes, break is unnessary here.

Fabien-Chouteau: And that's why you don't push a commit Friday evening... :confused:

Fabien-Chouteau:
Updated.

xor-gate:
Thanks!

PR #559, opened 2017-02-18
Usb error more informative to user. Refactored vartypes of device's serial.

theg4sh:
A little change to inform user about problems during device search or open.


src/usb.c
```diff
@@ -777,37 +777,47 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, bool reset, char serial[16
...
         libusb_get_device_descriptor( list[cnt], &desc );
-        if (desc.idVendor != STLINK_USB_VID_ST)
-            continue;
-
         if (devBus && devAddr) {
             if ((libusb_get_bus_number(list[cnt]) != devBus)
                 || (libusb_get_device_address(list[cnt]) != devAddr)) {
                 continue;
             }
+            if (desc.idVendor != STLINK_USB_VID_ST) {
```
xor-gate: Why are we warning vendors other than ST, this seems useless.

src/usb.c
```diff
@@ -777,37 +777,47 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, bool reset, char serial[16
...
+                continue;
         }
 
         if ((desc.idProduct == STLINK_USB_PID_STLINK_32L) || (desc.idProduct == STLINK_USB_PID_STLINK_NUCLEO)) {
             struct libusb_device_handle *handle;
 
             ret = libusb_open(list[cnt], &handle);
-            if (ret)
-		continue;
+            if (ret) {
```
xor-gate: Better check against legit libusb return code macros instead of != 0. The return code can actual be printed as a string (look it up in the libusb api).

src/usb.c
```diff
@@ -777,37 +777,47 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, bool reset, char serial[16
...
+                WLOG("Cannot open device %04x:%04x on %03d:%03d. %s\n", desc.idVendor, desc.idProduct,
+                        libusb_get_bus_number(list[cnt]), libusb_get_device_address(list[cnt]),
+                        strerror(errno));
```
xor-gate: strerror is not valid in this context, libusb emits its own errors independent of errno.

src/usb.c
theg4sh: devBus and devAddr is set when you declare STLINK_DEVICE environment variable.
So, if you set it and made a mistake then you get nothing useful in output.
Only one line will be printed (it a little bit confused me at the beginning):
```
$ STLINK_DEVICE=007:003 ./st-util
2017-02-18T23:32:14 INFO /home/user/repos/stlink/src/usb.c: bus 007 dev 003
$
```

src/usb.c
theg4sh: @xor-gate, as it was in the code after 'while'-loop, I used strerror.
Fixed that in the new commits.

src/gdbserver/gdb-server.c
```diff
@@ -38,7 +38,7 @@
 static stlink_t *connected_stlink = NULL;
 static bool semihosting = false;
 static bool serial_specified = false;
-static char serialnumber[28] = {0};
+static uint8_t serialnumber[STLINK_SERIAL_SIZE] = {0}; // XXX? Why was size 28?
```
xor-gate: Be carefull with this. The discovery boards (stlinkv2) have a binary serial number (12bytes). The nucleo boards (stlinkv2-1) have a hex encoded serial number (probably 24 or 25 bytes not 100% sure right now).

theg4sh: Got it, thanks.
Anyway, in code is've seen different serial sizes - from 16 to 28... but seems all of them is incorrect :)

I've found the following document http://www.st.com/resource/en/application_note/cd00169020.pdf
where described that serial could have custom length (didn't said maximum exactly) and ascii representation can be greater than 28, e.g. on my STM32F407G-DISC1, the key have 48 bytes in ascii representation and struct stlink_t with buffer serial[16] (as it was) can't contain all data of libusb_get_string_descriptor_ascii call in stlink_open_usb. I'm working on it.

xor-gate: I think the AN you reference is unrelated to serial numbers from the programmers as this is a note for manufacturing serial number programming. Probably the programmers use the STM32 96-bit unique ID as serial number (12 bytes binary, 24 bytes hex encoded).

I prefer to have all serial number code centralized in a few functions and use it in the different tools. Currently this is a mess as you have seen.

xor-gate:
I'm closing this PR as it is not 100% what we want because there is a mixture of hex/binary serial numbers and needs to be resolved in a generic way by writing a hex2bin/bin2hex implementation or using a existing one (e.g https://raw.githubusercontent.com/xor-gate/stlink2/master/src/utils/hexstr.c).

theg4sh:
@xor-gate, got it.
I did the same conclusion earlier, so, that's why didn't push any commits.

Could you, please, take a look for branch https://github.com/theg4sh/stlink/tree/libusb_minimize
There I did changes with serial, so, it structured and always converted to binary and then compared.
Functions: _stlink_serial_convert_, _stlink_serial_init_, _stlink_serial_clear_.

Also, I found that all usb devices walked by 3 times per tool run, e.g. 'st-info --probe', so, did change regarding this - in branch stlink devices initialized once and then used prepared pointer to get next stlink device.
Functions used to iterate stlink devices: _stlink_init_devices_, _stlink_device_next_, _stlink_devices_rewind_ and _stlink_free_devices_.
So, it could be possible to clear sources from non-DRY code.
Please, note, that branch is not clear now.

In other hand, devices iterated for both versions - v1 and v2. I guess, it could be good for DRY, but stlinkv1 uses similar, but not the same backend callbacks.
I want to know your opinion regarding that modifications and will continue to work on it if you approved it.

xor-gate:
Looks good! Why you don't convert directly after a programmer instance is created so it is directly usable without the tools/apps needing to convert themself. Do we really need this complex thing:

```
```

xor-gate:
+    enum stlink_serial_format_type {
+        STSERIALF_UNKNOWN=0,
+        STSERIALF_BINARY,
+        STSERIALF_HEX,
+        STSERIALF_ASCII
+    };
```

I think preferred would always to make sure it is ascii (or ascii hex).

PR #560, opened 2017-02-20
Fix incorrect release folder name in docs

joostrijneveld:
This fixes a minor inconvenience in the documentations w.r.t. compiling from source; the directory where the build ends up has changed.


xor-gate:
Nice catch, thanks for fixing up!

PR #561, opened 2017-02-21
Fix compilation when path includes spaces

joostrijneveld:
Currently compilation of `release` would break when the path to `stlink` included spaces. There seemed to be no proper way to fix this by escaping within `COMPILE_FLAGS`, but since CMake 3.0 there is a `COMPILE_DEFINITIONS` property that handles escaping itself.

### DIFF
--- a/src/tools/gui/CMakeLists.txt
+++ b/src/tools/gui/CMakeLists.txt
@@ -9,13 +9,13 @@ include_directories(SYSTEM ${gtk_INCLUDE_DIRS})
 
 add_executable(stlink-gui-local ${GUI_SOURCES})
 set_target_properties(stlink-gui-local PROPERTIES
-                      COMPILE_FLAGS -DSTLINK_UI_DIR=\\"${CMAKE_CURRENT_SOURCE_DIR}/gui\\")
+                      COMPILE_DEFINITIONS STLINK_UI_DIR="${CMAKE_CURRENT_SOURCE_DIR}/gui")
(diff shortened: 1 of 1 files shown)


xor-gate:
Hi @joostrijneveld, I understand your concern about this problem. Only I want people give the room to use older cmake versions. CMake 3.0 is release 3 years ago. 

https://blog.kitware.com/cmake-3-0-0-available-for-download/

joostrijneveld:
I see your concern.. It's not entirely clear to me if this strictly requires CMake 3.0 though, as I'm now also finding references to this in [older CMake documentation](https://cmake.org/cmake/help/v2.8.8/cmake.html). I'll spin up a VM and give it a test..

xor-gate:
Most systems come with CMake 2.6 or higher. Ubuntu LTS comes with 2.6, Debian oldstable with 2.8, Debian stable with 3.0.2. And I want also support older BSDs.

joostrijneveld:
I didn't realise 3.0 was that recent, but it appear it's not a problem after all. `COMPILE_DEFINITIONS` works just fine on older CMake; it's just that alternative `COMPILE_DEFINITIONS_<CONFIG>` was deprecated in favour of it.

To verify, I spun up a fresh Debian Wheezy VM, installed the requirements, and compiled from this branch. I'll amend my commit accordingly.

xor-gate:
Great! This looks good.

PR #569, opened 2017-03-14
Fix stlink on mingw64

oli-obk:
`%zx` gets printed as `zx`. Just using `%x` suffices in the cases in this PR.

This fixes stlink for us.


xor-gate:
Great, thanks for the improvement!

xor-gate:
Sorry reverted in PR #573

oli-obk:
Hmmm... That's unfortunate... Isn't there a platform independent formatting? We could cast it to something that works everywhere.
Also, why wasn't this caught by Travis?

xor-gate:
yes, `%zu` is in the C99 standard. but .... mingw64 wraps windows printf which doesn't support it :(
http://stackoverflow.com/a/22114959/1836746

oli-obk:
Would you accept a PR turning formatting warnings into errors on Travis and casting all size_t formatting to unsigned long long before formatting?

xor-gate:
Sure, just make a proposal!

PR #571, opened 2017-03-15
Fix possible memory leak

### LINKED ISSUE #570: Possible memory leak on libusb_open errors
In function [_stlink_open_usb_](https://github.com/texane/stlink/blob/master/src/usb.c#L761) there is a possible memory leak when the call to _libusb_open_ in line [854](https://github.com/texane/stlink/blob/master/src/usb.c#L854) fails. If _ret_ is not zero, the code jumps to the _on_error_ label and _list_ is never freed.

A possible fix would be to free _list_ in case of an error, or to check if _list_ in the code that is exceuted when jumping to the _on_error_ label.

thebaron06:
fixes #570


xor-gate:
Nice catch, thanks!

xor-gate:
Ignore the revert commit, github doesn't understand itself.

Issue #564, opened 2017-03-02
STM32L011G4 flash error (unknown coreid, write_half_pages failed)


zakqwy:
- [x] Programmer/board type: Stlink/v2-onboard (STM32F030 Discovery Board)
- [X] Operating system: Linux Mint 17.2
- [X] stlink version: v1.3.1/git-6f24bde
- [X] stlink commandline tool name: st-flash
- [X] Target chip: STM32L011G4

I am attempting to flash a simple 'hello-world' type LED blink program onto a bodged together STM32L011G4-based board, using my STM32F030 Discovery Board's on-board Stlink/v2. 

Command: `$ make helloworld.stlink-flash`

Output:

`  FLASH  helloworld.bin`
`st-flash 1.3.1`
`2017-03-02T16:27:46 INFO src/common.c: Loading device parameters....`
`2017-03-02T16:27:46 INFO src/common.c: Device connected is: L011 device, id 0x10086457`
`2017-03-02T16:27:46 INFO src/common.c: SRAM size: 0x800 bytes (2 KiB), Flash: 0x4000 bytes (16 KiB) in pages of 128 bytes`
`2017-03-02T16:27:46 INFO src/common.c: Attempting to write 480 (0x1e0) bytes to stm32 address: 134217728 (0x8000000)`
`Flash page at addr: 0x08000180 erased`
`2017-03-02T16:27:46 INFO src/common.c: Finished erasing 4 pages of 128 (0x80) bytes`
`2017-03-02T16:27:46 INFO src/common.c: Starting Half page flash write for STM32L core id`
`2017-03-02T16:27:46 ERROR src/flash_loader.c: unknown coreid, not sure what flash loader to use, aborting! coreid: bc11477, chipid: 457`
`2017-03-02T16:27:46 WARN src/flash_loader.c: Failed to write flash loader to sram!`
`2017-03-02T16:27:46 WARN src/common.c: stlink_flash_loader_init() == -1`
`2017-03-02T16:27:46 WARN src/common.c: `
`write_half_pages failed == -1`
`  2/  3 pages written`
`2017-03-02T16:27:46 INFO src/common.c: Starting verification of write complete`
`2017-03-02T16:27:46 ERROR src/common.c: Verification of flash failed at offset: 0`
`stlink_fwrite_flash() == -1`
`make: *** [helloworld.stlink-flash] Error 255`

[...]

PR #572, opened 2017-03-15
fixed support for STM32L011 (#564/#565)

zakqwy:
Updated `chipid.c` to match RM0377 (pg 52-53).
Added STM32L011 to flash address if statement.


xor-gate:
Ignore the revert commit, github doesn't understand itself.

PR #575, opened 2017-03-19
Added -chipid=xxxx to st-flash command line for F103 128k support

ve3wwg:
In order to take advantage of the 128k flash that seemingly most stm32f103c8t6 chips support, I added command line --chip=xxxx to allow forcing the chip id to 0x414 (STLINK_CHIPID_STM32_F1_HIGH). In addition to changing the chipid in flash.c, I also upped the sl->flash_size to 128k when the chipid was STLINK_CHIPID_STM32_F1_HIGH.

Warren

### DIFF
--- a/include/stlink/tools/flash.h
+++ b/include/stlink/tools/flash.h
@@ -4,6 +4,8 @@
 #include <stdint.h>
 #include <stlink.h>
 
+#include <stlink/chipid.h>
+
 #define DEBUG_LOG_LEVEL 100
 #define STND_LOG_LEVEL  50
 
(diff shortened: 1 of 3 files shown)


xor-gate:
I'm not against forcing a chipid, but reading the flash size register is probably the better approach. There are other STM32F103 variant which have up to 1M flash and then this will not work.

ve3wwg:
Ok, allow me to test a change where I just override the flash size. If that works, then the --chipid= isn't required. I just assumed that the programming method might change with the increased size.

xor-gate:
I have a list of flash size regs here:
https://github.com/xor-gate/stlink2/blob/master/include/stlink2/stm32.h

Which where borrowed from:
https://github.com/pavelrevak/pystlink/blob/master/lib/stm32devices.py

ve3wwg:
I'm going to resubmit based upon new changes. Closing this request.

PR #576, opened 2017-03-20
Added --flash=n[k][m] command line option to override device model

ve3wwg:
By default, st-flash works the same as before (there is no more chipid override in this pull request).

Now _optionally_, you can specify **--flash=128k** for example, to override the stm32f103c8t6 to assume 128k of flash instead of the default of 64k. This option accepts decimal (128k), octal 0200k, or hex 0x80k. Obviously leaving the multiplier out is equally valid, for example: --flash=0x20000

The size may be followed by an _optional_ "**k**" or "**m**" to multiply the given value by 1024 or 1 Meg respectively.

Tests on the stm32f103c8t6 confirm that overriding the flash size is all that is required. I was still able to flash and read 128k without changing the chipid. I don't know if this will be true of all other devices but I suspect that they will.

### DIFF
--- a/include/stlink/tools/flash.h
+++ b/include/stlink/tools/flash.h
@@ -4,8 +4,6 @@
 #include <stdint.h>
 #include <stlink.h>
 
-#include <stlink/chipid.h>
-
 #define DEBUG_LOG_LEVEL 100
 #define STND_LOG_LEVEL  50
 
@@ -22,7 +20,7 @@ struct flash_opts
     int reset;
     int log_level;
     enum flash_format format;
-    uint32_t flash_size;	/* --flash=n[k][m] */
+    size_t flash_size;	/* --flash=n[k][m] */
 };
 
 #define FLASH_OPTS_INITIALIZER {0, NULL, {}, NULL, 0, 0, 0, 0, 0, 0 }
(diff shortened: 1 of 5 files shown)


include/stlink/tools/flash.h
```diff
@@ -20,9 +22,10 @@ struct flash_opts
     int reset;
     int log_level;
     enum flash_format format;
+    uint32_t flash_size;	/* --flash=n[k][m] */
```
xor-gate: you could better use size_t here, as we don't map it back to the microcontroller

include/stlink/tools/flash.h
```diff
@@ -4,6 +4,8 @@
 #include <stdint.h>
 #include <stlink.h>
 
+#include <stlink/chipid.h>
```
xor-gate: I'm not 100% sure but probably this is not necessary

xor-gate:
I also recommend updating the manpage source with a small description about this feature:
https://github.com/texane/stlink/blob/master/doc/man/st-flash.md

ve3wwg:
The suggested changes have been added to this request.

Thanks, Warren

ve3wwg:
Sorry, one more change added to rm old ref to chipid in the flash.c usage display.

xor-gate:
Looks good, finished and ready to go?

xor-gate:
ping @ve3wwg

ve3wwg:
Hi Jerry:

I just assumed you went ahead. Yes, all seems good to me. It will be nice
to have this feature included as standard equipment.

Thanks, Warren

PR #578, opened 2017-03-24
Don't use `%zx` because it's unsupported on mingw64

oli-obk:
Also check that no warnings happen on travis

cc #569


xor-gate:
Looks good i'm only not 100% sure if clang or other strict compiler will eat this. You should prefer adding a size to the cast so:

`(unsigned)` becomes `(unsigned int)`

Then we are good to go.

oli-obk:
Ok... so after this PR:

* travis will check gcc and clang on both linux and os-x
* all compilations must be warning free
* O2 is used instead of O3 optimizations (this was intended by the `CMakeLists.txt`, but never actually called due to `elseif()` not doing the same as `else()`

> i'm only not 100% sure if clang or other strict compiler will eat this.

clang accepted it, but I still made the change.

Maybe there's a warning we can enable to enforce this?

xor-gate:
Thank you for this improvement and understanding!

xor-gate:
I'm not sure, probably GCC or some analyser have warned me some day about no defined typed, assuming int.

PR #581, opened 2017-03-30
Fix installation path for shared objects

karol-szuster:
There is no way to reliably configure installation path for shared
objects.  It looks like this should be configurable by setting
CMAKE_LIBRARY_PATH but this approach doesn't work as intended.
Moreover, according to cmake docs, purpose of this variable is slightly
different.
With this patch it should be possible to configure where all libs should
go by setting STLINK_LIBRARY_PATH to either /usr/lib or /usr/lib64
(/usr/lib64 is new default value assuming that most users run 64bit
systems).


xor-gate:
This will break on non-linux systems, MinGW, Cygwin, OS X, BSD etcetera. Probably the library path should be injected during build by commandline arguments by distro maintainers instead of hardcoding /usr/lib64.

karol-szuster:
I have updated this pull request. It should now behave exactly as before with only one small change. Package maintainers should now be able to override library path using STLINK_LIBRARY_PATH variable.

xor-gate:
Looks good, but probably you should use `option()` for this.

xor-gate:
Thank you!

PR #582, opened 2017-04-02
fix few -Wformat warnings

slyshykO:


xor-gate:
Thank you for your contribution, was some time away from keyboard to review this correctly!

PR #588, opened 2017-05-05
skip GTK detection if we're cross-compiling

martin-kaiser:
It seems that stlink cannot use CMake's standard mechanism to find GTK3.
Cmake's GTK3 detection supports only Unix at the moment.

Therefore, stlink uses pkg-config to find GTK3. The downside of this is
that Cmake's cross-compiler settings are not taken into account. Even if
we're cross-compiling, CMake will find a local installation of GTK3 and
compilation of the GUI fails.

As a simple fix, we skip the GTK3 detection if we're cross-compiling.

(In the long run, we should use a more advanced mechanism to find GTK3.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ include(cmake/CFlags.cmake)
 # Dependencies
 ###
 find_package(LibUSB REQUIRED)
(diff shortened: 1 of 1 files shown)


xor-gate:
Thanks!

PR #592, opened 2017-05-10
Added support for flashing second bank on STM32F10x_XL

bmellstrom:
Some of the chips in the STM32F10x series has two separate flash banks. The first is always at most 512kB and the size of the second can be 0, 256 or 512 kB. For example, the STM32F103RF has 512 + 256 kB of flash memory. To be able to flash the second you need to use another set of registers instead of the usual SR/CR/AR registers. I made a patch for this that works for me(tm). Let me know if I should change anything; would be nice to get this included in upstream!

### DIFF
--- a/include/stlink.h
+++ b/include/stlink.h
@@ -83,7 +83,8 @@ extern "C" {
         STLINK_FLASH_TYPE_F0,
         STLINK_FLASH_TYPE_L0,
         STLINK_FLASH_TYPE_F4,
-        STLINK_FLASH_TYPE_L4
+        STLINK_FLASH_TYPE_L4,
+        STLINK_FLASH_TYPE_F1_XL,
     };
 
     struct stlink_reg {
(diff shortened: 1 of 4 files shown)


xor-gate:
@texane could you also have a look and approve the review?

xor-gate:
Thank you!

bmellstrom:
Wohoo!

xor-gate:
@bmellstrom @skalidindi3 had some problems with F0 target after this patch. Could you test out latest master works with STM32F10x_XL on second bank?

bmellstrom:
@xor-gate yes, retested successfully - it works.
There is something fishy going on in the code there though. If you check the loader code for the F0 it always set's the PG bit by itself anyway, so it shouldn't be needed there. At first I thought the same loader was used for the F1(XL) devices (which would make sense since they have more or less the same interface), but then I realized the loader_code_stm32vl that was actually used, and I'm not sure what it's actually doing. It must be setting the PG bit in the correct CRx register since otherwise the flashing cannot succeed, and since it _does_ succeed flashing both banks I was thinking that meant the initial call to set_flash_cr_pg() would be pointless. I was about to disassemble the loader and double-check what it's actually doing, but never got around to it. Anyway, I'll let you know if I come up with something better!

PR #602, opened 2017-06-09
Stop the sulk with GPL2016 on Windows

orinem:
Trying to use stlink with GNAT/GPL2016/GPS was an exercise in frustration on Windows (no, another OS is **NOT** an option).  GPS would sulk trying to flash/debug.  However, running st-util then flash/debug would work, but one copy of st-util would crash.

These changes result in an st-util.exe that builds in Visual Studio 2017 (**don't complain**, it's free for open source development) and **does not** exhibit the above problems.  However, it is far from complete.  It requires that libusb be built from source in Visual Studio 2017 and the resulting binaries copying into the build folder.  It's beyond my current inclination to make cmake do that.  It's likely that a previous version of Visual Studio will work fine with the libusb that is downloaded.  I have no time to check that either.

The critical commit is probably f881677.  The code was using the wrong function to close a socket and furthermore, subsequent code that used the correct function closed the incorrect socket.

Feel free to cherry pick from these commits (especially ignoring the VS2017 project files that are particularly messy - cmake should be able to make them now).


xor-gate:
That's great news, I would prefer generating the MSVC 2017 project using CMake. I'm not a big fan of having for every IDE and compiler version different projects. Have you tried this ? I will review your code changes ASAP.

orinem:
On Fri, Jun 9, 2017 at 12:47 AM, Jerry Jacobs <notifications@github.com>
wrote:

> That's great news, I would prefer generating the MSVC 2017 project using
> CMake. I'm not a big fan of having for every IDE and compiler version
> different projects. Have you tried this ? I will review your code changes
> ASAP.
>

Generating the VS2017 project files sort of works - I generated them with
"cmake .." in the build directory.  They get regenerated all the time and
VS2017 seems happy with them, so the ones I committed in my fork are best
ignored.  The main problem at the moment is that the libusb that gets
downloaded is built for a previous version of MSVC so you get undefined
symbols until you build libusb with VS2017 and copy the relevant files into
the build tree.

It's a work in progress to say the least.  I do not expect it to be taken
in its entirety - especially the random getopt and unistd.h I had to add.

xor-gate:
Yeah I know how it works, it costs some effort to get it working. But happily you have it sort of compiling and working.

orinem:
These commits can be ignored: 3206c29, c46b3f0 and 8dae077 - they just clutter up the history.  I'd force a rebase, but it's a bit late for that.

xor-gate:
No problem we just squash them when finished.

orinem:
At this point, the MSVC projects generated by cmake work fine with VS2017 as long as the libusb archive has libs built with VS2017.  I suspect everything would be fine on a previous version of VS as well, but haven't tried it.

I downloaded the latest libusb source, built the VS2015 solution using VS2017 (works fine) and copied libusb-1.0.lib from the libusb-1.0.21\Win32\Debug\lib directory into libusb-1.0.21.7z\MS32\static using 7-Zip.  Then, I cleaned out the stlink/build directory, copied the modified libusb-1.0.21.7z in to the build directory, ran cmake and build the resulting solution (Debug target).  This sequence produced an st-util.exe that I put in my GNAT/2016/bin directory.

Debugging under GNAT/GPS is working fine, so I'll leave this branch alone now.

src/getopt/getopt.h
```diff
@@ -0,0 +1,180 @@
...
+   The GNU C Library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public
+   License along with the GNU C Library; if not, write to the Free
+   Software Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA
+   02111-1307 USA.  */
+
```
xor-gate: stlink is staticly compiled and mixing BSD and GPL code as far as I know is not allowed. See: http://www.softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html

> The GPL is a copyleft license; it formally requires that modified versions of GPL’d programs be distributed under the terms of the GPL.

xor-gate: We better could use a BSD licensed getopt from e.g https://github.com/kimgr/getopt_port

orinem:
Actually, it's LGPL like libusb, so I thought it should be OK.

However, I agree it would be better to use a BSD licensed getopt, so I'll
make the change.

orinem:
I copied the suggested BSD licensed files in for getopt.  I also made the links to libusb dynamic for MSVC (it's no longer necessary to modify the libusb archive with this change).

xor-gate:
Great, looks good to me. A few things left:

* Could you write a section about MSVC 2017 usage/compilation in document https://github.com/texane/stlink/blob/master/doc/compiling.md
* (optional, can be a hassle) Could you try to let it compile with AppVeyor continuous integration with the MSVC 2017 image (I only own Mac and *Nix machines), this should also be able to create builds for releases which are reproducable.
* There is no way to have the executables static for windows now as you have it pointing to the DLL folder, maybe we should find a way to have it static compiled (which makes distribution easier).

xor-gate:
@texane can we finally enable AppVeyor Github integration for this project repository? Just like Travis continuous integration. This makes release/artifacts for windows with MSVC/MinGW compiled much easier to maintain for me because I only own Mac and *Nix machines.

orinem:
On Sat, Jun 10, 2017 at 1:12 PM, Jerry Jacobs <notifications@github.com>
wrote:

> Great, looks good to me. A few things left:
>
>    - Could you write a section about MSVC 2017 usage/compilation in
>    document https://github.com/texane/stlink/blob/master/doc/compiling.md
>
>
Will do.


>
>    - (optional, can be a hassle) Could you try to let it compile with
>    AppVeyor continuous integration with the MSVC 2017 image (I only own Mac
>    and *Nix machines), this should also be able to create builds for releases
>    which are reproducable.
>
>
I'll look into it - never done it before.

I actually tested it in a 64bit Windows 10 Virtual Box virtual machine
under Centos Linux!  (I also tested on a 32bit Windows 7 system, so I got
reasonable coverage.  I didn't, but suppose I could build it for XP if the
free VS2017 supports that; I know the full version I have at work does.
There isn't be anything that requires later than XP that I've noticed.)


>    - There is no way to have the executables static for windows now as
>    you have it pointing to the DLL folder, maybe we should find a way to have
>    it static compiled (which makes distribution easier).
>
>
Unfortunately, after you mentioned the GPL problems with the getopt from
libusb, I checked libusb and it's LGPL with no exclusion that I could see
for static linking.  I could be wrong about that, but I switched it to
dynamic linking just in case.  It actually makes the VS C runtime issues go
away, at the cost of having to distribute the libusb dll.

xor-gate:
Thanks for looking into it! Keep me posted.

orinem:
The existing AppVeyor script fails: "mv: cannot move `mingw32' to `/MinGW/mingw32': Permission denied" for x86 and "error: 'usleep' is deprecated" for x64.

I did have success from the AppVeyor web UI.  This exported appveyor.yml successfully builds the project under VS2017.  There are a lot of warnings that should be reviewed, mostly signed/unsigned conversions.  Log is attached.

[log-2.txt](https://github.com/texane/stlink/files/1066278/log-2.txt)


version: 1.0.{build}
image: Visual Studio 2017
configuration: Debug
before_build:
- cmd: >-
    mkdir build

    cd build

    cmake ..
build:
  project: build/stlink.sln
  verbosity: minimal

xor-gate:
Nice work @orinem ! Its possible the mingw32 stuff is broken. Looks good, last note: why are there two unistd.h compat files? Probably we should just keep one under src or in include. When you are finished we can merge!

orinem:
Two unistd.h files was an accident.  I've deleted the extra.

src/getopt/LICENSE.txt
```diff
@@ -0,0 +1,24 @@
+Copyright (c) 2012, Kim Gr�sman
```
xor-gate: Last comment before merge, it seems there is something wrong after copying.

orinem:
The file had "ä" and "Ä" characters in a character set encoding incompatible with UTF8.  I resaved the file from vi as utf8 and the latest diff looks OK.

xor-gate:
Sorry for my late reply. All set!

orinem:
Cool.

Hopefully it proves useful.

xor-gate:
Sure, since I build and released a windows executable there are many downloads.

PR #605, opened 2017-06-17
Fix wrong counting when flashing

kilroy42:
When flashing STM32F1 models, the counting is off by one (e.g. when flashing 8188 bytes it counts from "0/7 pages written" to "7/7 pages written". This small patch fixes this (1/8 --> 8/8)


xor-gate:
LGTM, thanks!

PR #611, opened 2017-06-29
use correct macros for switch usleep realisation

slyshykO:
Sorry, I made silly a copypaste error in a previous pull request.


xor-gate:
No problem

PR #615, opened 2017-07-14
Add support of STM32L496xx/4A6xx devices

homer242:
Hi,

I quickly added the support of my STM32L496G (discover) in stlink.

I can read and write data to the micro-controller memory. I think all is good.

Before write to the flash, I do a mass erase otherwise it fails with this message:

> 2017-07-14T15:48:32 ERROR src/common.c: Verification of flash failed at offset: 6144

Regards,
Anthony V.

### DIFF
--- a/include/stlink/chipid.h
+++ b/include/stlink/chipid.h
@@ -36,9 +36,11 @@ enum stlink_stm32_chipids {
 	STLINK_CHIPID_STM32_F4_DSI           = 0x434,
 	/*
 	* 0x435 covers STM32L43xxx and STM32L44xxx devices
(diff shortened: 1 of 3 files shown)


xor-gate:
Thank you, you should associate your email adres (`...@syscom-instruments.com`) with your account which makes it easier for other people to look you up after merge.

rdlim:
Hi, thanks for adding support for STM32L496x. I have seen the similar "verification of flash" error on my STM32L496 Nucleo-144 device. This is the error.

> 2017-07-14T22:00:04 ERROR src/common.c: Verification of flash failed at offset: 4096

I have the fix for it and will file an issue and pull request.

Issue #617, opened 2017-07-15
ERROR src/common.c: Verification of flash failed at offset: 4096


rdlim:
**NOTICE: The issue may be closed without notice when not enough information is provided!**

Thank you for giving feedback to the stlink project. Take some time to fill out
 check boxes with a X in the following items so developers and other people can try to
 to find out what is going on. And add/remove what is appropriate to your problem.

When submitting a feature request, try to reuse the list and add/remove what is appropriate.
 Place a `X` between the brackets `[X]` to mark the list item.

- [X] Programmer/board type: Stlink/v2-onboard
- [x] Programmer firmware version: STSW-LINK007 2.27.15
- [X] Operating system: Ubuntu 14.04 LTS
- [X] Stlink tools version and/or git commit hash: v1.4.0-2-g0c84c29
- [X] Stlink commandline tool name: st-flash
- [X] Target chip (and optional board): e.g STM32L496ZGT6 Nucleo-144

A as-detailed description possible of the problem with debug output when available.

When flashing, I occasionally get the following error message when I make a slight firmware change and reflash. The work-around is doing a mass erase before doing st-flash write

OUTPUT/ERROR of the commandline tool(s)
```
2017-07-14T22:00:04 ERROR src/common.c: Verification of flash failed at offset: 4096
stlink_fwrite_flash() == -1
```

Expected/description:
`st-flash write should work `

Thank you,
The stlink project maintainers

xor-gate:
Fix is merged to master.

PR #618, opened 2017-07-15
Fix verification of flash error for STM32L496x device (#617)

rdlim:
There were missing chipid checks of STM32L496xx and STM32L46x devices.


xor-gate:
Thanks!

homer242:
Hi rdlim,

Thanks for this fix. I am happy it will solve the problem I see with my STM32L496G discovery board while I am trying to write a new firmware on it. I will test it tomorrow.

Regards,
Anthony.

homer242:
Hi rdlim,

The fix solves the write problem for me. Thanks !

Regards,
Anthony.

PR #635, opened 2017-09-07
Added information about Fedora and RedHat/CentOS packages.

Vascom:
I am just added stlink package to Fedora and EPEL repositories. It available now for Fedora Rawhide, after 7-8 days other Fedora and after 14-15 days for EPEL.

If someone need stlink for RedHat/CentOS 6 it may be done later and without GUI because GTK 3 not available.


xor-gate:
Thank you

Issue #633, opened 2017-09-04
On 64-bit system libs installed to /usr/lib instead of /usr/lib64.


Vascom:
In Fedora linux 32-bit and 64-bit packages can be installed simultaneously. 32-bit libs to /usr/lib and 64-bit libs to /usr/lib64.
Please add to cmake configuration support for LIB_INSTALL_DIR parameter.

xor-gate:
Yes this is a bug, feel free to submit a PR.

PR #636, opened 2017-09-08
Added LIB_INSTALL_DIR to correct libs install on 64-bit systems

Vascom:
Solve for #633

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,8 @@ project(stlink C)
 set(PROJECT_DESCRIPTION "Open source version of the STMicroelectronics Stlink Tools")
 set(STLINK_UDEV_RULES_DIR "/etc/udev/rules.d" CACHE PATH "Udev rules directory")
 set(STLINK_MODPROBED_DIR "/etc/modprobe.d" CACHE PATH "modprobe.d directory")
-set(STLINK_LIBRARY_PATH "lib/${CMAKE_LIBRARY_PATH}" CACHE PATH "Target lib directory")
+set(LIB_INSTALL_DIR "lib" CACHE PATH "Main library directory")
+set(STLINK_LIBRARY_PATH "${LIB_INSTALL_DIR}/${CMAKE_LIBRARY_PATH}" CACHE PATH "Target lib directory")
 
 option(STLINK_GENERATE_MANPAGES "Generate manpages with pandoc" OFF)
 
--- a/doc/compiling.md
+++ b/doc/compiling.md
@@ -118,6 +118,15 @@ $ cmake -DSTLINK_UDEV_RULES_DIR="/usr/lib/udev/rules.d" \
         -DSTLINK_MODPROBED_DIR="/usr/lib/modprobe.d" ..
 ```
 
+## Build using different directory for shared libs
+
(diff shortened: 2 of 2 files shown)


xor-gate:
This looks good, but it should autoselect lib and lib64. We probably need to implement the solution which is on the cmake mailinglist here: https://cmake.org/pipermail/cmake/2013-July/055374.html

Vascom:
This not shoud autodetect lib and lib64. It set by external parameter:
`cmake .. -DLIB_INSTALL_DIR:PATH=/usr/lib64`

xor-gate:
That is not true autodetection, that is passed by the developer which builds the package. CMake should be able to autodetect it if the default libdir is lib64. I'm not against adding build parameters but the idea would be good by default.

Vascom:
Fedora has /usr/lib and /usr/lib64 simultaneously and you can't autodetect where I want place builded libs. Let maintainer do it via -DLIB_INSTALL_DIR:PATH= .

xor-gate:
Yeah the lib contains 32 bit stuff which can run with the 32bit compatibility layer. But the default detected path should be lib64 because the system booted an amd64 or arm64 kernel anyway.

Vascom:
You can add autodetect if you with but keep force redefinition of it by this parameter.

xor-gate:
Yes that would be fine, but please document the addition of the this supported build parameter in [doc/compiling.doc](https://github.com/texane/stlink/blob/master/doc/compiling.md)

Vascom:
This is enough?

xor-gate:
Great, thanks for this. The name LIB_INSTALL_DIR comes from autoconf GNU projects, but I can live with this as it is convenient for everybody.

Vascom:
Thanks.
I don't know whereit from but it used in %cmake macro for building Fedora packages:
```
  /usr/bin/cmake \
        -DCMAKE_C_FLAGS_RELEASE:STRING="-DNDEBUG" \
        -DCMAKE_CXX_FLAGS_RELEASE:STRING="-DNDEBUG" \
        -DCMAKE_Fortran_FLAGS_RELEASE:STRING="-DNDEBUG" \
        -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
        -DCMAKE_INSTALL_PREFIX:PATH=/usr \
        -DINCLUDE_INSTALL_DIR:PATH=/usr/include \
        -DLIB_INSTALL_DIR:PATH=/usr/lib64 \
        -DSYSCONF_INSTALL_DIR:PATH=/etc \
        -DSHARE_INSTALL_PREFIX:PATH=/usr/share \
%if "lib64" == "lib64" 
        -DLIB_SUFFIX=64 \
%endif 
        -DBUILD_SHARED_LIBS:BOOL=ON
```

xor-gate:
It also adds the `-DLIB_SUFFIX=64` build parameter which seems it uses the solution from the CMake mailing list. The LIB_SUFFIX is probably not supported currently in stlink. I think still we need to implement the solution from the mailing list as I see the fedora specfile to build the RPM as it uses the "lib64" conditional and no /usr/lib64 is hardcoded.

Vascom:
But -DLIB_SUFFIX is the same external parameter as -DLIB_INSTALL_DIR.

PR #639, opened 2017-09-29
Fix memory map for stm32l496xx boards.

sznaider:
STM32l496xx boards had incorrect configuration of the memory map, making memory inaccessible inside GDB.

To see the fix is right, take a look at http://www.st.com/content/ccc/resource/technical/document/reference_manual/02/35/09/0c/4f/f7/40/03/DM00083560.pdf/files/DM00083560.pdf/jcr:content/translations/en.DM00083560.pdf section **2.6.2 Boot configuration for STM32L496xx/4A6xx devices** table 7.


xor-gate:
After a quick look it looks good to me. Thanks for providing a fix for this missing xml file.

PR #641, opened 2017-10-10
add unknown chip dummy

ruefo:
The reason for my proposal of adding a dummy for the unknown chip is that I wanted to obtain the serial of the ST-link by a call to `st-info --probe`. This returns nothing (not even a warning) when an unknown chip is attached (in my case an NXP lpc17xx). With the changes here, I could easily obtain the serial, which I needed in OpenOCD, as multiple programmers are attached. (BTW: Although not supported, I am able to flash the chip, mentioned above)


xor-gate:
This is a valid fix, thank you!

PR #678, opened 2018-03-02
gcc4 error fixed

yamori813:
This is gcc4 error fixed on Mac OS X 10.6 with Xcode3.

### DIFF
--- a/src/tools/gui/stlink-gui.c
+++ b/src/tools/gui/stlink-gui.c
@@ -257,8 +257,7 @@ stlink_gui_populate_devmem_view (gpointer data)
             }
         }
         /* reads to sl->q_buf */
-        stlink_read_mem32(gui->sl, addr + off, n_read);
-        if (gui->sl->q_len < 0) {
+        if (stlink_read_mem32(gui->sl, addr + off, n_read) != 0) {
             stlink_gui_set_info_error_message (gui, "Failed to read memory");
             g_free (gui->flash_mem.memory);
             gui->flash_mem.memory = NULL;


xor-gate:
* Try to use everywhere `size_t` type for loop counter

xor-gate:
It seems the build is broken with your changes https://travis-ci.org/texane/stlink/jobs/348193686#L1113. There are two variables with the same name "variable name shadowing".

xor-gate:
We should also fix for some variable from signed to unsiged: https://travis-ci.org/texane/stlink/jobs/348203856#L1114. As `sl->q_len` should also be `size_t` instead of `int`, and must be edited in the struct. `len` should also be `size_t` https://travis-ci.org/texane/stlink/jobs/348203856#L1119. The build almost passes.

xor-gate:
Almost, i'm not sure why it is complaining https://travis-ci.org/texane/stlink/jobs/348218733#L1114. Argument is type `ssize_t` and write should return `ssize_t`.

yamori813:
I think ssize_t is better than size_t. Because of write() is return ssize_t. And some code error check use -1.

xor-gate:
You will never write negative size. But the check is indeed valid, still it complains.

yamori813:
Who to fix this ? https://travis-ci.org/texane/stlink/jobs/348218733#L1010

xor-gate:
I have not tried to build the gui, one last thing (hopefully) https://travis-ci.org/texane/stlink/jobs/349555461#L1013

yamori813:
Sorry I have no idea this. https://travis-ci.org/texane/stlink/jobs/349555461#L1013

xor-gate:
The `q_len` is a unsigned integer so a comparision with < 0 is useless. Because its boundary is 0 or higher.

yamori813:
How to fix this?

xor-gate:
https://github.com/texane/stlink/blob/master/src/tools/gui/stlink-gui.c#L260-L261

Change it to:

```
if (stlink_read_mem32(gui->sl, addr + off, n_read) != 0) {
.... cleanup
} 
```

yamori813:
Sorry I'm abandon this issue because of so many compile error. Thanks for your patient.

xor-gate:
I'm sorry it was not so easy. We will leave it as-is for now. Good luck.

PR #680, opened 2018-03-06
fix serial number size mismatch with stlink_open_usb()

ZogFluff:


xor-gate:
Looks good, i'm thinking where this has caused problems?

ZogFluff:
In my case I am using a STM32F410 Nucleo board and the serial number is long, the short array in stlib_prob_usb_devs() results in a partial serial number being captured, but the serial number compare in stlink_open_usb() is expecting a larger serial number and the serial compare in the while loop fails.

I think this would only be seen when the serial number is longer.

xor-gate:
Yes I agree. Thanks for fixing.

ZogFluff:
:-)

PR #688, opened 2018-03-19
Add desktop file for linux

Vascom:
Hi.
I am think it is good use one desktop file for all linux distributions.


xor-gate:
Please put it under src/tools/gui, it should also be installed using cmake. Not sure which is the exact path.

xor-gate:
Thanks

PR #696, opened 2018-03-28
remove exit() from lib. fix #634

### LINKED ISSUE #634: stlink library calls exit() or _exit()
When I am build package stlink for Fedora it make some checks. And I see this warning:
stlink.x86_64: W: shared-lib-calls-exit /usr/lib64/libstlink-shared.so.1.4.0 exit@GLIBC_2.2.5

shared-lib-calls-exit:
This library package calls exit() or _exit(), probably in a non-fork()
context. Doing so from a library is strongly discouraged - when a library
function calls exit(), it prevents the calling program from handling the
error, reporting it to the user, closing files properly, and cleaning up any
state that the program has. It is preferred for the library to return an
actual error code and let the calling program decide how to handle the
situation.

Can you correct it?

slyshykO:


src/logging.c
```diff
@@ -44,8 +44,6 @@ int ugly_log(int level, const char *tag, const char *format, ...) {
     case UFATAL:
```
xor-gate: We should also remove the UFATAL loglevel completely. And also the https://github.com/texane/stlink/blob/master/include/stlink/logging.h#L36-L37 which are completely unused and dangerous for a lib.

xor-gate:
Thanks

PR #697, opened 2018-03-28
Icons for stlink gui

cedricboudinet:
Hello
I create an icon for the gui as well as the desktop launcher for linux.
The icon is very basic and can be improved a lot... (I will try to make it a bit better).
But at least everything is set up.
Bests
Cedric


xor-gate:
Looks good, please move the png and svg into the ` src/tools/gui` folder as it is the only place it is used (currently).

PR #702, opened 2018-03-30
Disable static library installation by default

Vascom:
Static library not needed if shared library installed, isn't it?


bluca:
One might have applications using either at the same time, distributions always ship both

Vascom:
@bluca Are you shure? Why someone need use static and shared libs at the same time?
And about distributions you completely wrong
https://fedoraproject.org/wiki/Packaging:Guidelines#Packaging_Static_Libraries

> Packages including libraries SHOULD exclude static libs as far as possible (eg by configuring with --disable-static). Applications linking against libraries SHOULD link against shared libraries not static versions.

bluca:
I know that's the aspiration, in reality it's a bit different and users often ask for both - anyhow I don't care that much about the default, as long as there's an option for it

Vascom:
Where option for it?

xor-gate:
Having a static library installed doesn't make any sense. Static library should only be used during static builds and distributed. We should stick with dynamic library for all systems except windows (mingw)/osx builds.

Vascom:
Or may be you can add parameter to not build static lib at all?

xor-gate:
Maybe about the static library build we should do something like the nanomsg project https://github.com/nanomsg/nanomsg/blob/master/CMakeLists.txt#L99

Vascom:
It would be great.

CMakeLists.txt
```diff
@@ -163,9 +163,12 @@ endif()
 
 set_target_properties(${STLINK_LIB_STATIC} PROPERTIES OUTPUT_NAME ${PROJECT_NAME})
 
-install(TARGETS ${STLINK_LIB_STATIC}
-	ARCHIVE DESTINATION ${STLINK_LIBRARY_PATH}
-)
+set(STATIC_LIB_INSTALL ON CACHE BOOL "Install static lib")
```
xor-gate: Place the flag at the top of the cmake file so people know they can control the build. Rename the flag to `STLINK_STATIC_LIB`.

CMakeLists.txt
```diff
@@ -163,9 +163,12 @@ endif()
...
+if (STATIC_LIB_INSTALL)
+    install(TARGETS ${STLINK_LIB_STATIC}
```
xor-gate: We should not build the static library at all when `STLINK_STATIC_LIB` is not set to `ON`. We should also document this behaviour and set it default to OFF. See as an example https://github.com/nanomsg/nanomsg/blob/309af15f808546f5f04b0761d7685b90e6bcbb94/README.md#static-library

Vascom: I am try disable build static lib but fail because it used somwhere in tools. And I not understand how to disable it clearly. Thats why I am only disable install of them.

xor-gate: You are right, currently it is a bit of a mess because for windows and mac builds we use the static build (easy distribution) and for others we use the dynamic library build. We should keep it this way. Then you can ignore my comment above.

Vascom: OK

xor-gate:
Thanks

PR #706, opened 2018-04-08
rename STLINK_CHIPID_STM32_L4R9 to STLINK_CHIPID_STM32_L4RX

jnbr:
STM32L4R5ZI uses the same chip id, the reference manual (DM00310109.pdf) is the same for all L4Rxxxx devices.


xor-gate:
Thanks

Issue #700, opened 2018-03-29
Installation issue under linux


cedricboudinet:
Hello,
stlink installation under Linux (Debian 9) is broken since modifications for #695: libstlink.so (and others) installs in /usr/local/lib/lib instead of /usr/local/lib
It seems that the error comes from these instructions:
 if(IS_DIRECTORY CMAKE_LIBRARY_PATH)
 else()
    set(CMAKE_LIBRARY_PATH "lib")
 endif()
If I change it to 
 if(IS_DIRECTORY CMAKE_LIBRARY_PATH)
 else()
    set(CMAKE_LIBRARY_PATH "")
 endif()
then it works again.
I can create a PR if you want, but I cannot test the modification under other OS.

Vascom:
I can confirm.
Also hardcode path to "lib" breake x86_64 with /usr/lib64 folder.

xor-gate:
Hi guys, I agree with your both. We must find a working solution for all cases (with less cmake code).

slyshykO:
In my defence, I would say that I encounter a situation when CMAKE_LIBRARY_PATH is a list of paths. That broke all install command completely.
In my opinion, we should remove CMAKE_LIBRARY_PATH  from all install commands.

Cheng-SG:
I encountered the same issue

Vascom:
@Cheng-SG please try #701 fix

PR #701, opened 2018-03-30
fix #700

slyshykO:
Use CMAKE_INSTALL_PREFIX for install paths

Need to approve from @cedricboudinet and @Vascom

PR #707, opened 2018-04-11
Fix of build errors for Debian Stretch

joede:
Since I have to start using STM32 MCUs, I've built Debian packages out of your sources. There were some errors which I have fixed. As a result of these fixes, I have working Debian packages which I can install. I haven't had the time and hardware to test the code!

This PR should fix the issue #700.

Note: the entry in `debian/Changelog` is not what you need. ;-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ endif()
 if (WIN32)
     set(STLINK_LIBRARY_PATH ${CMAKE_LIBRARY_PATH} CACHE PATH "Target lib directory")
 else()
-    set(STLINK_LIBRARY_PATH "${LIB_INSTALL_DIR}/${CMAKE_LIBRARY_PATH}" CACHE PATH "Target lib directory")
+    set(STLINK_LIBRARY_PATH "${LIB_INSTALL_DIR}" CACHE PATH "Target lib directory")
 endif(WIN32)
 
 option(STLINK_GENERATE_MANPAGES "Generate manpages with pandoc" OFF)
--- a/debian/.gitignore
+++ b/debian/.gitignore
@@ -5,6 +5,7 @@ debhelper-build-stamp
 *.substvars
 libstlink-dev
 libstlink
+libstlink1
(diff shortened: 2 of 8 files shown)


slyshykO:
I try to fix #700 in #701.
Can you check it?

joede:
@slyshykO doesn't work for me here with Debian 9! IMO the `debian/*.install` files are wrong.

~~~~
   dh_install -O--buildsystem=cmake
dh_install: Cannot find (any matches for) "usr/lib/*/lib*.a" (tried in "." and "debian/tmp")
dh_install: libstlink-dev missing files: usr/lib/*/lib*.a
dh_install: Cannot find (any matches for) "usr/lib/*/pkgconfig/*" (tried in "." and "debian/tmp")
dh_install: libstlink-dev missing files: usr/lib/*/pkgconfig/*
dh_install: Cannot find (any matches for) "usr/lib/*/lib*.so" (tried in "." and "debian/tmp")
dh_install: libstlink-dev missing files: usr/lib/*/lib*.so
dh_install: Cannot find (any matches for) "usr/lib/*/lib*.so.*" (tried in "." and "debian/tmp")
dh_install: libstlink1 missing files: usr/lib/*/lib*.so.*
dh_install: missing files, aborting
debian/rules:14: die Regel für Ziel „binary“ scheiterte
~~~~

bluca:
See section 9.1.1.4 of the Policy: https://www.debian.org/doc/debian-policy/

joede:
@bluca thanks for the link. So the CMake system must be configured to use "`/usr/lib/`*triplet*" (i.e. `/usr/lib/x86_64-linux-gnu`) instead of the "deprecated" `/usr/lib`?

bluca:
See https://github.com/texane/stlink/pull/701#issuecomment-380747954

joede:
@xor-gate yes, it's obsolete now. Sorry, I've forgotten to close it.

xor-gate:
No problem, thanks for the effort.

Issue #684, opened 2018-03-16
Please add png icon


Vascom:
Can you add icon in PNG format for use in desktop file? Now stlink menu entry has blank icon and it is not good.
It should be 512x512 picture.

xor-gate:
Feel free to submit a PR, I only run OS X desktop.

Vascom:
And it has no any icon on OS X?

xor-gate:
Its just an CLI application, no GUI

joede:
I've added PR #708 to make the first step. I don't know CMake, so there is work needed.

PR #708, opened 2018-04-17
Add creation of icons for .desktop file

joede:
The GUI doesn't provide the icons needed to work with the `.desktop` file. See  #684! With this PR, I've added a script to create the necessary `hicolor` icons from the given `SVG` file.

I'm no CMake user, so I don't know how to add the `.desktop` and the icons to `CMakelist.txt`. After the installation is added to the `CMakelist.txt`, the `debian/stlink-gui.install` must be patched too.

**This is still needed:**
* copy `stlink.desktop` as `stlink-gui.desktp` to `/usr/share/applications/` (`${CMAKE_INSTALL_DATADIR}/applications/`?)
* copy `stlink-gui.xpm`  to `/usr/share/pixmaps/`
* copy `hicolor/*` to `/usr/share/icons/hicolor/`
* copy `stlink-gui_icon.svg` to `/usr/share/icons/hicolor/scalable/stlink-gui.svg`


xor-gate:
Maybe @Vascom can help you out with this?

Vascom:
As I found it is enough have only svg file.

joede:
@xor-gate I can try it within the next days. I expect that GNOME should work with this, but I can't say which other desktops needs images. So I *only can test the icon usage with GNOME3*!

joede:
OK, the icon is visible in GNOME3.

I've updated to PR to install the .desktop and the SVG icon with the Debian package. Therefore I have renamed the two files to `stlink-gui.*` which is IMO better. I have removed the installation instructions from the main `CMakeList.txt` to the `CMakeList.txt` of the stlink-gui one. Now all GUI related action is inside this file and no longer spread over two files.

*Again:* I don't know if the images are necessary for other desktops! Not all desktops use SVG...

Vascom:
I suggest leaving only SVG and keeping track of bug reports.

xor-gate:
@Vascom agreed

xor-gate:
Thanks @joede for your contribution.

PR #709, opened 2018-05-18
send memory map for STM32F411RE

dacodas:
I was having trouble loading .elf files for the Multitech mDot using the default memory map. 

Thanks!
Dacoda Strack


xor-gate:
Thanks!

PR #711, opened 2018-05-25
Fixed flash memory map for F72XXX chip_id

trentgill:
The F722RE uses the same flash memory map as 512kB flash f4 series. Without these changes the flash would be erased in pages of 2048bytes, rather than the full 16/64/128kb blocks on the hardware.

All the F72x and F73x devices have this same memory layout, so should be fixed by this commit.


xor-gate:
Sorry for the late reply, but looks good. Thank you!

Issue #540, opened 2017-01-09
the system reset register write instead of asserting the reset line


feathertw:
**NOTICE: The issue may be closed without notice when not enough information is provided!**

Thank you for giving feedback to the stlink project. Take some time to fill out
 check boxes with a X in the following items so developers and other people can try to
 to find out what is going on. And add/remove what is appropriate to your problem.

When submitting a feature request, try to reuse the list and add/remove what is appropriate.
 Place a `X` between the brackets `[X]` to mark the list item.

- [ ] Programmer/board type: e.g Stlink/v1, Stlink/v2, Stlink/v2-onboard
- [ ] Programmer firmware version: e.g STSW-LINK007 2.27.15
- [ ] Operating system: e.g Linux, Mac OS X, Windows (with specific version)
- [ ] Stlink tools version and/or git commit hash: e.g v1.1.0/git-c722056
- [ ] Stlink commandline tool name: e.g `st-info`, `st-flash`, `st-util`
- [ ] Target chip (and optional board): e.g STM32F402VG (STM32Fxxx Discovery)

A as-detailed description possible of the problem with debug output when available.

Output:

```
OUTPUT/ERROR of the commandline tool(s)
```

Expected/description:

I found that a strange things in st-link reset,
I use the stlink V2 (the image I post)
I can use only GND,SWDIO,SWCLK(without reset pin) to burn code into stm32
And I can use ST-Link Utility in windows "System Reset" to reset whole stm32 system(including reset peripheral)

I have no idea why I don't connect reset pin and it still can work hardware reset,
But in texane stlink, it need connect reset pin to operate hardware reset,
Is it possible to do hardware reset without connecting reset pin in texane stlink?
if it can, it's more flexible and can save the reset pin connecting

![unnamed](https://cloud.githubusercontent.com/assets/1853106/21771072/f5351946-d6c0-11e6-9891-338dd10a4041.png)
![unnamed](https://cloud.githubusercontent.com/assets/1853106/21771105/093c7682-d6c1-11e6-8160-34f9a5623838.jpg)

Thank you,
The stlink project maintainers

Nightwalker-87:
Please verify if the problem still exists in Release v1.6.0.

PR #712, opened 2018-05-30
Reset through AIRCR (#540)

timothytylee:
In addition to asserting the NRST line, also request system reset through the
Application Interrupt and Reset Control Register (AIRCR).

This patch has been tested against an STM32L071 using the SWD without connecting the NRST pin.


xor-gate:
Sorry for the late reply. Good addition, thanks!

Issue #713, opened 2018-06-10
GUI failed to parse .hex file


FoxyHawk:
- [x] Programmer/board type: Stlink/v2-onboard
- [x] Programmer firmware version: V2.J29.M18
- [x] Operating system: Arch Linux (Manjaro)
- [x] Stlink tools version and/or git commit hash: v1.5.0-2 (lastest Manjaro repo)
- [x] Stlink commandline tool name: `stlink-gui`
- [x] Target chip (and optional board): STM32F767ZI

Reading of .hex files in stlink-gui differ from Windows Official ST-link Utility.
It seems to failed to interpret the first line of the file as an address offset.
```
:020000040800F2
:1000000000000820290300082F030008330300081C
:100010007D03000881030008C903000800000000F8
:1000200000000000000000000000000011040008B3
:10003000150400080000000019040008150D000850
:100040002D0400082D0400082D0400082D040008CC
......
```
Based on https://en.wikipedia.org/wiki/Intel_HEX, the first line must be read as start at 0x08000000.
On Arch Linux, the file become unusable (but the programming  and writing are fine btw) and the result on the card is unpredictable.
![rep_arch_read](https://user-images.githubusercontent.com/37553238/41199816-0eb220b6-6c99-11e8-8a37-be75d865bb76.png)
![rep_windows_card_read](https://user-images.githubusercontent.com/37553238/41199817-10ba3f74-6c99-11e8-9d9f-c8a4b7aa6e70.png)

On Windows the file is correctly parsed and the result is what I expect.
![rep_windows_read](https://user-images.githubusercontent.com/37553238/41199819-16263d0a-6c99-11e8-968d-29a1e0d15dcc.png)
![rep_windows_card_write](https://user-images.githubusercontent.com/37553238/41199818-14466154-6c99-11e8-94b5-9677f728042b.png)

The .hex file in .txt (only a basic blink LED for my card):
[test2.txt](https://github.com/texane/stlink/files/2087654/test2.txt)

xor-gate:
@dev26th I would like you to have a look at this issue because you have integrated intel hex reading some time ago in https://github.com/texane/stlink/pull/459.

dev26th:
Sure, it's just a small piece of code

xor-gate:
Thanks for reporting and resolving!

PR #718, opened 2018-06-24
implement IHEX for the GUI

dev26th:
#713 The GUI interpreted all files as binaries. Try to load '*.hex' as Intel-HEX


xor-gate:
Looks good @dev26th ! @FoxyHawk could you verify it works with this patch for you?

FoxyHawk:
The reader part of the GUI work fine now.

update_file_reader

But the programme use all the file, not only the data.
update_post_prog

FoxyHawk:
Tested with my file, everything is fine ^^

xor-gate:
Thanks all!

PR #721, opened 2018-07-06
Proper flash page size calculation for F412

rojer:


rojer:
this speeds up flashing tremendously by avoiding unnecessary erase operations

xor-gate:
Sorry for the late reply, thanks !

PR #727, opened 2018-07-26
For Issue 726, Semihost SYS_READ returns incorrect value on EOF

donmr:
Use local variable for read_result instead of *ret, and fix calculation of *ret for EOF case.

### DIFF
--- a/src/gdbserver/semihosting.c
+++ b/src/gdbserver/semihosting.c
@@ -307,6 +307,7 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         int      fd;
         uint32_t buffer_len;
         void    *buffer;
+	int	 read_result;
 
         if (mem_read(sl, r1, args, sizeof (args)) != 0 ) {
             DLOG("Semihosting SYS_READ error: "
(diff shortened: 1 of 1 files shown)


xor-gate:
Great, looks like a good fix! Thanks for your PR.

src/gdbserver/semihosting.c
```diff
@@ -337,7 +338,7 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         DLOG("Semihosting: read(%d, target_addr:0x%08x, %zu)\n", fd,
              buffer_address, buffer_len);
 
-        *ret = (uint32_t)read(fd, buffer, buffer_len);
+        read_result = (uint32_t)read(fd, buffer, buffer_len);
         saved_errno = errno;
 
         if (*ret == (uint32_t)-1) {
```
Fabien-Chouteau: I think this has to be changed to `if (read_result == -1) {`

src/gdbserver/semihosting.c
```diff
@@ -337,7 +338,7 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         DLOG("Semihosting: read(%d, target_addr:0x%08x, %zu)\n", fd,
              buffer_address, buffer_len);
 
-        *ret = (uint32_t)read(fd, buffer, buffer_len);
+        read_result = (uint32_t)read(fd, buffer, buffer_len);
```
Fabien-Chouteau: uint32_t cast is not needed anymore

src/gdbserver/semihosting.c
```diff
@@ -307,6 +307,7 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         int      fd;
         uint32_t buffer_len;
         void    *buffer;
+	int	 read_result;
```
Fabien-Chouteau: Type should probably be `ssize_t`

donmr:
I agree with Fabien-Chouteau's review items.  Thanks for catching my sloppy errors.

I'll submit another PR shortly (after I check it myself one more time).

PR #728, opened 2018-07-26
Further fix for read at EOF, don't write if length is zero.

donmr:
I found a further problem when reading at the end of a file.

If there is some data in the file, but not enough, then after the first read call returns some data, fread will ask for the rest with an updated pointer.  If the first read returned an odd modulo 4 number of bytes, then this second read call will happen with a pointer that is not 32 bit aligned.

In do_semihosting, the second read gets 0 bytes but still calls mem_write.  The code in mem_write which forces an aligned address then writes more data than it should altering the values returned from the first read call.

This change simply has mem_write return without doing anything if the length is zero.

### DIFF
--- a/src/gdbserver/semihosting.c
+++ b/src/gdbserver/semihosting.c
@@ -307,6 +307,7 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         int      fd;
         uint32_t buffer_len;
         void    *buffer;
+	int	 read_result;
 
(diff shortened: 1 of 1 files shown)


donmr:
Opps, the git push didn't get in (I tried to use ssh).  It's there now.  Is this PR sufficent or do I need to create another??

donmr:
I'm going to close this PR, and revert the change so I can get the fixes based on the previous review in by themselves.  Then I'll re-submit this one.

PR #729, opened 2018-07-27
Fixes to SYS_READ changes in PR #727 per review.

donmr:
I agree with Fabien-Chouteau's review items. Thanks for catching my sloppy errors.
Here is a PR with those things fixed.

### DIFF
--- a/src/gdbserver/semihosting.c
+++ b/src/gdbserver/semihosting.c
@@ -307,6 +307,7 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         int      fd;
         uint32_t buffer_len;
         void    *buffer;
+	int	 read_result;
 
         if (mem_read(sl, r1, args, sizeof (args)) != 0 ) {
             DLOG("Semihosting SYS_READ error: "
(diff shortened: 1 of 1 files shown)


xor-gate:
Thanks @Fabien-Chouteau and @donmr !

src/gdbserver/semihosting.c
```diff
@@ -338,20 +338,20 @@ int do_semihosting (stlink_t *sl, uint32_t r0, uint32_t r1, uint32_t *ret) {
         DLOG("Semihosting: read(%d, target_addr:0x%08x, %zu)\n", fd,
              buffer_address, buffer_len);
 
-        read_result = (uint32_t)read(fd, buffer, buffer_len);
+        read_result = read(fd, buffer, buffer_len);
         saved_errno = errno;
 
-        if (*ret == (uint32_t)-1) {
+        if (read_result == (uint32_t)-1) {
```
slyshykO: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

donmr: You are right, read_result is a ssize_t now (which is a signed integer).  I'll fix this too since it's my mistake.

PR #730, opened 2018-07-27
Mem write odd

donmr:
Here is the first fix to mem_write() again.  This is simply to have it return when len == 0 and that solves my issues with reading at the end of a file when the remaining data is not an even multiple of 4 bytes.

I have also recognized that this function will overwrite target data that is shouldn't if it is ever called with a non-zero length and either the address or length is not a multiple of 4.  I don't know if that situation will ever occur, it does not happen in my limited uses (so far :-) ).  I have added some comments with possible ways to address that issue.


xor-gate:
This is a good fix, checking before calculation is always a good idea to not go beyond boundaries. Thanks!

PR #731, opened 2018-07-29
Fix to cast in changes for SYS_READ.

donmr:
As slyshykO notes in #729, read_result is a signed type so the -1 constant should not be cast to unsigned.
I belive compilers will be happy with this without any cast.


xor-gate:
Thanks!

PR #733, opened 2018-08-01
FreeBSD define LIBUSB_API_VERSION instead of LIBUSBX_API_VERSION.

lhondareyte:
LIBUSB_API_VERSION is not defined on FreeBSD, so it fail to compile.


xor-gate:
Thanks @lhondareyte for the fix.

Issue #750, opened 2018-12-11
Flash verification error at 128K mark on STM32L432C6 (NUCLEO-L432KC)


dhylands:
**NOTICE: The issue may be closed without notice when not enough information is provided!**

Thank you for giving feedback to the stlink project. Take some time to fill out
 check boxes with a X in the following items so developers and other people can try to
 to find out what is going on. And add/remove what is appropriate to your problem.

When submitting a feature request, try to reuse the list and add/remove what is appropriate.
 Place a `X` between the brackets `[X]` to mark the list item.

- [ST-LINK/V2-1 onboard ] Programmer/board type: e.g Stlink/v1, Stlink/v2, Stlink/v2-onboard
- [STSW-LINK007 2.32.22] Programmer firmware version: e.g STSW-LINK007 2.27.15
- [Linux - Ubuntu 18.04 LTS ] Operating system: e.g Linux, Mac OS X, Windows (with specific version)
- [ a201d3e5bd4a52a6e2a7f89947688f580c2b3554 (latest master)] Stlink tools version and/or git commit hash: e.g v1.1.0/git-c722056
- [st-flash write firmware2.bin 0x08000000] Stlink commandline tool name: e.g `st-info`, `st-flash`, `st-util`
- [ STM32L432C6] Target chip (and optional board): e.g STM32F402VG (STM32Fxxx Discovery)

A as-detailed description possible of the problem with debug output when available.

The firmware files I was using is available here:
https://www.dropbox.com/sh/nxgho1y9yj4bgfj/AACn_ZOkEN2DRAsuXyCRe9G-a?dl=0

They are almost identical versions of Micropython. Of note is that they're both larger than 128K.

Output:
Flashing onto the NUCLEO-L432KC with an erased flash (i.e. first time or after `st-flash erase`) works as expected:
```
$ st-flash write firmware1.bin 0x08000000
st-flash 1.4.0-54-ga201d3e
2018-12-11T12:52:28 INFO common.c: Loading device parameters....
2018-12-11T12:52:28 INFO common.c: Device connected is: L43x/L44x device, id 0x10016435
2018-12-11T12:52:28 INFO common.c: SRAM size: 0xc000 bytes (48 KiB), Flash: 0x40000 bytes (256 KiB) in pages of 2048 bytes
2018-12-11T12:52:28 INFO common.c: Attempting to write 219420 (0x3591c) bytes to stm32 address: 134217728 (0x8000000)
[...]

PR #751, opened 2018-12-13
Only do bank calculation on STM32L4 devices with dual banked flash

dhylands:
RM0394 covers the STM32L41xx, 42xx, 43xx, 44xx, 45xx, and 46xx. These
devices are all employ single banked flash and have chip id's
of 0x464 for the 41xx/42xx, 0x435 for 43xx/44xx, and 0x462 for 45xx/46xx
It's also worth noting that bit 21 of the FLASH_OPTR register is marked
as resevred for these chips, and isn't an indicator of dual banked
flash.

RM0392 covers the STM32L4x1, cpu_id 0x415 and can be dual banked.

RM0351 covers the STM32L4x5/4x6, cpu_ids 0x415 & 0x461 and can be dual
banked

RM0432 covers the STM32L4Rx/4Sx, cpu_id 0x470 and can be dual banked.

This PR modifies the calculate_L4_page functio to only factor bank
calculations for the devices above which can support dual banked flash.

I also checked the STM32L4 HAL code and it uses the following preprocessor #if statment
```
#if defined (STM32L471xx) || defined (STM32L475xx) || defined (STM32L476xx) || defined (STM32L485xx) || defined (STM32L486xx) || \
    defined (STM32L496xx) || defined (STM32L4A6xx)
```
to identify dual-banked devices,

### DIFF
--- a/include/stlink/chipid.h
+++ b/include/stlink/chipid.h
@@ -45,7 +45,7 @@ enum stlink_stm32_chipids {
 	STLINK_CHIPID_STM32_L43X             = 0x435,
 	STLINK_CHIPID_STM32_L496X            = 0x461,
 	STLINK_CHIPID_STM32_L46X             = 0x462,
-	STLINK_CHIPID_STM32_L41X			 = 0x464,
+	STLINK_CHIPID_STM32_L41X             = 0x464,
 	/*
 	* 0x436 is actually assigned to some L1 chips that are called "Medium-Plus"
 	* and some that are called "High".  0x427 is assigned to the other "Medium-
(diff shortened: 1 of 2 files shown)


dhylands:
This fixes #750

include/stlink/chipid.h
```diff
@@ -40,10 +40,12 @@ enum stlink_stm32_chipids {
 	* 0x435 covers STM32L43xxx and STM32L44xxx devices
 	* 0x461 covers STM32L496xx and STM32L4A6xx devices
 	* 0x462 covers STM32L45xxx and STM32L46xxx devices
+	* 0x464 covers STM32L41xxx and STM32L42xxx devices
 	*/
 	STLINK_CHIPID_STM32_L43X             = 0x435,
 	STLINK_CHIPID_STM32_L496X            = 0x461,
 	STLINK_CHIPID_STM32_L46X             = 0x462,
+	STLINK_CHIPID_STM32_L41X			 = 0x464,
```
xor-gate: Keep same alignment of the number

dhylands:
Whoops - thanks for noticing that. My editor inserted tabs instead of spaces. Fixed.

xor-gate:
Thanks!

PR #760, opened 2019-01-13
Add preliminary support for some STM32G0 chips. (closes #759)

### LINKED ISSUE #759: Add support for STM32G0 chips
It'd be nice to support the fairly new STM32G0 line of chips.

- [x] Programmer/board type: Stlink/v2
- [X] Programmer firmware version: V2.J29.S7
- [X] Operating system: Ubuntu 18.04 (all platforms?)
- [X] Stlink tools version and/or git commit hash: git-7651d2116fd74c7803ea00ab1da7cf3d00faf44c
- [X] Stlink commandline tool name: st-util, st-flash
- [X] Target chip (and optional board): STM32G071GB

Right now they aren't supported:

```
2019-01-12T16:04:23 INFO common.c: Loading device parameters....
2019-01-12T16:04:23 WARN common.c: unknown chip id! 0x20006460
```

And it's possible that I'm reading the manuals wrong, but it also looks like the core flash register offsets (CR, SR, KEYR) are slightly different from most other STM32 lines.

I wrote a small patch that seems to work for erasing and flashing an STM32G071GB chip, but I have only tested it with a small LED-blinking example which only occupies a single page of Flash:

https://github.com/WRansohoff/stlink/commit/c1ad2cf6f7bed2c22f18dfdd0087cd0d5e8b9997

And besides the lack of testing, I don't know how well that conforms to the rest of the project's code style and formatting, so it's probably not worth making a pull request yet.

WRansohoff:
This is just a starting point and I've only tested it with a small 'hello world' example on one STM32G071GB board, but I ordered one of ST's Nucleo boards now that they're back in stock, so hopefully I can try it with some STCube examples before long.

TODO:
* Use the MER 'mass erase' option for erasing flash instead of going page-by-page.
* The STM32G0 reference manual mentions a 'fast programming' mode which I haven't looked at.
* It would probably be easier to read if I updated and used the methods like 'wait_flash_busy', 'unlock_flash', etc.
* Some of the flash register definitions are not used.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1640,8 +1640,8 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr)
         /* Set the page to erase. */
         uint32_t flash_page = ((flashaddr - STM32_FLASH_BASE) / sl->flash_pgsz);
         stlink_read_debug32(sl, STM32G0_FLASH_CR, &val);
-        // sec 3.7.5 - PNB[5:0] is offset by 3.
-        val |= (flash_page & 0x3F) << 3;
+        // sec 3.7.5 - PNB[5:0] is offset by 3. PER is 0x2.
+        val = ((flash_page & 0x3F) << 3) | (2);
         stlink_write_debug32(sl, STM32G0_FLASH_CR, val);
         /* Set the 'start' bit. */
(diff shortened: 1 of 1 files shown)


WRansohoff:
I found a bug in the flash-erasing section which caused some pages to not be erased; sorry. I think it works now, though.

PR #773, opened 2019-03-01
Added call to clear PG bit after Writing to Flash

ronnymajani:
The st-flash write utility wasn't clearing the PG bit after it finished writing to Flash.
Luckily you've already implemented the function for clearing the PG bit, so I just added the call right before locking Flash CR again.
The reason this is important is that if I just flashed a firmware, then any subsequent operation will fail, until I clear the PG bit, so I'll have to add code just for that one case. Or a simple power cycle the STM32 clears it.
Anyway it's just a convenience for the flasher utility to clean up after itself :)

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -320,7 +320,7 @@ static void set_flash_cr_pg(stlink_t *sl) {
     stlink_write_debug32(sl, cr_reg, x);
 }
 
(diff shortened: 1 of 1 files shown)


xor-gate:
Thanks @ronnymajani !

PR #775, opened 2019-03-06
doc/compiling.md: fix package name "devscripts"

aoand:
the command 'debuild' is (and always was) part of the devscripts package


xor-gate:
Thanks @aoand !

PR #776, opened 2019-03-09
Added howto for sending NRST signal through GDB

ceremcem:
See https://github.com/texane/stlink/issues/774


xor-gate:
Thanks @ceremcem !

ceremcem:
You're welcome. But I'm sorry for the silly backtick mistake.

PR #778, opened 2019-03-15
Added support to write option bytes for the STM32G0

xeniter:
tested with STM32G070

### DIFF
--- a/include/stlink.h
+++ b/include/stlink.h
@@ -58,6 +58,7 @@ extern "C" {
     // Constant STM32 memory map figures
 #define STM32_FLASH_BASE 0x08000000
 #define STM32_SRAM_BASE 0x20000000
+#define STM32_G0_OPTION_BYTES_BASE ((uint32_t)0x1FFF7800)
 
 // Baud rate divisors for SWDCLK
 #define STLINK_SWDCLK_4MHZ_DIVISOR		0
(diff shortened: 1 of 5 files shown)


include/stlink.h
```diff
@@ -58,6 +58,7 @@ extern "C" {
     // Constant STM32 memory map figures
 #define STM32_FLASH_BASE 0x08000000
 #define STM32_SRAM_BASE 0x20000000
+#define STM32_G0_OPTION_BYTES_BASE ((uint32_t)0x1FFF7800)
```
xor-gate: we should create an include/stlink/stm32.h header file (and new folder). and put it in there also with the ones on line 59 + 60. as the stlink.h is "the entry point" for the library it is good to not pollute it with lowlevel register definitions.

src/common.c
```diff
@@ -2407,3 +2421,116 @@ int stlink_fwrite_flash(stlink_t *sl, const char* path, stm32_addr_t addr) {
...
+
+    if(len != 4) {
+        ELOG("Wrong length for writting option bytes, must be 4 is %d\n", len);
+        return -1;
+    }
+
+    // Make sure we've loaded the context with the chip details
+    stlink_core_id(sl);
+
+    if((sl->chip_id == STLINK_CHIPID_STM32_G0X1) && (addr == STM32_G0_OPTION_BYTES_BASE)) {
```
xor-gate: if you invert this conditional you could do an early return with -1 and the log error other series are not supported yet. this makes it more readable because the code is fairly long.

xeniter: good point, done

include/stlink.h
xeniter: created

xor-gate:
Thanks @xeniter !

PR #796, opened 2019-04-26
Merge most recent version of STLink into the fork.

WRansohoff:


Nightwalker-87:
Commits are included in PR #822 & #825 instead.

PR #797, opened 2019-04-26
Fix apparent STM32G0 flashing issue.

WRansohoff:
I'm not entirely sure why, but this seems to fix an issue with flashing STM32G0 chips.


xor-gate:
Thanks!

PR #802, opened 2019-05-11
build for windows under Debian/Ubuntu

slyshykO:


xor-gate:
great work, thanks @slyshykO !

PR #803, opened 2019-05-12
fixed few potential memory/resource leaks

slyshykO:
also few cppcheck warnings


xor-gate:
Thanks @slyshykO we should also have some sort of `scan-build` in our CI system where static analysis with Clang checks for errors. E.g https://stackoverflow.com/questions/31177633/how-to-find-memory-leaks-with-clang

PR #809, opened 2019-06-06
Allow for 64 Bytes serials (with new Nucleo-L476RG)

DL6AKU:
On Nucleo boards the serial gets truncated. This allows for longer serials.


xor-gate:
Thanks, for which nucleo board this was an issue? This makes the changelog a little bit more clear!

DL6AKU:
> Thanks, for which nucleo board this was an issue? This makes the changelog a little bit more clear!

Nucleo-L476RG

xor-gate:
Thanks!

PR #812, opened 2019-06-26
fixes texane/stlink#780

ceremcem:


src/usb.c
```diff
@@ -52,7 +52,7 @@ ssize_t send_recv(struct stlink_libusb* handle, int terminate,
             3000);
     if (t) {
         printf("[!] send_recv send request failed: %s\n", libusb_error_name(t));
-        return -1;
+        exit(1);
```
xor-gate: this solution is not acceptable as stlink can also be used as a library, and a library may never use exit. probably the higher layer should catch the error and exit.

* https://github.com/texane/stlink/blob/master/src/gdbserver/gdb-server.c#L264-L267
* https://github.com/texane/stlink/blob/master/src/gdbserver/gdb-server.c#L1063 
** This function complexity is insane, a lot of `stlink_` calls are not checked for error return values and the main loop should exit not the library itself

Let me know what you think, its a dirty job to fix but its worth it.

ceremcem: Does every `-1` response mean `exit` in the higher layer?

xor-gate: That is a very good question! IMHO when the libusb backend call fails, indeed there is some sort of failure with the I/O which most of the time you are unable to recover. E.g a stlink instance destroy is nessary, rescan of usb devices and attach again or try with exponential backoff. This is the behaviour `openocd` has when you "hotplug" an stlink programmer which is very nice.

xor-gate:
As my comment before, I think in the upper layers we should catch the error -1 and exit (or retry with exponential backoff using rescan). I think a different implementation is necessary for that. Do you feel to fix this or leave it as is ?

src/usb.c
olikraus: Only my opinion: If the backend fails, then it should return an error code (which is "-1").
What I have seen so far, the higher levels will react accordingly to the -1 return value.
So: I would keep this as it is.

Nightwalker-87: @xor-gate mentioned an important point here. Therefore I don't see a reason for changing this either.

ceremcem:
Sorry for being unresponsive for that long time. I had had many distractions and I still have. Current workaround works for me but it is apparently a dirty hack and needs to be improved as you suggested. I'm willing to implement the proper solution ASAP.

Nightwalker-87:
@ceremcem: Thx for your contribution. Could you target the develop branch for that PR? That would be great.

ceremcem:
> Could you target the develop branch for that PR?

Sure thing. 

> Can we be sure in the meanwhile that this does not break anything currently working?

I'll take extreme caution on this.

Nightwalker-87:
@ceremcem: Any progress on this?

src/usb.c
Nightwalker-87: Conversation continues on 21.02.2020.

Nightwalker-87:
@ceremcem: This PR will be closed in 3 days, should there be no further feedback.

ceremcem:
Hi! I'm on this one at last. 

Any fix would need a test case in order to prevent any breakage in the higher layers. What would you recommend? Should I write `my-st-util` or is there any application I can use as a test case?

Nightwalker-87:
Can you just use `st-util` from the `develop` branch?

Nightwalker-87:
Closed by PR #913.

PR #822, opened 2019-07-27
Add support for STM32G4 chips

WRansohoff:
This uses the same logic as the G0/WB chips. And the G0 / G4 series seem to use the same general Flash register layout, so I merged some of those registers and bit fields.

I also tested this with a program which spans more than 3 pages on an STM32G431 and an STM32G071 'Nucleo-64' board, so hopefully this change will work the first time unlike the G0/WB support.

Knock on wood...


Kezii:
Can someone merge this?

Nightwalker-87:
Redirected to develop branch. Closes #822.

Nightwalker-87:
@WRansohoff: Thx for your contribution!

PR #824, opened 2019-08-10
Add aliased SRAM2 region in the L496 memory map

MFaehling:
The SRAM2 on the L496 devices is aliased at 0x2004000. With this change, it's possible to read the SRAM2 contents at the aliased addresses.


MFaehling:
PS: I can confirm that the NUCLEO-L496ZG board works, so it can also be added to the tested boards list

Nightwalker-87:
Merged with commit 364f24132e2d39f6148d71f2ad41b6344e9e1f56.

Nightwalker-87:
@MFaehling: Thx for your contribution!

Issue #833, opened 2019-09-15
CKS32: Unable to flash STM32F103C8T6 with stlink 1.5.1


kynguyen98:
I'm on Arch linux and i install stlink through it repository and when I install all the require dependencies and run `stlink-flash hello.bin 0x08000000` it gave me this error

> st-flash 1.5.1
2019-09-15T00:41:53 INFO usb.c: -- exit_dfu_mode
2019-09-15T00:41:53 INFO common.c: Loading device parameters....
2019-09-15T00:41:53 INFO common.c: Device connected is: F1 Medium-density device, id 0x20036410
2019-09-15T00:41:53 INFO common.c: SRAM size: 0x5000 bytes (20 KiB), Flash: 0x20000 bytes (128 KiB) in pages of 1024 bytes
2019-09-15T00:41:53 INFO common.c: Attempting to write 1228 (0x4cc) bytes to stm32 address: 134217728 (0x8000000)
Flash page at addr: 0x08000400 erased
2019-09-15T00:41:53 INFO common.c: Finished erasing 2 pages of 1024 (0x400) bytes
2019-09-15T00:41:53 INFO common.c: Starting Flash write for VL/F0/F3/F1_XL core id
2019-09-15T00:41:53 ERROR flash_loader.c: unknown coreid, not sure what flash loader to use, aborting! coreid: 2ba01477, chipid: 410
2019-09-15T00:41:53 WARN flash_loader.c: Failed to write flash loader to sram!
2019-09-15T00:41:53 ERROR common.c: stlink_flash_loader_init() == -1
stlink_fwrite_flash() == -1
> 

I'm been searching online how to fix this but no luck and also the preloaded led blink is gone so I don't know if I broke the chip or not. If anybody find a solution for this I'll be appreciated thanks

nbenm:
Me too.
Running
`st-flash --reset --format binary write generic_boot20_pc13_bootloader_par_defaut_led_en_pc13.bin 0x8000000`

gives:

```
st-flash 1.5.1
2019-09-16T15:02:03 INFO usb.c: -- exit_dfu_mode
2019-09-16T15:02:03 INFO common.c: Loading device parameters....
2019-09-16T15:02:03 INFO common.c: Device connected is: F1 Medium-density device, id 0x20036410
2019-09-16T15:02:03 INFO common.c: SRAM size: 0x5000 bytes (20 KiB), Flash: 0x10000 bytes (64 KiB) in pages of 1024 bytes
[...]

PR #805, opened 2019-05-19
Add support for CS32F103C8T6

Humanoid0:
CS32F103C8T6 is clone of STM32F103C8T6 but with incorrect core_id number. ( #757 )

All original STM32F1xx has core id STM32VL_CORE_ID, but CS32F103 has different core id 0x2BA01477 (as original STM32F3xx, STM32F4xx and STM32L1xx).

This commit should not affect other devices because all original STM32F1xx (include STLINK_CHIPID_STM32_F1_MEDIUM) has core id STM32VL_CORE_ID.

Issue #761, opened 2019-01-14
Unable to flash with latest revision of master - STM32F2/F4/L4 (Regression)


michaelsobczak-qeexo:
**NOTICE: The issue may be closed without notice when not enough information is provided!**

Thank you for giving feedback to the stlink project. Take some time to fill out
 check boxes with a X in the following items so developers and other people can try to
 to find out what is going on. And add/remove what is appropriate to your problem.

When submitting a feature request, try to reuse the list and add/remove what is appropriate.
 Place a `X` between the brackets `[X]` to mark the list item.

- [X] Programmer/board type: stlink/v2
- [x] Programmer firmware version: Not sure
- [X] Operating system: Mac OS X
- [X] Stlink tools version and/or git commit hash: 6a9d390a729f381ecec45f212354bfe98e27790f
- [X] Stlink commandline tool name: `st-flash`
- [X] Target chip (and optional board): STM32L476JG (STSW-STLKT01V1)

A as-detailed description possible of the problem with debug output when available.

Output:

```
Release$ st-flash write ~/data/input/st-ces/lib/qxostlib-datalog-20190104112820-ed73f5.bin 0x8000000 st-flash 1.4.0-58-g6a9d390
2019-01-14T16:35:23 INFO common.c: Loading device parameters.... 
2019-01-14T16:35:23 INFO common.c: Device connected is: L4 device, id 0x10076415
2019-01-14T16:35:23 INFO common.c: SRAM size: 0x18000 bytes (96 KiB), Flash: 0x100000 bytes (1024 KiB) in pages of 2048 bytes
2019-01-14T16:35:23 INFO common.c: Attempting to write 109160 (0x1aa68) bytes to stm32 address: 134217728 (0x8000000) Flash page at addr: 0x0801a800 erasedEraseFlash - Page:0x35 Size:0x800
2019-01-14T16:35:24 INFO common.c: Finished erasing 54 pages of 2048 (0x800) bytes 
2019-01-14T16:35:24 INFO common.c: Starting Flash write for F2/F4/L4
2019-01-14T16:35:24 INFO flash_loader.c: Successfully loaded flash loader in sram size: 32768 2019-01-14T16:35:27 ERROR flash_loader.c: flash loader run error
2019-01-14T16:35:27 ERROR common.c: stlink_flash_loader_run(0x8000000) failed! == -1 stlink_fwrite_flash() == -1
```

Expected/description:
[...]

PR #757, opened 2019-01-03
Add support for CS32 micro-controller

VictorLamoine:
Fixes #756 

I'm pretty sure the name `CS32_CORE_ID` is not appropriate so tell me what I should use.
I have tested and it works, I can upload the STM32duino bootloader on the CS32F103C8T6 board just fine.

PR #834, opened 2019-09-16
added CSK32F103C8T6 a stm32f103c8t6 clone with coreid 0x2ba01477

nbenm:
Here's the pull request we talked about a few minutes ago.
This adds support to CSK32F103C8T6 a stm32f103c8t6 clone with an unsupported coreid

### DIFF
--- a/src/flash_loader.c
+++ b/src/flash_loader.c
@@ -262,6 +262,7 @@ int stlink_flash_loader_write_to_sram(stlink_t *sl, stm32_addr_t* addr, size_t* 
         loader_code = loader_code_stm32l;
         loader_size = sizeof(loader_code_stm32l);
     } else if (sl->core_id == STM32VL_CORE_ID
+            || sl->chip_id == CS32VL_CORE_ID
             || sl->chip_id == STLINK_CHIPID_STM32_F1_MEDIUM
             || sl->chip_id == STLINK_CHIPID_STM32_F3
             || sl->chip_id == STLINK_CHIPID_STM32_F3_SMALL
--- a/include/stm32.h
+++ b/include/stm32.h
@@ -9,6 +9,7 @@
 
 // cortex core ids
 #define STM32VL_CORE_ID 0x1ba01477
+#define CS32VL_CORE_ID 0x2ba01477
 #define STM32F7_CORE_ID 0x5ba02477
 
 // Constant STM32 memory map figures


nbenm:
Is there anything else you need before merging?

src/flash_loader.c
```diff
@@ -262,6 +262,7 @@ int stlink_flash_loader_write_to_sram(stlink_t *sl, stm32_addr_t* addr, size_t*
         loader_code = loader_code_stm32l;
         loader_size = sizeof(loader_code_stm32l);
     } else if (sl->core_id == STM32VL_CORE_ID
+            || sl->chip_id == CS32VL_CORE_ID
```
Humanoid0: You are comparing Chip ID with Core ID.
Even if you fix it and will compares sl->core_id then this code will break work with many STM32F4 chips with Core ID 0x2ba01477
See #757 and #761 

What Chip ID printed if you run "st-info --probe"?

nbenm: Oh my bad!
I feel ashamed.
chipid is 0x0410 and coreid is 0x2ba01477

Humanoid0: 0x410 (STLINK_CHIPID_STM32_F1_MEDIUM) is already in use and should work. Please check latest version. CS32F103 support ( #805 ) was recently added (3 month ago).

nbenm: The problem is **not** with the chipid. It's the coreid that is unknown.
For this chipid it is waiting for 0x1ba01477 not 0x2ba01477
That's why it says: `ERROR flash_loader.c: unknown coreid, not sure what flash loader to use, aborting! coreid: 2ba01477, chipid: 410` see issue #833
It's not a real STM32F103C8T6, it's a CKS32F103C8T6

Humanoid0: Message about unknown coreid is not entirely correct. Function stlink_flash_loader_write_to_sram() uses chipid to select flash loader type. And only for all STM32F1 and STM32F7 chips function uses coreid. If function can not detect chip, it prints a message. I think message must be "unknown chipid and coreid" or "can not detect chip type".

CKS32F103C8T6 and STM32F103C8T6 have same chipid, but CKS32 uses incorrect coreid. In #805 added check for STLINK_CHIPID_STM32_F1_MEDIUM and now CS/CKS32F103C8T6 should works. Did you try latest version?

nbenm: I don't understand.
I've just downloaded latest version, compiled it, and it works fine.
I was sure I was using last version. Maybe a PATH issue with multiple versions.
This PR becomes useless.
Sorry for the wasted time.

Nightwalker-87:
Finally closed by #805.

src/flash_loader.c
Nightwalker-87: Issue featured in Release v1.6.1 which resolves this discussion.

Nightwalker-87:
Already fixed in #805.

PR #836, opened 2019-10-07
Added Read/Write command for Option-Bytes on STM32F2 series 

Orie22:
Option Bytes were not accessible for STM32F2 devices, so I added read/write command for Option Bytes on STM32F2 devices.

### read operation  
 st-flash --reset --area=option read > option_byte 
### write operation  
 st-flash --reset --area=option write 0x0FFFAAE0

### DIFF
--- a/src/tools/flash.c
+++ b/src/tools/flash.c
@@ -38,6 +38,7 @@ static void usage(void)
     puts("                       fsize: Use decimal, octal or hex by prefix 0xXXX for hex, optionally followed by k=KB, or m=MB (eg. --flash=128k)");
     puts("                       Format may be 'binary' (default) or 'ihex', although <addr> must be specified for binary format only.");
     puts("                       ./st-flash [--version]");
+    puts("example write option byte: ./st-flash --debug --reset --area=option write 0xXXXXXXXX");
 }
 
 int main(int ac, char** av)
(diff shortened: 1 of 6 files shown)


tuankiet65:
This breaks the reading functionality I think.

src/tools/flash.c
```diff
@@ -208,13 +209,25 @@ int main(int ac, char** av)
...
+                err = stlink_read_option_bytes_f2(sl,&option_byte);
+                printf("%x\n",option_byte);
+            }else{
+                printf("This format is available for STM32F2 Only\n");
+            }
+        }
+        else if ((o.addr >= sl->flash_base) && (o.size == 0) &&
+                (o.addr < sl->flash_base + sl->flash_size)){
             o.size = sl->flash_size;
+            err = stlink_fread(sl, o.filename, o.format == FLASH_FORMAT_IHEX, o.addr, o.size);
```
tuankiet65: The call to stlink_fread to read out the content is placed inside the if intended to automatically fill in the correct flash/SRAM size.

src/tools/flash.c
```diff
@@ -208,13 +209,25 @@ int main(int ac, char** av)
...
+        }
         else if ((o.addr >= sl->sram_base) && (o.size == 0) &&
-                (o.addr < sl->sram_base + sl->sram_size))
+                (o.addr < sl->sram_base + sl->sram_size)){
             o.size = sl->sram_size;
-        err = stlink_fread(sl, o.filename, o.format == FLASH_FORMAT_IHEX, o.addr, o.size);
+            err = stlink_fread(sl, o.filename, o.format == FLASH_FORMAT_IHEX, o.addr, o.size);
```
tuankiet65: Ditto as above.

src/tools/flash.c
Orie22: Thanks for comment. I will fixed today and submit pull request.
https://github.com/Orie22/stlink/commit/4a32157143547755cab260e58d1b00f8ddaa7780#diff-63af8289eae43567b379c0c480159f85R226

PR #847, opened 2019-11-05
Support for STM32L4, SM32L1 option bytes write.

zx81a:
1) STM32L4 has single 32-bit word with all needed options. I just works.
2) STM32L1 is a bit more complicated. There are two 32-bit words to modify. Sometimes you need to change only one. When reprogramming a device with RDP level 1 you only have to change RDP to 0,
everything else is erased. But even with RDP 0 you can't reflash large code while IWDG_SW is cleared. So I've decided to allow writing to 0x1ff80000 and 0x1ff80004.


Nightwalker-87:
Redirected to develop branch. Closes #847.

Nightwalker-87:
@zx81a: Thx for your contribution!

PR #853, opened 2020-01-07
Fix dead link in README.md

piccaso:
I noticed this dead link and would like to suggest to remove the release name (cosmic) from the link.  
In that case packages.ubuntu.com will list the releases where this package is available.


Nightwalker-87:
Fixed in upcoming Release 1.6.0.

PR #858, opened 2020-01-29
[doc/trivial] Fix st-flash manpage read example

yumkam:
SIZE must be specified as hexdecimal number (as per both documentation and implementation), current manpage example unexpectedly reads 0x4096 (16534) bytes instead of 4096 (0x1000).
(actually, 0x prefix is not necessary, number interpreted as hexdecimal anyway, and maybe confusing readers into thinking that number will be interpreted as decimal [without prefix] or octal [with 0 prefix])

### DIFF
--- a/doc/man/st-flash.1
+++ b/doc/man/st-flash.1
@@ -65,7 +65,7 @@ Read firmware from device (4096 bytes)
 .IP
 .nf
 \f[C]
-$ st\-flash read firmware.bin 0x8000000 4096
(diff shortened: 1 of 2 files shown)


Nightwalker-87:
Merged into develop branch with commit fa35d5b7ba365a4509e97f07c5c43d0ddb74657f.

Nightwalker-87:
@yumkam: Thx for your contribution!

PR #861, opened 2020-02-14
Fix stlink support with no mass storage

pinaraf:
Add support for the STLink-2.1 when flashed with no mass storage

Since STLink 2.33.25, it is possible to flash the STLink in two modes:
− STM32 Debug + VCP
− STM32 Debug + Mass storage + VCP

When flashed in the first mode, the device id changes, but the behaviour remains the same.
This commit thus introduces support for this new device id. Also, fix broken installation by including the stm32.h file while I'm at it.


Nightwalker-87:
Merged into develop branch with commit 8cd30690643dc8117c715f2484e9e6799818ba53.

Nightwalker-87:
@pinaraf: Thx for your contribution!

Nightwalker-87:
This PR has been manually merged with commit 8cd30690643dc8117c715f2484e9e6799818ba53.

PR #882, opened 2020-03-19
Fixed wait-loop for flash_loader_run() & F72/F73 device description

### LINKED ISSUE #290: flash loader run problem....
Looks similar to issue #269 but is not. 

It seems this is hardware dependent: One board has this issue, the other does not. 

About (exactly?) half the times I run a "reset-load-continue", I get: 
    2015-01-07T17:28:03 INFO src/stlink-common.c: Successfully loaded flash loader in sram
    2015-01-07T17:28:23 ERROR src/stlink-common.c: flash loader run error
and then the load/flash fails. 

As an extra "hint" as to what's happening: It seems to be NOT booting from SRAM but from my flash: It prints the welcome message on the serial port. On the other hand, the CPU seems halted shortly after booting: I cannot interact with the program. 

Any idea how this could come to be? (I have boot0 pulled low: boot from flash... Hmm. It seems I've got "boot0" floating on the board that works.... Never worried about it because it simply works).

rewolff:
It seems my code also has the F72 fix. Do you really need that in a separate pull request?

Looking at that: Shouldn't it be "F72x / F73x device"? Wouldn't that be clearer? All devices have a 3-digit type number. I noticed that I often use the F072, but thats a different beast than the F72x and F730....

### DIFF
--- a/src/chipid.c
+++ b/src/chipid.c
@@ -27,7 +27,7 @@ static const struct stlink_chipid_params devices[] = {
         {
             //RM0431 and DS document was used to find these paramaters
             .chip_id = STLINK_CHIPID_STM32_F72XXX,
-            .description = "F72 device",
+            .description = "F72/F73 device",
             .flash_type = STLINK_FLASH_TYPE_F4,
             .flash_size_reg = 0x1ff07a22,      // section 35.2
             .flash_pagesize = 0x800,           // No flash pages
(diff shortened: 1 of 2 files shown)


Nightwalker-87:
I wonder if there are more open issues that are addressed by this...

Nightwalker-87:
@rewolff: Thx for your contribution!

PR #886, opened 2020-03-21
Silence unnecessary messages (Fixes #240)

### LINKED ISSUE #240: Reading the chipid gives nonzero "random" number.
When starting st-util on my F4 discovery board, I got: 

 WARN src/stlink-common.c: unknown chip id! 0xe0042000
 Chip ID is 00000000, Core ID is  00000000.

which I know spells trouble. But when I see some non-zero numbers for the "chip id" I think I have a connection to the target. Turns out that is not the case: When I switched back to my target I got the same 0xe0042000 chip id, even when the target was not powered. 

My STM32F407 on my discovery board might be fried, I don't know yet. I'll just use the F4DISCOVERY as my STLINK for now. 

It'd be nice if the "detection" of "no chip connected" was a bit better. Apparently "chip id = 0xe0042000, ... core id = 0" means "no chip"....

rewolff:
As requested. The startup of st-util is now: 
```
assurancetourix:~> ~/stm32/stlink.git2/build/Release/src/gdbserver/st-util -m
st-util 1.6.0-33-gb8813fa-dirty
2020-03-21T16:07:15 INFO common.c: F0xx small: 4 KiB SRAM, 16 KiB flash in 1 KiB pages.
2020-03-21T16:07:15 INFO gdb-server.c: Listening at *:4242...

```
The user may want the "listening at" message and the "what's connected" message has been reduced from two lines to just one. It now contains just the info that normal users may want to know.: stuff that helps determine if say the right target is connected. 

I removed all the "device" parts of the strings in the description of the chipID table. 
Some of the "description" entries in the table had a check for say a CHIPID_F412 and then proceded to call the device an F4xx. Lets just call it F412 then. But maybe there are other devices that share the same chipID let me know if you find one.

### DIFF
--- a/src/chipid.c
+++ b/src/chipid.c
@@ -5,7 +5,7 @@ static const struct stlink_chipid_params devices[] = {
         {
             //RM0410 document was used to find these paramaters
             .chip_id = STLINK_CHIPID_STM32_F7XXXX,
-            .description = "F76xxx device",
+            .description = "F76xxx",
             .flash_type = STLINK_FLASH_TYPE_F4,
             .flash_size_reg = 0x1ff0f442,      // section 45.2
             .flash_pagesize = 0x800,           // No flash pages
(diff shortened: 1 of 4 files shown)


Nightwalker-87:
Info: Merge is going to wait until Travis CI maintenance has successfully concluded. Used services are currently limited or are not available at all.

Nightwalker-87:
@rewolff: Thank you for your efforts!

PR #889, opened 2020-03-21
Error return in failed probe

rewolff:
playing around with my recent/ patched sources I found that  some errors were treated as "warning". Turns out a lot of "ret=-1" were missing. Someone incompletely changed the code. I took the easy way out and added an if where it belonged and changed the on_error label to always return error.


Nightwalker-87:
@rewolff: Closed?

rewolff:
I noticed that this contained the patches for the other stuff as well. So I researched how to split that out (took me a while) and then created a separate pull request (#890)  for just this issue.

PR #894, opened 2020-03-25
setting up a libusb log level accordingly to verbosity.

slyshykO:


Nightwalker-87:
Shall we have some documentation for this besides the inline-comments?

Unfortunately there is a problem with Travis CI _again_ which prevents build tests to take place for this project at all currently. :-( As it seems it is related to git project permissions, I've consulted @texane on this and hope to get a reply soon. I'll merge when this is working again.

slyshykO:
I think it deeply internal changes and no needed user docs for it.

src/sg.c
```diff
@@ -936,6 +936,12 @@ static stlink_t* stlink_open(const int verbose) {
         return NULL;
     }
 
+#if LIBUSB_API_VERSION < 0x01000106
+    libusb_set_debug(slsg->libusb_ctx, ugly_libusb_log_level(verbose));
+#else
+    libusb_set_option(slsg->libusb_ctx, LIBUSB_OPTION_LOG_LEVEL, ugly_libusb_log_level(verbose));
+#endif
+
```
Nightwalker-87: Ah, so this is the correct condition compared to the former and no longer executed one, based on `LIBUSBX_API_VERSION`, in commit e3c76aa67315db32bd303303023d6c74c51e900b. :+1:

slyshykO: correct

Nightwalker-87:
Changed base-branch to branch `ext_pkg_versions` that will deal with all libusb and cmake stuff.

_Beware:_ This skips the automatic Travis CI test build which is not working currently, so please ensure local testing with `sudo ./.travis.sh` instead until this is fixed.

PR #895, opened 2020-03-25
[compatibility] update of libusb

slyshykO:


Nightwalker-87:
Oh, jolly good! I planned to take action on this topic as well. :-)

What is the reason for raising to the latest release available?
Are there any troubles with earlier versions? (I'm thinking about compatibility here).
Should we raise version requirements for unix-based systems as well?
Let me check the current `libusb` package versions for the most common distributions.

Whatever, we also need to update the documentation together with this change.
As far as I can see this affects `/doc/compiling.md`.

slyshykO:
This is not obligatory changes. They apply only in a rare case when building for windows and libusb not found. For example in CI.
I think at this point of time no need to increase the requirement for libusb versions.

Nightwalker-87:
Ok, let me check currently supplied `libusb` versions from the repos.

I think that we should set the version requirement in such a way that the oldest currently supported operating system versions of common linux/unix distributions are capable of running our stlink tools. This would also define the oldest `libusb` version to support (unless there is a bug that causes problems of course). I don't assume it to be a good idea to define different `libusb` versions for different OSs, as it likely raises confusion...

Nightwalker-87:
Here is the result from https://pkgs.org/search/?q=libusb (as of Mar 2020):

pine Edge - 1.0.23
Alpine 3.11 - 1.0.23
Alpine 3.10 - 1.0.22
Alpine 3.9 - 1.0.22
ALT Linux Sisyphus - 1.0.23
Arch Linux - 1.0.23
CentOS 8 - 1.0.22 libusbx _--> not compatible_
CentOS 7 - 1.0.21 libusbx _--> not compatible_
CentOS 6 - 1.0.**9** _--> will not be supported_
Debian Sid - 1.0.23
Debian 10 (Buster) - 1.0.22
Debian 9 (Stretch) - 1.0.**21**
Debian 8 (Jessie) - 1.0.**19** _--> will not be supported_
Fedora Rawhide - 1.0.22 libusbx _--> not compatible_
Fedora 31 - 1.0.22 libusbx _--> not compatible_
Fedora 30 - 1.0.22 libusbx _--> not compatible_
FreeBSD 13 - linux_libusb-13.0r358841 --> compatible (?)
FreeBSD 12 - linux_libusb-11.0r261448_4 --> compatible (?)
FreeBSD 11 - linux_libusb-11.0r261448_4 --> compatible (?)
KaOS - 1.0.23
Mageia Cauldron - 1.0.22
Mageia 7.1 - 1.0.22
NetBSD 9.0 - 1.0.22
NetBSD 8.1 - 1.0.22
NetBSD 7.2 - 1.0.22
OpenMandriva Cooker - 1.0.23
OpenMandriva Lx 4.1 - 1.0.23
OpenMandriva Lx 3.0 - 1.0.**20**
openSUSE Tumbleweed - 1.0.23
openSUSE Leap 15.2 - 1.0.**21**
openSUSE Leap 15.1 - 1.0.**21**
OpenWrt 19.07 - 1.0.22
OpenWrt 18.06 - 1.0.22
Slackware Current - 1.0.23
Slackware 14.2 - 1.0.**20**
Slackware 14.1 - 1.0.**9** _--> will not be supported_
Solus - 1.0.23
Ubuntu 20.04 LTS - 1.0.23
Ubuntu 19.10 - 1.0.23
Ubuntu 18.04 LTS - 1.0.**21**
Ubuntu 16.04 LTS - 1.0.**20**
Ubuntu 14.04 LTS - 1.0.**17** _--> will not be supported_

Looking at this may go to `libusb` **1.0.20** to ensure widespread compatibility.

Nightwalker-87:
@Vascom: You are on Fedora right? What do you think about this? Why does Fedora still use `libusbx` BTW? Is your library that outdated as well?

Vascom:
@Nightwalker-87 I don't know why it still not updated in rawhide. Need to ask maintainer.
But even if this update will be done it will be never updated for current releases of Fedora and RHEL/CentOS.

slyshykO:
Looks like starting from this commit https://src.fedoraproject.org/rpms/libusbx/commits/master Fedora switched from libusbx to libusb but don't change the package name.

Vascom:
No.
libusb - for compatibility with 0.1 API.
libusbx - main libusb as I know.

slyshykO:
@Vascom Yes. My point is they (Fedora) use sources from libusb-project for libusbx-package

Vascom:
# upstream libusbx has merged back with libusb and is now called libusb again
# but we already have a libusb package for the old libusb-compat-0.1, renaming
# that to libusb-compat while at the same time giving this its name is a bit
# tricky, lets stick with the libusbx name for now

Nightwalker-87:
Hi there, moved out the started discussion to #897 including all replies. Please continue there otherwise this PR is recycled for a basic discussion of this topic. Cleaning this up now.

cmake/modules/FindLibUSB.cmake
```diff
@@ -67,7 +67,7 @@ if(NOT LIBUSB_FOUND)
     if(WIN32 OR MSVC OR MINGW OR MSYS)
         find_package(7Zip REQUIRED)
 
-        set(LIBUSB_WIN_VERSION 1.0.22)
+        set(LIBUSB_WIN_VERSION 1.0.23)
```
Nightwalker-87: I think we should go to 1.0.20 here, unless there is no significant reason not to do so.

slyshykO: My opinion is we should use the newest libusb that we can. So not see reasons to downgrade.

Nightwalker-87: Let me have a look at how this check is implemented at all...

Nightwalker-87: Yes, we should get the newest version here, if no previous version >= 1.0.20 is detected.

Nightwalker-87:
For windows the workflow could look like this:
- check if libusb is present:
   - [case 1]: yes --> which version? --> < 1.0.20 ? --> ignore present and install 1.0.23 (newest)
   - [case 2]: yes --> which version? --> >= 1.0.20 ? --> use present
   - [case 3]: no: install 1.0.23 (newest)

The version which serves as the condition should be 1.0.20 for now, with the option to raise it at a later time (but then synchronous with the requirement for linux-based systems as proposed in #897).

Nightwalker-87:
Feel free to extend this PR with related stuff, I'll leave it open for another while, but will submit own changes afterwards.

CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ set(STLINK_HEADERS
 	include/stlink/mmap.h
 	include/stlink/chipid.h
 	include/stlink/flash_loader.h
+    include/stlink/stlinkusb.h
```
Nightwalker-87: Indentation?

slyshykO: tab/space HELLLLL 😁

cmake/modules/FindLibUSB.cmake
```diff
@@ -79,6 +79,7 @@ if(NOT LIBUSB_FOUND)
             file(DOWNLOAD
                 https://sourceforge.net/projects/libusb/files/libusb-1.0/libusb-${LIBUSB_WIN_VERSION}/libusb-${LIBUSB_WIN_VERSION}.7z/download
                 ${LIBUSB_WIN_ARCHIVE_PATH}
+                EXPECTED_MD5 750e64b45aca94fafbdff07171004d03
```
Nightwalker-87: Good idea, but what happens when a new version is released? The checksum will change, won't it? I don't want to make in very complicated, but can we retrieve it automatically? What do you think?

src/usb.c
```diff
@@ -2,19 +2,12 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdint.h>
+#include <limits.h>
```
Nightwalker-87: Where is this include used in `usb.c`?

cmake/modules/FindLibUSB.cmake
slyshykO: Yes, when someone increases libusb version also should update MD5 accordingly.
But this prevents errors in archives.
Like this https://github.com/texane/stlink/issues/896#issuecomment-603877423 .

src/usb.c
slyshykO: Try to resolve this https://github.com/texane/stlink/issues/897#issuecomment-604588214

lhondareyte:
That's ok for me : `limits.h` include `sys/limits.h`.

include/stlink/stlinkusb.h
```diff
@@ -0,0 +1,38 @@
...
+#error unsupported libusb version
+#endif
+#endif
+
+#define MINIMAL_API_VERSION 0x01000102
+
+#if ( LIBUSB_API_VERSION < MINIMAL_API_VERSION )
+#error unsupported libusb version
+#endif
+
```
Nightwalker-87: Check typo to avoid bug: LIBUSBX_API_**V**VERSION --> LIBUSB_API_VERSION

... but we may simplify this (suggestion):

/* for all systems */
#if defined ( LIBUSBX_API_VERSION )
#error unsupported libusb version
#endif

/* FreeBSD */
// libusb is integrated in FreeBSD, only ensure 1.0.26 or later codebase...
#if defined (__ FreeBSD __)
#define MIN_LIBUSB_API_VERSION 0x01000102 
#if ( LIBUSB_API_VERSION < MIN_LIBUSB_API_VERSION ) 
#error unsupported libusb version
#endif
#endif

/* Linux-Distributions */
// install-package used --> version check via cmake compilation routine

/* macOS */
// TBD

/* Windows */
// external install-package used --> version check and fetch via cmake compilation routine if necessary

Nightwalker-87: See also #733 regarding LIBUSBX_API_VERSION. We don't need it anymore at all not even for FreeBSD.
By now (#897) we would solely define libusb versions that don't have it in their codebase anymore.

src/usb.c
Nightwalker-87: Ah, thx :+1:

cmake/modules/FindLibUSB.cmake
```diff
@@ -79,6 +79,7 @@ if(NOT LIBUSB_FOUND)
             file(DOWNLOAD
                 https://sourceforge.net/projects/libusb/files/libusb-1.0/libusb-${LIBUSB_WIN_VERSION}/libusb-${LIBUSB_WIN_VERSION}.7z/download
                 ${LIBUSB_WIN_ARCHIVE_PATH}
+                EXPECTED_MD5 cf3d38d2ff053ef343d10c0b8b0950c2
```
Nightwalker-87: This would change on every new release, wouldn't it?
Can we retrieve it from the repo? That would be very friendly for maintenance. ^^

slyshykO: Libusb project produces new releases approx 1 per year. Also, we usually no need to update it quickly after release. So, I think, there is no problem for the maintainer to download archive and check MD5. And not, they don't provide checksums for archives.

include/stlink/stlinkusb.h
slyshykO: this is incorrect
```c
#if defined ( LIBUSBX_API_VERSION )
#error unsupported libusb version
#endif
```
because it really defined for compatibility.

So, I prefer stay this without changes at this time.

cmake/modules/FindLibUSB.cmake
Nightwalker-87: Yes, no doubt - I did not question whether to have it or not...
I was only on about how to implement it, but referring to your mention in the main thread:
If it's only once per year an average we shall be fine and can do this manually.

include/stlink/stlinkusb.h
Nightwalker-87: Yes, you are right I have convinced myself which 1.0.18 as a reference.
If we don't keep `LIBUSBX_API_VERSION` until this define finally vanishes from `libusb.h`, they will all fail to install.
Good that we had a separate look at it.

Nightwalker-87: We may do it like this then:

```
/* libusb is integrated in FreeBSD */
#if defined (__FreeBSD__)

    /* Verify API-Type */
    #if !defined (LIBUSBX_API_VERSION)
        // (still) defined in libusb-codebase for compatibility (to be deprecated in the future)
        #define LIBUSBX_API_VERSION LIBUSB_API_VERSION
    #elif !defined (LIBUSB_API_VERSION)
        #error unsupported libusb version // ... we find to have a libusbx-codebase
    #endif

    /* Verify API-Version */
    #define MIN_LIBUSB_API_VERSION 0x01000102
    #if ( LIBUSB_API_VERSION < MIN_LIBUSB_API_VERSION )
        #error unsupported libusb version // ... we find libusb-codebase to be older than 1.0.16
    #endif
#endif
```

/* Linux-Distributions */
// install-package used --> version check via cmake compilation routine

/* macOS */
// TBD

/* Windows */
// external install-package used --> version check and fetch via cmake compilation routine if necessary

cmake/modules/FindLibUSB.cmake
Nightwalker-87: As mentioned before, it shall be ok, if there is one release per year on average.
Maybe it makes sense to have a reminder on that for yearly maintenance.

include/stlink/stlinkusb.h
slyshykO: I think current checks are enough at this moment.
If issues will be founded we fix them.

Nightwalker-87: Discussed this with a conclusion resulting in commit db6e93e52aa86add8ff0453ca2d26ca13310fa8e.

PR #901, opened 2020-03-31
stm32g4 chips update

grevaillot:
- Add id and infos for stm32g47/48 chips.
- Update comments regarding g43/g44

I followed current g43x scheme of considering sram1/sram2/ccm sram as contiguous, but documented it. 

Note that I've only worked with the first of the two banks. stm32g474 can also be configured as a single 512kb bank, with 4k pages via option bytes. but i think there's no logic to handle that properly ?

### DIFF
--- a/src/chipid.c
+++ b/src/chipid.c
@@ -550,6 +550,9 @@ static const struct stlink_chipid_params devices[] = {
             .flash_type = STLINK_FLASH_TYPE_G4,
             .flash_size_reg = 0x1FFF75E0,    // Section 47.2
             .flash_pagesize = 0x800,         // 2K (sec 3.3.1)
+            // SRAM1 is 16k at 0x20000000
+            // SRAM2 is 6k at 0x20014000
+            // SRAM3/CCM is 10k at 0x10000000, aliased at 0x20018000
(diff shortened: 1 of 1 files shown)


Nightwalker-87:
Thank you for keeping this up to date.
We shall have some additional testing here...
As I can't help out here due to missing hardware, we should ask the community for assistance.
Maybe someone also has an idea regarding your note.

Nightwalker-87:
> Note that I've only worked with the first of the two banks. stm32g474 can also be configured as a single 512kb bank, with 4k pages via option bytes. but i think there's no logic to handle that properly ?

Looking at this again and the respective datasheet (page 19) this would likely make use of the MCU's S-bus. If this should require option bytes to be set, we don't need to address this with this PR necessarily. I'm not familiar with this device though.

Nightwalker-87:
@grevaillot How do you think about this?

Nightwalker-87:
ping @grevaillot

grevaillot:
i think that PR is okay as it. i'll push some enhanced support for Gx devices with option read/write and dynamic dual bank flash handling soom, in a new pr.

Nightwalker-87:
Thanks!

PR #907, opened 2020-04-06
Add uninstall target to generated Makefile (fixes #619)

### LINKED ISSUE #619: [install] There is no uninstall target created by CMake
I'm upgrading from v1.3.1 to v1.4.0 in my Ubuntu 16.04 and I want to do a clean install. So I entered the v.1.3.1 folder that I had previously compiled and generated the Makefile and ran `make uninstall` but the rule uninstall doesn't exist. Is there an easy way to uninstall it that doesn't involve searching and deleting all files manually from my system?

FreddieOliveira:
When compiling from source, the Makefile generated by running `cmake ../..` after `git clone https://github.com/texane/stlink && mkdir -p stlink/build/Release && cd stlink/build/Release` now has a target called *uninstall* that basically deletes the files added to the system by `make install`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,6 +214,17 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
 	endif()
 endif()
 
+# uninstall target
+if(NOT TARGET uninstall)
+  configure_file(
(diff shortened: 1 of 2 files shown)


FreddieOliveira:
Closing since this was already pushed into `ext_pkg_versions` branch with this commit:  e7ed76bd1

PR #908, opened 2020-04-06
Fix int length issues

chenguokai:
Make stlink build on 32 bit systems.

Build passed on ubuntu 16.04 i386 and macOS Mojave 10.14.6 AMD64.

`make test` failed for flash, which seems to be the same as current master.

Related issue:

[https://github.com/texane/stlink/issues/629](https://github.com/texane/stlink/issues/629)


Nightwalker-87:
Thx for your contribution!

PR #910, opened 2020-04-06
stm32gx improvements

grevaillot:
Hi,

here are a few patches to improve g4/g0 support
 - introduce a helper to handle devices with dual bank flash
 - handle cat3 g4 devices with configurable dual bank flash
 - enable mass erase for g4 and g0 devices. 
 - add a flash programming check at the end of mass erase.

has_dual bank flag could probably avoid multiple device type check around the code with a bit of work. 
flash programming check could probably be added in page erase and flash write and extended to other devices, to debug some silent issues.


Nightwalker-87:
Merci :1st_place_medal:

PR #913, opened 2020-04-08
Fix dead loop after an unexpected unplug

chenguokai:
* Fix by adding error handler in gdb-server
* Enhance logs for functions with return values

Fixes:
https://github.com/texane/stlink/issues/780#event-3213582141


Nightwalker-87:
Closes #812.

Nightwalker-87:
Note: There is a build failure on linux related to the `./.travis-lin-mingw.sh` script.
However this is not related to this PR and will be solved afterwards.

Issue #609, opened 2017-06-29
add to documentation: st-flash write is ignoring 0xff at end of file


ponty:
- [X] Programmer/board type: `Stlink/v2`
- [x] Programmer firmware version: 
- [X] Operating system: `Linux Ubuntu 16.04`
- [X] Stlink tools version and/or git commit hash: `v1.3.0-9-g88c6162`
- [X] Stlink commandline tool name: `st-flash`
- [X] Target chip (and optional board): `STM32F415RGT6`

Command:
```
st-flash write firmware.bin 0x8000000
```

Output:

```
Loading device parameters....
Device connected is: F4 device, id 0x10076413
SRAM size: 0x30000 bytes (192 KiB), Flash: 0x100000 bytes (1024 KiB) in pages of 16384 bytes
Ignoring 49004 bytes of 0xff at end of file
Attempting to write 213652 (0x34294) bytes to stm32 address: 134217728 (0x8000000)
EraseFlash - Sector:0x0 Size:0x4000 EraseFlash - Sector:0x1 Size:0x4000 EraseFlash - Sector:0x2 Size:0x4000 EraseFlash - Sector:0x3 Size:0x4000 EraseFlash - Sector:0x4 Size:0x10000 EraseFlash - Sector:0x5 Size:0x20000 2017-06-29T09:38:50 INFO /home/gerecsl/aha/workspace/ext/stm32/stlink/src/common.c: Finished erasing 6 pages of 131072 (0x20000) bytes
Starting Flash write for F2/F4/L4
Successfully loaded flash loader in sram
Starting verification of write complete
Flash written and verified! jolly good!
```

After reading the man page I expected that the whole file will be written, 
but the end of the file was ignored because of 0xFF values.
So in this special case an `erase` should be executed first always.
I think that the documentation could have a warning for this.

I tried QSTLink2 for write, and there was no such feature.

xor-gate:
Sorry for the late reply, but you could propose a pull request for updating the documentation.

Nightwalker-87:
@ponty or anyone else: Please re-test this with Releases v1.4.0 up to v1.6.0 to check if this has been fixed in the meanwhile. If not, please submit a pull request.

Nightwalker-87:
As far as I read, EOF (0xff 0xff) is returned by fgetc() due to end of input, so I assume that all further characters beyond are ignored, but I am not sure yet if this is the correct explanation for this behaviour.

I found the following commits that may relate to this:
[...]

PR #917, opened 2020-04-10
Fix st-flash write 0xff ignored issue

chenguokai:
strcpy function will copy string with the ending NULL while strlen will not count NULL into length.
Added one commandline argument `--opt` to enable optimization of ending empty bytes.


chenguokai:
@Nightwalker-87  This fix is not linked so tightly with the 0xff issue.
Just a drop-by fix when I was checking the argument processing procedure.

Nightwalker-87:
Well that does not matter, I may leave it open for a bit...

Nightwalker-87:
... what would allow you to add your proposal from #609 to it.
BTW: Since `develop` is now the default branch in this project, every PR will close attached issues per automation. So this is actually the real background to it.

src/tools/flash_opts.c
```diff
@@ -93,30 +96,30 @@ int flash_get_opts(struct flash_opts* o, int ac, char** av) {
...
+				    break;
+			    default:
+				    fprintf(stderr,"Invalid --flash=%s\n",arg);
+				    return -1;
+			    }
+		    }
+	    }
+	    else {
             break;  // non-option found
         }
```
Nightwalker-87, line 125: Minor note: Could you use 4 spaces instead of tab-spaces for indentation? The latter results in inhomogeneous appearance on different systems that affects readability. I know this is not consistent yet throughout the codebase, but it will be rather sooner than later. I'll define this in our new contribution guidelines as well.

Nightwalker-87:
The build fail derives from the script .travis-lin-mingw.sh. I know about this, so you can simply delete that file as I did in `develop` as well. I am addressing this on a separate branch.

src/tools/flash_opts.c
Nightwalker-87: Thx.

chenguokai: Switched to four spaces at the latest commit, thanks!

chenguokai:
`.travis-lin-mingw.sh` removed

Nightwalker-87:
Anything else still open on your side or are you satisfied with this now as it is?

chenguokai:
The current state is suitable for a final merge, except for minor puzzles I mentioned in https://github.com/stlink-org/stlink/issues/609
If you think that feature is no problem for now, no more things open for me.

Nightwalker-87:
Ah there is a leftover in .travis.yml if forgot about, damn it. :-/

chenguokai:
Should I perform a deletion in my PR?

Nightwalker-87:
Yeah, remove the reference to that old mingw build script. Hopefully this makes CI builds finally succeed... Sry for that, this was my mistake.

Nightwalker-87:
Thanks for your efforts! :tada:

Issue #919, opened 2020-04-11
Build on 32-bit systems is broken again


chenguokai:
Thank you for giving feedback to the stlink project.

In order to allow developers and other contributors to isolate and target your respective issue, please take some time to fill out each of the following items appropriate to your specific problem:
- Operating system and version: [Mac OS X] (e.g Linux, Mac OS X, Windows)
- **Stlink tools version** and/or git commit hash: [ecbbd6db7aa8dff983ee43f629ac920f20995d12] (e.g v1.1.0/git-c722056)

Futher we kindly ask you to describe the detected problem as detailed as possible and to add debug output if available, by using the following template:

Commandline-Output:

```
...
Scanning dependencies of target stlink-static
[  3%] Building C object CMakeFiles/stlink-static.dir/src/chipid.c.o
[  6%] Building C object CMakeFiles/stlink-static.dir/src/common.c.o
cc1: warnings being treated as errors
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c: In function ‘set_flash_cr_pg’:
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c:440: warning: implicit conversion shortens 64-bit value into a 32-bit value
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c: In function ‘write_flash_cr_bker_pnb’:
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c:758: warning: implicit conversion shortens 64-bit value into a 32-bit value
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c:759: warning: implicit conversion shortens 64-bit value into a 32-bit value
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c:763: warning: implicit conversion shortens 64-bit value into a 32-bit value
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c:763: warning: implicit conversion shortens 64-bit value into a 32-bit value
/Users/xim/Desktop/Parallels Shared Folders/Home/Documents/macports-maintain/stlink/src/common.c: In function ‘calculate_L4_page’:
[...]

Issue #909, opened 2020-04-06
[refactoring] Clean code with unified variable type


chenguokai:
As far as I can see, many functions use fixed-length variables. It is indeed useful for functions that need to deal with fixed data format. However the abuse of fixed-length variables are causing potential build failures and inessential casts.

For example in src/usb.c:
<img width="842" alt="image" src="https://user-images.githubusercontent.com/11240171/78568947-704eff00-7855-11ea-965e-0c5c05c2fefc.png">

send_recv accept `txsize` and `rxsize` as `size_t` but in the actual context, those arguments have to be casted to `int` or `unsigned int`, without a fixed length, and whoever calls it provides arguments of a different type(`uint32_t` or magic numbers in this case). unintentional cast from 32 bit to 64 bit is applied at the calling process. Also, to prevent a build failure, many explicit casts are required.

I suggest replacing fixed length variables according to their actual usages. 
Staying fuzzy to be more portable.

I would like to give it a try if others find it meaningful.

Nightwalker-87:
Thx for putting this up on the agenda. As it is a general problem throughout the codebase, I'm not sure at the moment if we should put that into v1.6.1 straight away. @martonmiklos previously requested to do some cleanup related to code style, which I pushed into the v1.6.2 milestone. Maybe this would be a good place for this issue as well, though it will push it out a bit further.

chenguokai:
Maybe I can do some cleanup locally first and follow up the develop process.😃

Nightwalker-87:
Well of course you can, but please ensure to branch off from `develop` and keep following this branch, instead of `master`, which is no longer used for frequent changes, but only for releases and hotfixes.

Nightwalker-87:
With respect to compatibility between ILP32, LLP64 and LP64 I think it would be a good idea to replace the data type `long` with `int32_t`, but if one wants to rule out surprises in the future related to conversion, we may consider to move to fixed width integer types wherever possible...

Nightwalker-87:
@chenguokai: I'll leave this to you. You may contribute, if you fancy and have some some time.

chenguokai:
Sure I will, before the release. Getting busy now.😔

Nightwalker-87:
Is there anyone else who can deal with this?

Nightwalker-87:
As there is no common quality level of coding throughout the codebase, the mentioned explicit casting may be excessively used in some parts of the code while others appear to be designed and written quite well.

[...]

PR #920, opened 2020-04-11
Fixed broken build on 32-bit systems

chenguokai:
Fixes https://github.com/stlink-org/stlink/issues/919

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -190,10 +190,10 @@
 #define STM32L4_FLASH_CR_OBL_LAUNCH 27  /* Option bytes reload */
 // Bits requesting flash operations (useful when we want to clear them)
 #define STM32L4_FLASH_CR_OPBITS                                     \
-    ((1lu<<STM32L4_FLASH_CR_PG) | (1lu<<STM32L4_FLASH_CR_PER)       \
+    (uint32_t)((1lu<<STM32L4_FLASH_CR_PG) | (1lu<<STM32L4_FLASH_CR_PER)       \
     | (1lu<<STM32L4_FLASH_CR_MER1) | (1lu<<STM32L4_FLASH_CR_MER1))
 // Page is fully specified by BKER and PNB
-#define STM32L4_FLASH_CR_PAGEMASK (0x1fflu << STM32L4_FLASH_CR_PNB)
+#define STM32L4_FLASH_CR_PAGEMASK (uint32_t)(0x1fflu << STM32L4_FLASH_CR_PNB)
 
 #define STM32L4_FLASH_OPTR_DUALBANK     21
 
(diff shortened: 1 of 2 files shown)


src/common.c
```diff
@@ -437,7 +437,7 @@ static void set_flash_cr_pg(stlink_t *sl) {
         x |= 1 << FLASH_CR_PG;
     } else if (sl->flash_type == STLINK_FLASH_TYPE_L4) {
         cr_reg = STM32L4_FLASH_CR;
-        x &= ~STM32L4_FLASH_CR_OPBITS;
+        x &= (uint32_t)~STM32L4_FLASH_CR_OPBITS;
         x |= 1 << STM32L4_FLASH_CR_PG;
```
slyshykO, line 441: Maybe it is more useful to make definition of STM32L4_FLASH_CR_OPBITS typed?
For example like this : 
```cpp
#define STM32L4_FLASH_CR_OPTLOCK  30u	/* Lock option bytes */
```

or

```cpp
#define STM32L4_FLASH_CR_OPTLOCK  (uint32_t)30	/* Lock option bytes */
```

src/common.c
```diff
@@ -2000,7 +2000,7 @@ int stlink_erase_flash_mass(stlink_t *sl) {
     if (sl->flash_type == STLINK_FLASH_TYPE_L0 ||
         sl->flash_type == STLINK_FLASH_TYPE_WB) {
         /* erase each page */
-        int i = 0, num_pages = (int) sl->flash_size/sl->flash_pgsz;
+        int i = 0, num_pages = (int) sl->flash_size/(int)(sl->flash_pgsz);
```
slyshykO, line 2003: Think brackets should help there.
```cpp
int i = 0, num_pages = (int) (sl->flash_size / sl->flash_pgsz);
```

src/common.c
chenguokai: Indeed better, changed

slyshykO:
I think we should add such configuration to Travis build.
I mean configuration that causes issues, that this PR try to fix.

slyshykO:
I think this config should be 64bit Linux with 32bit clang.

Nightwalker-87:
Yes @slyshykO, it should be regularly tested with Travis. I agree on that.

Nightwalker-87:
Thanks for fixing this! :+1: 

To be honest, we should change some variables to the appropriate sizes / datatypes. The requirement for manual casts not only here, but at several places throughout the codebase, indicates that something is not as it is supposed to be per design. If implemented nicely, there hardly is any need for explicit casting. I'd like to see that addressed in v1.6.2 at a later point in time, maybe within #909.

Issue #650, opened 2017-10-28
STM32F103C8T6: ERROR src/common.c: map_file()==-1


renn0xtek9:
I got an error during flash operation. I don't know what it is/means.
The compilation has no problem. Flashing another binary (a simple thing toggling a LED) is not problem either.

- [x] Programmer/board type: Stlink/v2, clone
- [x] Operating system: e.g linux ubuntu 16.04
- [x] Stlink tools version and/or git commit hash: e.g git-a2a707e4f75fd8aafce3fa6095dc34206c5bb801
- [x] Stlink commandline tool name:  `st-flash write STM32USASRT_IR_StdPeriph.bin 0x8000000`
- [x] Target chip (and optional board): STM32F103C8T6

A as-detailed description possible of the problem with debug output when available.

Output:

```
2017-10-28T19:54:26 ERROR src/common.c: map_file() == -1

```

Expected/description:
I expected to work as usual

xor-gate:
I'm not sure why the mapping of the binary file to memory fails at your system. The line is located here: https://github.com/texane/stlink/blob/a2a707e4f75fd8aafce3fa6095dc34206c5bb801/src/common.c#L1164

ugifractal:
not sure if your case same with mine.
but if st-flash integrated with other tools, when filename parameter is blank, above error may raised.

here is mine
```
sugiartos-MacBook-Pro:~ sugiarto$ st-flash write ""  0x8000000
st-flash 1.5.1-18-g522188d
2019-03-15T13:41:09 INFO common.c: Loading device parameters....
2019-03-15T13:41:09 INFO common.c: Device connected is: F1 Medium-density device, id 0x20036410
2019-03-15T13:41:09 INFO common.c: SRAM size: 0x5000 bytes (20 KiB), Flash: 0x20000 bytes (128 KiB) in pages of 1024 bytes
open() == -1
2019-03-15T13:41:09 ERROR common.c: map_file() == -1
stlink_fwrite_flash() == -1
sugiartos-MacBook-Pro:~ sugiarto$ 
```

chenguokai:
@Nightwalker-87 I am not quite sure about what I need to do.
If the original issue is indeed the one caused by an empty/non-existing filename, I will be able to fix. Otherwise, without further information, I cannot figure out the problem.

Nightwalker-87:
[...]

Issue #879, opened 2020-03-16
st-flash: incorrect file name handling


tdjastrzebski:
In case of incorrect bin file name **st-flash** fails with error:
`ERROR common.c: map_file() == -1`
I suggest more descriptive message.

Nightwalker-87:
So what would you suggest instead?

tdjastrzebski:
I suggest message:
`Unable to open [file-name] file.`

Nightwalker-87:
Ok, are you with Release v1.6.0 ?

tdjastrzebski:
So far so good :)
It seems to work better than 1.5.1 although I have not tested it extensively yet.

Nightwalker-87:
@tdjastrzebski: Can you submit a PR?

Nightwalker-87:
Duplicate of #650.

PR #921, opened 2020-04-12
Enhance error log with file path for map_file()

chenguokai:
Trying to improve for https://github.com/stlink-org/stlink/issues/650

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1302,19 +1302,19 @@ static int map_file(mapped_file_t* mf, const char* path) {
     }
 
     if (fstat(fd, &st) == -1) {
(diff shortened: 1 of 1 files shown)


Nightwalker-87:
Thx for checking this. It appeared to be a minor issue then. :-)

chenguokai:
Not really, I suggest do not close https://github.com/stlink-org/stlink/issues/650 and close https://github.com/stlink-org/stlink/issues/879 only

Nightwalker-87:
It's ok like this, I think. The author there has not provided any further information that would allow to investigate. Also he no longer holds any fork of this project. I'd not assume a general problem here.

Issue #378, opened 2016-02-29
CLI tools need better argument parsing


smithbone:
The v1/v2 checking code for st-flash is dependent on the number of arguments provided.  <size> is listed as an option argument but if you you provide that argument then the argument parser assumes you are talking to a v1 device because 4 arguments are provided.

gerryqd:
Just checked the code, seems the <size> info is useless when it is in "write" mode.

Anyway, this is a bug.

xor-gate:
Did you fixed this issue with pull request #413 ? I did review it and seems to be fine. I'm currently not very happy how the option parsing is implemented and it needs some love and reworking. Thanks for looking into it.

gerryqd:
Sorry, I did not fix it in #413. Personally, I think the opt parse function should be rewritten. Some cases should be taken into account.

xor-gate:
Yeah you are right, I will rename this issue because it is major. I take two small option parsing libraries into account which are both MIT license:

* https://github.com/clibs/flag
* or https://github.com/clibs/commander

xor-gate:
Related to #208

Nightwalker-87:
@gerryqd: What is the current state of this?

Nightwalker-87:
@gerryqd: What is the current state of this?

Nightwalker-87:
@grevaillot: Can you confirm if this resolves with your recent PR?

grevaillot:
i'm a bit puzzled by that piece of code, the v1/v2 stlink handling is a bit broken, the device path is not actually used, and i think it's useless now.

still it would be nice to replace the whole custom opt parsing by something more nice / standard..

PR #918, opened 2020-04-10
Fixed st-info --probe mechanism

grevaillot:
Hi,

here's a small patchset to fixup st-info probe mecanism and lower layers.

 - handle stlink v1
 - fetch serial info for all stlinks
 - return proper number of functionnal and connectable stlinks
 - fix probe fail when one stlink v1 and more recent one are connected
 - add macros for usb pid filtering.

This has been tested with a couple of devices connected (v1, v2, v2.1 and v3). Note that it seems that all v1 have the same serial, so, i guess that selecting them will be complicated - there's apparently some kind of hidden env variable filter mecanism based on usb addr in addition to the --serial parameter, but i've not tested it.

> Found 3 stlink programmers
>  serial:     493f6c06483f53491743073f
>  hla-serial: "\x49\x3f\x6c\x06\x48\x3f\x53\x49\x17\x43\x07\x3f"
>  flash:      32768 (pagesize: 1024)
>  sram:       6144
>  chipid:     0x0445
>  descr:      F04x
>  serial:     303033413030343233313337353130413339333833353338
>  hla-serial: "\x30\x30\x33\x41\x30\x30\x34\x32\x33\x31\x33\x37\x35\x31\x30\x41\x33\x39\x33\x38\x33\x35\x33\x38"
>  flash:      524288 (pagesize: 4096)
>  sram:       98304
>  chipid:     0x0469
>  descr:      G4 Category-3
>  serial:     303030303030303030303031
>  hla-serial: "\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x30\x31"
>  flash:      131072 (pagesize: 1024)
>  sram:       8192
>  chipid:     0x0420
>  descr:      F1xx Value Line
>

PR #922, opened 2020-04-12
flash.c: stlink_open_usb is actually able to open v1 and v2 (and v3)

grevaillot:
flash.c: stlink_open_usb is actually able to open v1 and v2 (and v3 .…….)** devices.

no need to check that _not used_ o.devname ..
no need to specify devname in commandline

get rid of the devname cmdline stuff, and update comments

should probably help with #378 ..

### DIFF
--- a/src/tools/flash_opts.c
+++ b/src/tools/flash_opts.c
@@ -22,7 +22,6 @@ static int bad_arg(const char *arg) {
 }
 
 int flash_get_opts(struct flash_opts* o, int ac, char** av) {
-    bool serial_specified = false;
 
     // defaults
     memset(o, 0, sizeof(*o));
@@ -63,7 +62,6 @@ int flash_get_opts(struct flash_opts* o, int ac, char** av) {
                 memcpy(buffer, serial + j, 2);
                 o->serial[length - k] = (uint8_t)strtol(buffer, NULL, 16);
             }
-            serial_specified = true;
         }
         else if (strcmp(av[0], "--area") == 0 || starts_with(av[0], "--area=")) {
             const char * area;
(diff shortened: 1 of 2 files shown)


Nightwalker-87:
This is actually a result of #918, right?

grevaillot:
i think it was half-working before my previous PR - i'm pretty sure i managed to use a v1 without specifying any device path, at least this could work in specific conditions.

Nightwalker-87:
Yes, but what I meant was, that these commands in `flash.c` became obsolete with your previous PR because of your refactoring there - regardless of the state of functionality there was before.

Nightwalker-87:
@grevaillot: Can you update this PR to resolve the conflict in `flash.c`? Is there any specific ticket to be linked to this PR as seen from your side?

grevaillot:
pr conflict fixed - not much time to scan old tickest - This can probably close initial issue of #378 - v1/v2 selection issue. Still, switching to something "working" to parse cmdline opt would be nice, but we now avoid that bug.

src/tools/flash_opts.c
```diff
@@ -137,8 +135,8 @@ int flash_get_opts(struct flash_opts* o, int ac, char** av) {
         av++;
     }
 
-    // command and (optional) device name
-    while(ac >= 1) { // looks like for stlinkv1 the device name and command may be swaped - check both positions in all cases
+    // commandd
```
slyshykO, line 138: Is it possible to do something with this comment for improving its understandability?

grevaillot: hmm, looks like merge issue somehow, original comment should have been kept

Nightwalker-87: To me it looks like, as if this comment is obsolete now anyway.

Nightwalker-87:
> not much time to scan old tickets - This can probably close initial issue of #378 - v1/v2 selection issue. Still, switching to something "working" to parse cmdline opt would be nice, but we now avoid that bug.

Thx, not necessary for you to do so, I think we're fine to link #378 here. Any further more general improvements for opt parsing should be addressed by a new ticket, should such be desired.

Nightwalker-87:
Merci @grevaillot!

PR #925, opened 2020-04-13
st-flash: minor usage fix, and make cmdline parsing more user friendly

sderr:
Somewhat fix the help message that showed some parameters as mandatory,
when in reality they depend on the operation type and file format.

Try to show explicit error messages for the most common mistakes.

### DIFF
--- a/src/tools/flash_opts.c
+++ b/src/tools/flash_opts.c
@@ -11,6 +11,18 @@ static bool starts_with(const char * str, const char * prefix) {
     return (0 == strncmp(str, prefix, n));
 }
 
+static int invalid_args(const char *expected)
+{
+	fprintf(stderr, "*** Error: Expected args for this command: %s\n", expected);
+	return -1;
+}
+
+static int bad_arg(const char *arg)
+{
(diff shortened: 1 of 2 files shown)


src/tools/flash_opts.c
```diff
@@ -11,6 +11,18 @@ static bool starts_with(const char * str, const char * prefix) {
     return (0 == strncmp(str, prefix, n));
 }
 
+static int invalid_args(const char *expected)
+{
```
chenguokai, line 15: May you adjust the code style here and the one of `bad_arg`? It looks a little inconsistent considering the context.

sderr: Do you mean the { at the end of the line ?

chenguokai: Exactly

sderr: Hopefully fixed

Nightwalker-87: Thx :-)

Nightwalker-87:
... and take-off. Thx @sderr!

PR #927, opened 2020-04-16
overall option code rework

grevaillot:
hi,

Continuing my work on g4 fixes, i added support for g0/g4 chips option read/write, and spend a bit of time on option code cleanup/rework.

basically, I moved some device specific code out of st-flash util, aligned api regarding option/read write device specific code and added device specific parameter to chipid database, allowing some sanity checks before writing and some generic option read code. all devices are now using the same code to flash option, not depending on st-flash parameters.

 - write --area-option "option integer in hex"
 - write option.bin "option address" (not documented afaik)
 - read --area-option

note that current option write code mostly only support writing the first 32bits of option bytes, nor reading more, so this need some more reviews / rework to be safe, but i think this is not less safe than previously - currently option size is hardcoded to 32bits for concerned/not tested chips.

I can maybe split pr, this went a bit wild while attempting to add my stuff..


grevaillot:
derp, merge issue, lemme see..

grevaillot:
pushed force, better, not sure what happened

Nightwalker-87:
Great! Give me some time to review. I don't mind though if you add further commits in the meanwhile, if you feel like it. To me PRs don't necessarily need to stay below a certain size as long as they focus on one topic. Keep up the good work! :+1:

src/common.c
```diff
@@ -2736,41 +2738,18 @@ int stlink_fwrite_flash(stlink_t *sl, const char* path, stm32_addr_t addr) {
...
 
-    // Make sure we've loaded the context with the chip details
-    stlink_core_id(sl);
-
-    /* Check if chip is supported and for correct address */
-    if (sl->chip_id != STLINK_CHIPID_STM32_G0_CAT1 &&
-       sl->chip_id != STLINK_CHIPID_STM32_G0_CAT2) {
-        ELOG("Option bytes writing is currently only supported for the STM32G0\n");
+    if (unlock_flash_if(sl)) {
+        ELOG("Flash unlock failed! System reset required to be able to unlock it again!\n");
```
Nightwalker-87, line 2749: Good! :+1: We have seen this in issues before where people did not understand what happened ...

src/common.c
```diff
@@ -3220,30 +3274,42 @@ int stlink_read_option_bytes_f4(stlink_t *sl, uint32_t* option_byte) {
...
+			return stlink_write_option_bytes_f2(sl, base, addr, len);
+		case STLINK_CHIPID_STM32_F446:
+			return stlink_write_option_bytes_f4(sl, base, addr, len);
+		case STLINK_CHIPID_STM32_L0_CAT2:
+			return stlink_write_option_bytes_l0_cat2(sl, base, addr, len);
+		case STLINK_CHIPID_STM32_L496X:
+			return stlink_write_option_bytes_l496x(sl, base, addr, len);
+		case STLINK_CHIPID_STM32_L152_RE:
+		case STLINK_CHIPID_STM32_L1_HIGH:
         return stlink_write_option_bytes_l1(sl, base, addr, len);
```
Nightwalker-87, line 3305: Indentation?

src/common.c
```diff
@@ -3220,30 +3274,42 @@ int stlink_read_option_bytes_f4(stlink_t *sl, uint32_t* option_byte) {
...
-    }
-    else if ((sl->chip_id == STLINK_CHIPID_STM32_L0_CAT2) && (addr == STM32_L0_CAT2_OPTION_BYTES_BASE)) {
-        return stlink_write_option_bytes_l0_cat2(sl, base, len);
-    }
-    else if ((sl->chip_id == STLINK_CHIPID_STM32_L496X) && (addr == STM32_L496X_OPTION_BYTES_BASE)) {
-        return stlink_write_option_bytes_l496x(sl, base, len);
-    }
-    else if (((sl->chip_id == STLINK_CHIPID_STM32_L152_RE) || (sl->chip_id == STLINK_CHIPID_STM32_L1_HIGH) ) &&
-            ((addr == STM32_L1_OPTION_BYTES_BASE) || (addr == STM32_L1_OPTION_BYTES_BASE+4))) {
+	switch (sl->chip_id) {
```
Nightwalker-87, line 3294: Good concept to implement this. I like that as it is much easier to follow and understand.

grevaillot:
i've got a bigger patch to come with more refactoring, but this will need some testing and a bit of time .. that one can be merged i think

src/common.c
Nightwalker-87: Thx for the fix.

Nightwalker-87:
Thank you!

PR #929, opened 2020-04-16
Build settings

Nightwalker-87:
- Added new build configs for Travis CI
- Refactoring for cmake modules & settings
- Updated Travis build script

Should see some testing if compilation also works well unter mac and win in real-world conditions. So far I did some runs on a debian bullseye system. GUI works well. Focus should be on correct package creation when reviewing and anything else that catches your eye of course.

The 3rd test (flash.c) still fails, as I haven't found the root of the problem yet.

Further I plan some more changes to the project structure, but I am waiting with this until several other issues have been addressed towards our upcoming release. Otherwise we would just run into unnecessary merge conflicts.


Nightwalker-87:
This may be an undiscovered older problem...

We can make gtk3 an requirement for installations via brew and Macports which then would allow mac users to make use of this feature as well. As far as I can see there is no mention related to the GUI for macOS in our project documentation at all. To me it is not necessary to provide an additional package for the gui on linux either. So my suggestion on this is to include the gui into the main package and to install it by default if gtk can be retrieved automatically from a repo and skip it if this is not possible (e.g. on stand alone installations).

Nightwalker-87:
> CMake Error at CMakeLists.txt:153 (set_target_properties):
>   set_target_properties called with incorrect number of arguments.

This should not fail really. I think it is a problem with reading the project version number when git is not present. That would also explain why this only happens when downloading the zip archive.

Nightwalker-87:
@chenguokai: I found that some necessary cases were not addressed in the workflow of the `version.cmake` module. Please retry.

chenguokai:
Build succeeded  without git, nice work!

But I still cannot figure out why cmake failed to recognize either `gtk2` or `gtk3` which has been installed by MacPorts. Confirmed that both headers and binary dynamic libs are present.

Nightwalker-87:
I'm on that topic currently.

Nightwalker-87:
One reason for that seems to be in `/CMakeLists.txt`:

```
if (NOT APPLE AND NOT WIN32 AND NOT CMAKE_CROSSCOMPILING)
    find_package(PkgConfig)
    pkg_check_modules(gtk gtk+-3.0)
endif ()
```
Thus macOS is always excluded from GUI builds and the detection  in `/src/stlink-gui/CMakeLists.txt` does not help as  `gtk_FOUND` has not seen any result from the previous check:

```
if (NOT gtk_FOUND)
    message(STATUS "gtk not found!")
    return()                         # no GTK present => no GUI build
endif ()
```

Nightwalker-87:
@chenguokai Does the GUI code compile on macOS now?

Nightwalker-87:
@chenguokai: Here are some more improvements: We query for GTK3 on all unix-based systems. If cmake does not find it, there will be no compilation for the GUI. I've updated the related documentation as well, so have a short look at that first. Hence I expect the GUI to compile flawlessly on macOS now, if there is no mistake.
Please give it another try.

@grevaillot and @chenguokai you may finally review this now. I'm not planning any extra work on this PR apart from necessary fixes, if needed.

chenguokai:
Works so well!

Nightwalker-87:
@martonmiklos: Can you review this PR as well?

martonmiklos:
Managed to build it under 18.04 Ubuntu.

Issue #915, opened 2020-04-09
Rewrite of flashloader source files (BSD license)


Nightwalker-87:
Pick up of an issue mentioned before by @xor-gate in #682:

> we need to update LICENSE and add the GPL part also for those files. I had some discussion in an email with the original authors which seemed to be copied from the OpenOCD project into stlink project. So we have a mix of GPL van BSD licensed code. If something is unclear please post a comment.

Nightwalker-87:
I'm going to look at this now.
Generally speaking I would not add such additional notes on licensing to the LICENSE file because that may possibly break the correct license recognition and display in git-header bar of the project, (text does not match the template any more) - not definitely sure, but have seen something similar in another project before. Thus this should better go into our README, but somewhere in the header part, that it is easy to spot.

Nightwalker-87:
@xor-gate: Did you only mean the two flashloaders for which the source files have a header, or did you have in mind that (in addition to these) there were also others (without any header) copied from openocd, but without any reference? If the latter is the case, one would have to compare line by line to identify. Regarding the former ones: These were indeed correctly declared - but somewhere at the very bottom of our README file. I'll move this important note to the top.

Nightwalker-87:
ping @xor-gate

chenguokai:
Maybe we can just clean room the source code to make a pure BSD one.
I checked to confirm that these assemblies are small enough that a clean room can be finished in a short time. May need a volunteer  who has not read the GPL ones to do the actual coding part.

Nightwalker-87:
Good idea. Indeed it would be nice to have an independent solution for this project.

chenguokai:
Volunteer found.

I will take part in this process.
Before that, I wonder which file(s) will need a  clean room.
The one for F0 and lx seems to be more complex and may be not from the same source. Should all the assemblies be rewritten? A complete list of files that need a clean room would be appreciated.

Nightwalker-87:
Well at first I'd suggest to review what is currently there and whether there are any redundant files or sections. Here are the related references in git history, when the loader files appeared:

8151bf5c4a443608d4fec15cf7dfe79a02cdd096 F2/F4
e8b15c0e787c7d5fc90c35d53165fcf36afb39e3 F0
94c7ac86dcb2a18ab4b2b60309892f6102a53205 Lx
44c645b7d74b7669b0e3988155528c2edcfbb05e L0x
71535104e47f2a96ce6a87a6da0fc4519bc64d15 F4lv
[...]

PR #932, opened 2020-04-19
[refactoring] BSD-License-compliant flashloader rewrite

chenguokai:
This will finally close https://github.com/stlink-org/stlink/issues/915


Nightwalker-87:
@chenguokai: Can you rebase this PR-branch back to the current development commit 05cf37570dcb0a7791d64ba0be929e1ab48e5a6f, which has a clean build to prevent any corruption?
We've messed something up due to rushing around after several bug reports. I'm sorry for that.
I'll let you know as soon as we can proceed any further on `develop`.

chenguokai:
Hope this rebase will help

chenguokai:
My local test shows that the current code is known to be working on stm32f103 and stm32f401. I am dealing with some minor formatting issues and a critical improvement about the compile process. After that the code will be ready for test.

Nightwalker-87:
@chenguokai: I assume it will take some time to review this. However my first impression is positive and it appears we will see a valuable contribution to the project again. :+1:
Overlooking this quickly let me take note of that it may be good to have comments on flash-words, loader codes and memory addresses again to help users and developers along with more detailed explanations and descriptions. We can add a Chinese version in the markdown document, but I personally believe it may not be necessary to have it, as I assume that an overwhelmingly large part of our chinese users is capable of understanding the English version very well. Also we may change the naming of the markdown document and title of this PR later on, but I haven't made up my mind on that yet. As I see you are also still actively working on this topic. I just wanted to give you a feedback of my impression beforehand.

chenguokai:
The Chinese version is the initial version used in cleanroom process so I think it is needed for documentary purpose. Indeed the English version is more valuable for most users/developers. I will adjust the order and put the English version at the front.
The comments on flash_words and loader codes need more detailed explanations. Maybe a more detailed version should be placed in other places rather than the cleanroom doc. I think `doc/developers.txt` can be an option.

Nightwalker-87:
I think we should have more people looking at this as it is a key component.
What exactly do you mean with "cleanroom" - why did you choose this naming?
I don't think that explanations should go to `/doc/developer.txt` as this is only a quick and dirty copy-paste-doc created "once upon a time". I actually feel we should have a separate document `/doc/flashloaders.md` for it. Explanations which were present as comments in the old flashloader files could go into this document, maybe in form of a parameter | explanation table format - something like this. Maybe someone even has a better idea for that. In the end users should find a place to gain knowledge on how this works.

chenguokai:
> What exactly do you mean with "cleanroom" - why did you choose this naming?

Because it is what kind of method this PR have followed.

From https://opensource.stackexchange.com/questions/6216/can-i-cleanroom-code-by-myself-if-public-specifications-already-exist

> A clean room design is a preventative defence strategy in anticipation of copyright infringement claims. The idea is that a new creative work cannot infringe on the copyright of an earlier work if the new work's authors did not view the earlier work. A clean room design process implements this isolation by having different people handle reverse-engineering of the earlier work and creating the new work, with the two groups only communicating via a spec that can be more easily reviewed for non-infringement.

> I don't think that explanations should go to /doc/developer.txt as this is only a quick and dirty copy-paste-doc created "once upon a time". I actually feel we should have a separate document /doc/flashloaders.md for it. Explanations which were present as comments in the old flashloader files could go into this document, maybe in form of a parameter | explanation table format - something like this.

That's what I wanted, too. This clean room document provide only essential information for another one to implement a BSD code. What I wanted to add to any new document (either /doc/flashloaders.md or some file with another proper name) will and should include a more detailed explanation of why we should follow those restrictions.

Nightwalker-87:
Ah I understand. To be honest have never come across this topic before, thus I could not associate the name. Ok so this reads as being a somehow intermediate step towards a complete solution of this issue. I'm fine with that. Thx for explaining. :+1:
It would be desirable though to merge a complete solution if possible.

chenguokai:
Hope the latest commit will help

src/flash_loader.c
```diff
@@ -47,157 +55,134 @@ static const uint8_t loader_code_stm32vl[] = {
...
-        0xfb, 0xd1,             //      bne.n   e <wait>
-        0x00, 0xf1, 0x01, 0x00, //      add     r0, r0, #1
-        0x01, 0xf1, 0x01, 0x01, //      add     r1, r1, #1
-        0xa2, 0xf1, 0x01, 0x02, //      sub     r2, r2, #1
-        0xef, 0xe7,             //      b       next
-                                // done:
-        0x00, 0xbe,             //      bkpt
-        0x00, 0xbf,             //      nop
-                                // flash_base:
-        0x00, 0x3c, 0x02, 0x40  //      .word   0x40023c00
```
Nightwalker-87, line 200: We should have some comments like these for the new assembly code as well I think

flashloaders/stm32f0.s
```diff
@@ -1,32 +1,63 @@
...
-    bic     r3, r5
-    str     r3, [r4, #16]     /*  FLASH->CR &= ~FLASH_CR_PG  */
-    bkpt    #0x00
-
-STM32_FLASH_BASE: .word 0x40022000
+    .syntax unified
+    .text
+
+    .global mycopy
+mycopy:
```
Nightwalker-87, line 5: Unless there is no specific meaning for "my", we may just simply leave out this prefix everywhere. I don't think there can be any claims on that wording. :laughing:

doc/flashloaders.md
```diff
@@ -0,0 +1,54 @@
...
+
+If you find some flashloaders to be broken or you need to write a new flashloader for new devices, the following tricks may help.
+
+1. Modify `WAIT_ROUNDS` marco to a bigger value so that you will have time to kill st-flash when it is waiting for a halted core.
+2. run `st-flash` and kill it after the flashloader is loaded to SRAM
+3. launch `st-util` and `gdb`/`lldb`
+4. set a breakpoint at the base address of SRAM
+5. jump to the base address and start your debug
+
+The tricks work because by this means, most work (flash unlock, flash erase, load flashloader to SRAM) would have been done automatically, saving time to construct a debug environment.
```
Nightwalker-87, line 54: This is a well written documentation. :+1:

Nightwalker-87:
If this whole stuff works (what I can't verify by myself here) we can go with it I think, apart from some minor changes. We should have @grevaillot reviewing it as well, as I have the impression he may be more familiar with some details.

src/flash_loader.c
chenguokai: I doubt the necessity actually.
The comments should have shown people where they can find the assembly form of this flashloader binary. Maintaining another copy of the assembly is redunctant in my view.

An improvement here might be adding a note like:

**DO NOT EDIT BINARIES DIRECTLY, EDIT ASSEMBLIES INSTEAD**

flashloaders/stm32f0.s
chenguokai: Indeed, I will replace the name with no more "my".
The binaries will surely stay the same so no need to update `flashloaders.c`

src/flash_loader.c
Nightwalker-87: Yes, a note in the header section of this sourcecode file seems a good idea, but I would write "DO NOT MODIFY SOURCECODE DIRECTLY, EDIT ASSEMBLY FILES INSTEAD" (as binaries are commonly referred to as a compiled source code files). Correct me, shall I be wrong here.
However I just noticed that I placed my comment on the wrong file :expressionless:. This should have addressed the assembly files (.s).

chenguokai: The comment in assembly files seems a bit hard to deal with. Since it is undertaken using a clean room method, I cannot add those comments to the new version myself. The code, in my view, is self-explaining enough. When I was doing the clean room research, the only odd design I encountered is about the word count (passed  in `r2`), which is not always WORD count but sometimes double word / byte count.

chenguokai:
All "my"s have been replaced.
Added a note at the head of `flash_loader.c`

Nightwalker-87:
ping @grevaillot

src/flash_loader.c
Nightwalker-87: I may pay attention to this later on.

Nightwalker-87:
I'm tired of waiting any longer for a second review. :-1:
If nobody cares, I don't do either...

chenguokai:
Hope that untested devices work well🤔

Nightwalker-87:
It's ok and if so, to me it's not in your responsibility.

Issue #432, opened 2016-06-15
[feature] Programmer serial should be displayed when no target is connected


xor-gate:
When there is currently no target connected the programmer is found but no serial is displayed. This is not handy when we extract the serial number (and OpenOCD `hla_serial`) when the target is powered on.

Nightwalker-87:
@grevaillot: You just won this ticket, because I have found that your commit 804c38ead8aef3e4a1640a82a9b9c01f4f60eed1 somehow seems to address the issue above (per accident?) I find my STlink-v2 programmer to behave differently since (exactly) this commit. I don't know really what causes this change.

Here is the output of `st-info --probe` if no device is attached to the programmer:

Before: 
```
Found 1 stlink programmers
```

Afterwards:
```
Found 1 stlink programmers
 serial:     3f76050132124647524b4e00
 hla-serial: "\x3f\x76\x05\x01\x32\x12\x46\x47\x52\x4b\x4e\x00"
 flash:      0 (pagesize: 0)
 sram:       0
 chipid:     0x0000
 descr:      unknown device
```

I don't really understand why this is the case.
However, as you are the author, it just makes sense to let you have a look at this.

grevaillot:
I was not aware of that ticket, but yes, this is intended. I've got a couple of fixes for st-flash and gdb server to push to usb probe returning some st-link info with no target connected, i'll tag them with that ticket number.

grevaillot:
hmm, the commit you referenced is not the good one, it was fixed in the probe rework patchset

Nightwalker-87:
@grevaillot I don't really know what causes the observed change there then. Maybe it is only a partial fix that leads to the change with this device or only a few others as seen from the code side, but is not a complete fix that addresses all devices by then - no clue on that yet. I'll try my second programmer with this commit as well (before and after). They are both CKS32F103C8T6 devices.

Nightwalker-87:
Here is the output of `st-info --probe` with commit 804c38ead8aef3e4a1640a82a9b9c01f4f60eed1 for the second programmer, which appears to be the same result:

```
Found 1 stlink programmers
 serial:     3f70050132124647524b4e00
 hla-serial: "\x3f\x70\x05\x01\x32\x12\x46\x47\x52\x4b\x4e\x00"
 flash:      0 (pagesize: 0)
 sram:       0
 chipid:     0x0000
 descr:      unknown device
```

[...]

PR #933, opened 2020-04-19
handle probed stlink with unconnected target properly

grevaillot:
stlink_open_usb now returns without failure when no target is detected, or when target is not fully handled. Fix behaviour of tools in that case.

nb: st-util handling should maybe be discussed. currently it allows to start with no supported flash, which is what i would expected from a debugging tool, but flashing wont work obvs.

refs #432


Nightwalker-87:
@grevaillot: Thx for addressing this.

PR #935, opened 2020-04-19
Add CMAKEFLAGS and install target

### LINKED ISSUE #804: [install] MAKEFLAGS not propagated from toplevel Makefile to CMake generated makefiles
In the toplevel Makefile, the varialbe MAKEFLAGS gets assigned but is not used. Only CMAKEFLAGS is used but not assigned somewhere. Seems to be a minor thing but I'd like to add -DCMAKE_INSTALL_PREFIX="${HOME}/.local" to this variable

I also added an install target
install: build/Release
	@$(MAKE) -C build/Release install

Now I can simply type make install

Jochen0x90h:
Fix #804

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 ##
 # This Makefile is used to drive building of Debug and Release targets of CMake
 ##
-MAKEFLAGS += -s
+
+# additional flags for cmake, e.g. install path -DCMAKE_INSTALL_PREFIX=$(HOME)/.local
+CMAKEFLAGS +=
 
 all: release
 ci: debug release binary test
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,6 +34,10 @@ binary: build/Binary
 		@echo "[BINARY]"
 		@$(MAKE) -C build/Binary
 
+install: build/Release
+		@echo "[INSTALL] Release"
+		@$(MAKE) -C build/Release install
+
```
Nightwalker-87, line 40: Please add install to the help-section as well.

Nightwalker-87: Thx for fixing this. However one should avoid force pushing, if not strictly necessary. This should only be done if false changes have to be overridden in order to proceed. In this case an additional patch-commit would have done the job as well.

Nightwalker-87:
Thanks for contributing.

Issue #937, opened 2020-04-20
Error building from develop branch, cpuid problems


Vascom:
I try build commit c7874f805c63c60285f2b190f6589da8ae5e47fa
And it failed with this errors:
```
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c: In function 'main':
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c:34:9: error: 'cpuid.revision' may be used uninitialized in this func
tion [-Werror=maybe-uninitialized]
   34 |         printf("cpuid:part = %#x, rev = %#x\n", cpuid.part, cpuid.revision);
      |         ^
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c:31:27: note: 'cpuid.revision' was declared here
   31 |         cortex_m3_cpuid_t cpuid;
      |                           ^
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c:34:9: error: 'cpuid.part' may be used uninitialized in this function [-Werror=maybe-uninitialized]
   34 |         printf("cpuid:part = %#x, rev = %#x\n", cpuid.part, cpuid.revision);
      |         ^
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c:31:27: note: 'cpuid.part' was declared here
   31 |         cortex_m3_cpuid_t cpuid;
      |                           ^
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c:33:9: error: 'cpuid.variant' may be used uninitialized in this function [-Werror=maybe-uninitialized]
   33 |         printf("cpuid:impl_id = %0#x, variant = %#x\n", cpuid.implementer_id, cpuid.variant);
      |         ^
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c:31:27: note: 'cpuid.variant' was declared here
   31 |         cortex_m3_cpuid_t cpuid;
      |                           ^
/builddir/build/BUILD/stlink-org-stlink-c7874f8/tests/usb.c:33:9: error: 'cpuid.implementer_id' may be used uninitialized in this function [-Werror=maybe-uninitialized]
   33 |         printf("cpuid:impl_id = %0#x, variant = %#x\n", cpuid.implementer_id, cpuid.variant);
      |         ^
[...]

PR #938, opened 2020-04-20
Fix uninitialized cpuid

chenguokai:
Fixes https://github.com/stlink-org/stlink/issues/937


Nightwalker-87:
:+1:

Nightwalker-87:
Thanks for fixing this.

PR #939, opened 2020-04-20
Fix pkg-config libdir path

Vascom:
Path for *.pc files should be .../pkgconfig/ instead of .../pkg-config/


Vascom:
I need to do something?

Nightwalker-87:
Obsolete: This configuration is no longer used since PR #931 and will be removed soon.

Nightwalker-87:
> I need to do something?

No, you are fine, it's just that the `development` branch succeeded this PR in the meanwhile, so it is no longer needed.

PR #941, opened 2020-04-20
[refactoring] Cleanup for option bytes and flash settings

grevaillot:
hi,

Some more work on flash handling, mostly around implementing and reflactoring register access and flash locking helpers. option write/read device specific has been refactored based on the main flash familly, and f0 support primitives have been added.

a second pr will follow later to support then.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -24,8 +24,6 @@
 #define __attribute__(x)
 #endif
 
-/* todo: stm32l15xxx flash memory, pm0062 manual */
-
 /* stm32f FPEC flash controller interface, pm0063 manual */
 // TODO - all of this needs to be abstracted out....
 // STM32F05x is identical, based on RM0091 (DM00031936, Doc ID 018940 Rev 2, August 2012)
@@ -213,13 +211,14 @@
 
 #define STM32L4_FLASH_OPTR_DUALBANK     21
 
-//STM32L0x flash register base and offsets
-//same as 32L1 above
-// RM0090 - DM00031020.pdf
+//STM32L0x flash register base and offsets RM0090 - DM00031020.pdf
 #define STM32L0_FLASH_REGS_ADDR ((uint32_t)0x40022000)
-#define STM32L0_FLASH_PELOCK_BIT (1u << 0)
-#define STM32L0_FLASH_OPTLOCK_BIT (1u << 2)
-#define STM32L0_FLASH_OBL_LAUNCH_BIT (1u << 18)
+#define STM32L1_FLASH_REGS_ADDR ((uint32_t)0x40023c00)
+
+#define STM32L0_FLASH_PELOCK (0)
+#define STM32L0_FLASH_OPTLOCK (2)
+#define STM32L0_FLASH_OBL_LAUNCH (18)
+
 #define FLASH_ACR_OFF     ((uint32_t) 0x00)
 #define FLASH_PECR_OFF    ((uint32_t) 0x04)
 #define FLASH_PDKEYR_OFF  ((uint32_t) 0x08)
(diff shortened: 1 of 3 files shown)


grevaillot:
ok, so, this was supposed to be a draft pr but i validated too fast... I've got a couple of fixed to push but some review would be welcome

src/common.c
```diff
@@ -633,7 +846,7 @@ static inline uint32_t read_flash_sr(stlink_t *sl) {
     else if (sl->flash_type == STLINK_FLASH_TYPE_WB)
         sr_reg = STM32WB_FLASH_SR;
     else
-        sr_reg = FLASH_SR;
+        return -1;
```
slyshykO, line 849: If possible add some error message before exit.

src/common.c
```diff
@@ -2788,152 +3003,65 @@ static int stlink_write_option_bytes_gx(stlink_t *sl, uint8_t* base, stm32_addr_
...
  * @param sl
  * @param addr of the memory mapped option bytes
  * @param base option bytes to write
  * @return 0 on success, -ve on failure.
  */
-static int stlink_write_option_bytes_l1(stlink_t *sl, uint8_t* base, stm32_addr_t addr, uint32_t len) {
-
+static int stlink_write_option_bytes_l0(stlink_t *sl, uint8_t* base, stm32_addr_t addr, uint32_t len)
+{
+	uint32_t flash_base = get_stm32l0_flash_base(sl);
```
slyshykO, line 3022: the indentation, if possible, should be the same as the next line.

Nightwalker-87: Please use 4 spaces for indentation instead of tab spacing. (I know this is already inconsistent, but will be aligned throughout the codebase midterm.) Also there will be some code style guidelines soon...

src/common.c
grevaillot: I'd like to put some asserts on that kind of unexpected "not implemented" codepaths - ending here with a not supported chip will lead to bad thing and need an a abort. But yes. I'll do that.

src/common.c
grevaillot: Oops

Nightwalker-87: No worries, it would just be convenient to save some upcoming work where possible.

slyshykO:
@grevaillot I think, use an **assert()** call in library code not right.

We have closed issue about it. #634

grevaillot:
updated patchset with fixes to above comments

> @grevaillot I think, use an **assert()** call in library code not right.
> 
> We have closed issue about it. #634

hmm, i might have looked too fast then, i thought release builds where defining NDEBUG - discarding assert statements ? i'll check.

slyshykO:
> hmm, i might have looked too fast then, i thought release builds where defining NDEBUG - discarding assert statements

Agree.
But take in consideration that we directly apply NDEBUG only for msvc toolset.

Nightwalker-87:
I'd like to suggest to do feature- and code testing before opening a PR in future, because once opened this keeps updating the travis build system with loads of step by step build tests which actually should be done locally by running ./travis.sh The travis CI build history would be much easier to follow, if only builds would be listed that are about to merge or have merged. This also applies to drafts by the way. Please consider this for ongoing work.

grevaillot:
ah. sorry for the ci stuff. code and feature is pretty much done except the assert stuff.

> Agree.
> But take in consideration that we directly apply NDEBUG only for msvc toolset.

Release builds on linux too apparently, I could not find where this was setup, but build is run with -DNDEBUG. I'd be happy to understand why :)

slyshykO:
> I could not find where this was setup, but build is run with -DNDEBUG.

Usually CMAKE_<LANG>_FLAGS_RELEASE cmake variable contain -DNBEBUG option, but it may be changed by command-line options or by another way. This is not mandatory that cmake by default should provide NDEBUG. So, my opinion is or we will set NDEBUD directly in our cmake project, or we shouldn't use asserts in library code.

Nightwalker-87:
@grevaillot: It's ok. No problem yet. BTW: There is also the possibility to discuss within an issue with references to a personal active dev-branch. This should be the same user experience but could be a solution.

grevaillot:
okay, i removed asserts and fixed a regression, lets review merge these bits now if it's ok with you.

Nightwalker-87:
Thanks for your contribution!

Issue #893, opened 2020-03-24
[feature] Adjust the JTAG/SWD frequency via cmdline option


Nightwalker-87:
As far as I can see, we are lacking a HowTo for the feature above, added by @jkhax0r in #534 
Who can contribute and help with this?

Nightwalker-87:
Taking a closer look at this reveals that the related functionality is only hardcoded without the ability to pass the JTAG/SWD frequency as an option parameter to `st-util`, what I actually expected to find. Thus there is also no documentation that can be added until this feature becomes usable from the command line. In the current state seems useless to me. Against this background we yet have another topic to be dealt with... Volunteers?

chenguokai:
I am working on it.

PR #953, opened 2020-05-07
[feature] Adjust the JTAG/SWD frequency via cmdline option

chenguokai:
Fix https://github.com/stlink-org/stlink/issues/893
* Add --freq argument for st-flash

### DIFF
--- a/include/stlink/tools/flash.h
+++ b/include/stlink/tools/flash.h
@@ -26,9 +26,10 @@ struct flash_opts
     uint32_t val;
     size_t flash_size;	/* --flash=n[k][m] */
     int opt;
+    int freq;
 };
 
-#define FLASH_OPTS_INITIALIZER {0, { 0 }, NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+#define FLASH_OPTS_INITIALIZER {0, { 0 }, NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
 
 int flash_get_opts(struct flash_opts* o, int ac, char** av);
 
--- a/include/stlink/tools/flash.h
+++ b/include/stlink/tools/flash.h
@@ -25,8 +25,8 @@ struct flash_opts
     enum flash_area area;
     uint32_t val;
     size_t flash_size;	/* --flash=n[k][m] */
-    int opt;
-    int freq;
+    int opt;            /* enable empty tail data drop optimization */
+    int freq;           /* --freq=n[k][m] frequency of JTAG/SWD */
 };
 
 #define FLASH_OPTS_INITIALIZER {0, { 0 }, NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
--- a/src/tools/flash.c
+++ b/src/tools/flash.c
@@ -28,9 +28,9 @@ static void cleanup(int signum) {
 
 static void usage(void)
 {
-    puts("command line:   ./st-flash [--debug] [--reset] [--opt] [--serial <serial>] [--format <format>] [--flash=<fsize>] {read|write} <path> [addr] [size]");
-    puts("command line:   ./st-flash [--debug] [--serial <serial>] erase");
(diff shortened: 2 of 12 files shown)


chenguokai:
I found one missing problem in the test. I am fixing.

include/stlink/tools/flash.h
```diff
@@ -26,9 +26,10 @@ struct flash_opts
     uint32_t val;
     size_t flash_size;	/* --flash=n[k][m] */
     int opt;
+    int freq;
```
Nightwalker-87, line 29: I think an inline comment would be useful here  (equal to `stlink.h`).

src/tools/flash.c
```diff
@@ -28,9 +28,9 @@ static void cleanup(int signum) {
 
 static void usage(void)
 {
-    puts("command line:   ./st-flash [--debug] [--reset] [--opt] [--serial <serial>] [--format <format>] [--flash=<fsize>] {read|write} <path> [addr] [size]");
-    puts("command line:   ./st-flash [--debug] [--serial <serial>] erase");
-    puts("command line:   ./st-flash [--debug] [--serial <serial>] reset");
+    puts("command line:   ./st-flash [--debug] [--reset] [--opt] [--serial <serial>] [--format <format>] [--flash=<fsize>] [--freq=<Hz>] {read|write} <path> [addr] [size]");
+    puts("command line:   ./st-flash [--debug] [--freq=<Hz>] [--serial <serial>] erase");
+    puts("command line:   ./st-flash [--debug] [--freq=<Hz>] [--serial <serial>] reset");
```
Nightwalker-87, line 33: To have the documentation part of the related issue addressed we should also consider an update in our tutorial as well. In my opinion the section "Useful tool options" would be a good place. I wanted to extend that section anyway... There we can also include a table for the available frequency options as well.

tests/flash.c
```diff
@@ -93,6 +95,51 @@ static struct Test tests[] = {
...
+        { .cmd = FLASH_CMD_WRITE,
+          .serial = { 0 },
+          .filename = "test.bin",
+          .addr = 0x80000000,
+          .size = 0,
+          .reset = 1,
+          .log_level = DEBUG_LOG_LEVEL,
+          .freq = 5,
+          .format = FLASH_FORMAT_BINARY }
+    },
```
Nightwalker-87, line 133: Duplicate?

chenguokai: The one is in another syntax (--freq=5k), the former one is "--freq 5k"

include/stlink/tools/flash.h
chenguokai: I will add a comment in the coming commit tomorrow

src/tools/flash.c
chenguokai: Indeed, I did some searching, no proper positions were found. Your idea sounds quite good👏

tests/flash.c
Nightwalker-87: Correct, I recognise the slight difference now.

src/tools/flash.c
Nightwalker-87: Feel free to add such info there at the top. I can do some additional formatting later on, when I continue my work on this tutorial. I may set up a table listing all available cmdline-options which also states for which tool they are available. That would avoid duplicates and provide a clearer appearance.

include/stlink/tools/flash.h
chenguokai: Added

src/tools/flash.c
chenguokai: Added, also for `--opt`

Nightwalker-87: Very attentive -  Thank you! :+1:

include/stlink/tools/flash.h
Nightwalker-87: Thanks!

Nightwalker-87:
:smiley: :+1:

PR #863, opened 2020-02-24
[feature] STLink-v3: Add stlink v3 support

Nightwalker-87:

Issue #820, opened 2019-07-25
[feature] STLINK/V3 programmer support


WRansohoff:
- [X] Programmer/board type: STlink/v3 (+on-board)
- [X] Programmer firmware version: V3.J2.M1 'STM32 Debug+Mass storage+VCP'
- [X] Operating system: Ubuntu 18.04
- [X] Stlink tools version: v1.5
- [X] Stlink commandline tool name: `st-util`, probably others as well.
- [X] Target chip (and optional board): STM32G431KB (STM32G431 'Nucleo-32')

Output:

```
[Time] WARN usb.c: Couldn't find any ST-Link/V2 devices
```

Expected/description:

The utility should connect to the device. I'm not sure how different these new ST-Link bridges are from the old ones, but hopefully things won't look too different. ST has a marketing blog post describing the new debuggers, but the technical information might be more spread out:

https://blog.st.com/stlink-v3set-in-circuit-debugger-programmer/

Sorry if this has already been added - I tried `st-util -s 3`, but I got: `stlink version 3 unknown!`

unclebanjoman:
I also have a serious problem. I just bought a brand new STLINK-V3SET programmer / debugger original from StMicro.

- Programmer type: STLINK-V3SET
- Operating system: Debian 9 (stretch)
- Stlink tools version: v1.5.1-31-g625f4cd
- Stlink commandline tool name: st-info and all others
- Target chip: STM32F746VG and/or STM32F469NI using original STLINK-V3SET programmer/debugger

The st-info utility discovers and manage my new STM32F746G-Discovery board without problems.
When I plug my new STLINK-V3SET programmer, executing the command:

./st-info --probe

I only obtain:

_"Found 0 stlink programmers._"

I've already put the appropriate file in /etc/udev/rules.d (filename: 49-stlinkv3.rules).
Issuing the lsusb -v command I get the following:

mmazza@linux9:~/ARMtoolchains/stlink-master/build/Release$ lsusb -v

Bus 005 Device 003: ID 0483:374f STMicroelectronics 
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass          239 Miscellaneous Device
  bDeviceSubClass         2 ?
  bDeviceProtocol         1 Interface Association
  bMaxPacketSize0        64
  idVendor           0x0483 STMicroelectronics
  idProduct          0x374f 
  bcdDevice            1.00
  iManufacturer           1 STMicroelectronics
[...]

PR #954, opened 2020-05-09
Basic compatibility for STLink-v3 programmer

Nightwalker-87:
Added support for JTAG command API v2:
--> compatibility with the STLink-v3 firmware which dropped support for the previous API v1
--> as of firmware version J11 the ST-LINK-V1 supports API v2 commands as well

As described by @martonmiklos in #863: "Apart from utilising the V2 command API there are some other minor differences in the v3 handling like extended version query, which generally seems to do it's job. At the moment most of the commands are working, however the flash loaded never hits a breakpoint after loading as I mentioned in #820."


src/usb.c
```diff
@@ -262,6 +262,30 @@ int _stlink_usb_write_debug32(stlink_t *sl, uint32_t addr, uint32_t data) {
...
 
+int _stlink_usb_get_rw_status(stlink_t *sl)
+{
+    if (sl->version.jtag_api == STLINK_JTAG_API_V1)
+        return -1;
+
+    unsigned char* const rdata = sl->q_buf;
+    struct stlink_libusb * const slu = sl->backend_data;
+    unsigned char* const cmd  = sl->c_buf;
+    int i, ret = 0;
```
chenguokai, line 273: This 'ret' caused a build failure on 32bit systems

Nightwalker-87: Thx. So can we go for an int16_t here then?

Nightwalker-87: The answer obviously seems to be "yes". :laughing:

Nightwalker-87:
This PR is going to be merged next as soon as it is finally approved.

chenguokai:
Mostly good, at least backward compatibility has been kept.
There is one warning at the end:
`libusb: warning [libusb_exit] application left some devices open`

Not sure if this warning is caused by this PR

Nightwalker-87:
I don't know really - if so `usb.c` would be of interest, but I haven't found anything there yet.

Nightwalker-87:
I don't know really - if so `usb.c` would be of interest, but I haven't found anything there yet.
@chenguokai does it happen on `develop` as well in your test case?

chenguokai:
> does it happen on develop as well in your test case?

No, at least with commit ID `4c1f5b9575a90178a9f8fd29b64ccbbb00953ec9`

Nightwalker-87:
> No, at least with commit ID `4c1f5b9575a90178a9f8fd29b64ccbbb00953ec9`

Unfortunately I can't find this commit ID.

Nightwalker-87:
I'll have a look.

Nightwalker-87:
I'll have a look. What did you do exactly? How can I try to reproduce?

chenguokai:
> How can I try to reproduce?

In this pull request, I noticed that the default log level was set to DEBUG, and the warning is printed as debug level. So a reproduction on earlier commits should be launched with `--debug`

Nightwalker-87:
Uh, I can't reproduce that. This is exactly what I tried here. Could it be that this is related to clang somehow? I use gcc on Debian. That is the only difference I see currently.

chenguokai:
Trying `gcc` on macOS only to get a compile error🤔

```
stlink/src/flash_loader.c:194:9: error: 'size' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  194 |  size_t size;
      |         ^~~~
cc1: all warnings being treated as errors

```

Nightwalker-87:
Ha, I got it now!

chenguokai:
I am a bit confusing. So under what circumstances will the warning appear? And what does the old install mean, a registration of stlink hardware from another driver?

Nightwalker-87:
> I am a bit confusing. So under what circumstances will the warning appear? And what does the old install mean, a registration of stlink hardware from another driver?

I had an old commit installed locally with `sudo make install` what I lost track of. So when I did rebuild a later version, the terminal commands were still linked to that installed version of the toolset, so I did not notice at first.

The warning seems to appear whenever there is a cmd sent via usb. I havent found the commit when it first appeared. It is somewhere in the `stlink-v3_pre` branch.

Nightwalker-87:
The problem appears in commit 0a91936b263f3fd09a42fb1ec32a72a06d9a39d6 for the first time.

src/usb.c
```diff
@@ -1025,50 +1025,44 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, bool reset, char serial[ST
...
+            continue;
 
-            if (sl->serial_size < 0)
-                continue;
+        if (memcmp(serial, &sl->serial, sl->serial_size) == 0)
+            break;
 
-            if (memcmp(serial, &sl->serial, sl->serial_size) == 0)
-                break;
+		libusb_close(handle);
```
Nightwalker-87, line 1039: @chenguokai: I think I found the reason - it's a bug:` libusb_close(handle);` is called at the wrong point in the routine. The order has changed here.

Nightwalker-87: In 09ea99aea51f741ce9c8c0e2ad755d40b3d86021 it was called _before_

`/* could not open device, continue */ if (ret) continue;`

while it is called much later in the version above. It does make a change. In a first try everything seemed to work as expected when placing this call before

`if ((serial == NULL) || (*serial == 0)) break;`,

but I'm not sure yet, if this _really_ is the correct place where it should go. We need to investigate carefully in order not to break anything else...

Nightwalker-87: Taking a closer look reveals that the command order in the function `stlink_t *stlink_open_usb(...)` is mixed up between:

`sl->serial_size = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, (unsigned char *)sl->serial, sizeof(sl->serial));`

and

`WLOG ("Couldn't find %s ST-Link devices\n", (devBus && devAddr) ? "matched":"any");`

holding duplicated lines presumably deriving from old and new versions. This section clearly needs to be cleaned up.

This is one of these typical problems that are likely to appear, if branches are not updated regularly with changes from the main development branch. One ends up with a few merges which introduce loads of changes at once and in the end troubles arise in figuring out what went wrong. Clearly not a good practise. :disappointed:

Nightwalker-87: This should now be fixed with def770b0c5ea841a69e7828a11baa4c306d2ce6a.

PR #961, opened 2020-05-19
Set static link for libssp (stack-smashing protection)

slyshykO:


Nightwalker-87:
Thanks for your efforts to find this small but important fix. :+1:

PR #963, opened 2020-05-22
[feature] Added connect under reset to stlink_open_usb( )

geoffreymbrown:
I've added a small change to stlink_open_usb to support connect under reset.  In addition, I added an option to tests/usb.c  (--reset or --no-reset) to use this mode.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "cmake.configureOnOpen": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "cmake.configureOnOpen": false
-}
--- a/include/stlink/tools/flash.h
+++ b/include/stlink/tools/flash.h
@@ -27,9 +27,10 @@ struct flash_opts
     size_t flash_size;	/* --flash=n[k][m] */
     int opt;            /* enable empty tail data drop optimization */
     int freq;           /* --freq=n[k][m] frequency of JTAG/SWD */
+    bool connect_under_reset;
 };
 
-#define FLASH_OPTS_INITIALIZER {0, { 0 }, NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+#define FLASH_OPTS_INITIALIZER {0, { 0 }, NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
 
 int flash_get_opts(struct flash_opts* o, int ac, char** av);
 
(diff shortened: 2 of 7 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "cmake.configureOnOpen": false
+}
```
Nightwalker-87, line 3: Note: We don't need this IDE-specific config in the codebase. Can you remove it?

geoffreymbrown:
Mea culpa.  There are times when vscode does too much.
Geoffrey

.vscode/settings.json
geoffreymbrown: Very sorry !   I've removed it.

Nightwalker-87: Thx.

chenguokai:
As far as I have seen, the code added a new argument to `tests/usb.c`, not to the existing utilities like `st-flash`, while the original issue refers to `st-flash`. Would you be kind enough to add this argument for `st-flash`?

Relevant files are `tools/flash_opts.c`, `tools/flash.c` and some relevant headers.

geoffreymbrown:
I can but I need a bit more guidance.   What flag should I add ?
--connect-under-reset ?

 Since the flash tool halts the processor in any case, the right behavior
is probably
to

1) connect under reset
2) force debug
3) release the reset
4) reread the chip info (under reset, the flash size isn't returned
properly)

Should I make those changes ?   I have limited ways to test them, but they
would be localized to that option flag.

Geoffrey

chenguokai:
@Nightwalker-87 Maybe you are the best person to make decisions on this kind of issues:-)

geoffreymbrown:
I'll do it in the next couple of days.   Probably should do the same for
st-info.

Nightwalker-87:
> I can but I need a bit more guidance. What flag should I add ? --connect-under-reset ?

I think that is what @chenguokai referred to. You can proceed as proposed, but we will need someone to test it before merging.

geoffreymbrown:
I've added the additional code to my branch -- the other pull request was
still in progress, so I'm not quite sure how to proceed.   I tested the
code with an stm32l476RG nucleo board executing the attached small program
which blinks a bit and then goes to standby mode with the debugger
interface disabled.  Without the --connect-under-reset flag both st-info
and st-flash fail.   With the flag they both execute successfully.

The code is (relatively) non-invasive.  Other than checking for the flag
and passing it to the usb open routine, it doesn't do anything.  I did have
to extend the usb open code a bit to force debug mode and release the reset
line in order to minimize the changes to st-info and st-flash

Nightwalker-87:
@geoffreymbrown: This PR has been updated automatically with the update to your branch. Thx for testing and the related description. Looks good to me. :+1:

Nightwalker-87:
@chenguokai: Can you review the updated changes as well?

Nightwalker-87:
Thx for your contribution @geoffreymbrown.

PR #970, opened 2020-05-31
doc/man: Fix installation directory

nmeum:
On Unix-like operating systems man pages are commonly installed to
`/usr/share/man` not `/usr/share/stlink/man`.

---

Also: I noticed the pkgconfig file is no longer installed by default, how do I fix that?

### DIFF
--- a/doc/man/CMakeLists.txt
+++ b/doc/man/CMakeLists.txt
@@ -30,7 +30,7 @@ foreach (manpage ${MANPAGES})
     endif ()
 
     if (f AND NOT WIN32)
-        install(FILES ${f} DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/man/man1)
+        install(FILES ${f} DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/man/man1)
         unset(f)
     endif ()
 endforeach ()


nmeum:
> Also: I noticed the pkgconfig file is no longer installed by default, how do I fix that?

https://github.com/stlink-org/stlink/blob/1ae86f089c7bea8cd8f1dd90f8323df9dad10406/CMakeLists.txt#L52

:thinking:

Nightwalker-87:
@nmeum: I currently don't see the need for it since these settings should be set by cmake's internal PkgConfig module. In fact this seems to be a leftover originally intended to be removed.

nmeum:
> I currently don't see the need for it since these settings should be set by cmake's internal PkgConfig module.

The need for installing a pkgconfig file? Why even provide one in the first place then?

Nightwalker-87:
I was used up to the previous release, before the cmake package config was rewritten (that makes it a leftover from my point of view).

nmeum:
> It was used up to the previous release, before the cmake package config was rewritten (that makes it a leftover from my point of view).

Well the cmake find_package package config thingy (`cmake-packages(7)`) and pkg-config are two entirely different things. The former can only be used with CMake, the latter also with other build systems, e.g. GNU make.

Nightwalker-87:
This thread should continue in the appropriate PR. Different topics should not be mixed.

Issue #800, opened 2019-05-07
[doc] Library usage


BigfootN:
Hello,

I would like to know where I can find (apart from reading the headers), if possible, any documentation on how to use the `stlink` library ?

Thank you very much in advance.

xor-gate:
Thats a good question there is a very very minimal example here https://github.com/texane/stlink/tree/6db0fc243ffb90aed314ef1d59811486e9214883/doc/app-example. The library is not (very well) documented, the library is used for lowlevel access on the stlink programmer and is not often used outside the project.

Nightwalker-87:
@BigfootN: Did this help you any further? I agree that we need more documentation on this.

Nightwalker-87:
@Ant-ON You recently mentioned that you use the library itself. What would you document on how to use it from your personal experience?

Nightwalker-87:
Closed due to inactivity and missing interest.

PR #971, opened 2020-05-31
CMakeLists.txt: Install pkgconfig files again

nmeum:
See https://github.com/stlink-org/stlink/pull/970#issuecomment-636539652

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ find_package(libusb REQUIRED)
 
 ## Package configuration (pkg-config) on unix-based systems
 if (NOT WIN32 AND NOT CMAKE_CROSSCOMPILING)
-    #add_subdirectory(cmake/pkgconfig)
+    add_subdirectory(cmake/pkgconfig)
     find_package(PkgConfig)
     pkg_check_modules(GTK3 gtk+-3.0)
 endif ()


Nightwalker-87:
@chenguokai: Any ideas on this?

chenguokai:
I am not familiar with pkgconfig. Neutral on this topic.

Vascom:
Need to move string
`set(STLINK_LIBRARY_PATH ${CMAKE_INSTALL_LIBDIR} CACHE PATH "Main library install directory")`
before 
```
## Package configuration (pkg-config) on unix-based systems
if (NOT WIN32 AND NOT CMAKE_CROSSCOMPILING)
    add_subdirectory(cmake/pkgconfig)
    find_package(PkgConfig)
    pkg_check_modules(GTK3 gtk+-3.0)
endif ()
```

Nightwalker-87:
> Well the cmake find_package package config thingy (`cmake-packages(7)`) and pkg-config are two entirely different things. The former can only be used with CMake, the latter also with other build systems (e.g. GNU make).

@nmeum: Does it even work with GNU make? The whole build procedure for this project is setup with cmake (this is why it is also required for compilation). The top-level makefile only refers to the cmake subroutine...

Nightwalker-87:
@Vascom: ok, but `find_package(PkgConfig)` should go somewhere else then if not related to `pkgconfig`.

Vascom:
> @Vascom: ok, but `find_package(PkgConfig)` should go somewhere else then if not related to `pkgconfig`. I'll move the cited block further to the end of the top-level CMakeLists.txt, to achieve this order without breaking with the reorganised cmake build settings again.

OK

nmeum:
> @nmeum: Does it even work with GNU make? The whole build procedure for this project is setup with cmake (this is why it is also required for compilation).

Are we are talking about the same thing? I mean stlink shoulid install the [stlink.pc file](https://github.com/stlink-org/stlink/blob/develop/cmake/pkgconfig/pkgconfig.pc.cmake) (as it did previously, i.e. in 1.6.0). Without my PR it doesn't install that file.  The`pkgconfig.pc` file is used by [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) it allows developers to find the correct compiler options when building applications depending on stlink, i.e. applications linking against libstlink.

Nightwalker-87:
@nmeum: This is the definition of the cmake module, which is also called: https://cmake.org/cmake/help/latest/module/FindPkgConfig.html As far as I read this, it calls the same application.
To me it would be important to know if:

1) both are in conflict
2) there ever was any integration of libstlink into external applications, because we don't have ANY documentation on that library itself within this project yet (apart from the very basic and tentative ` /doc/developer.txt`). I can remember a ticket request once that was never addressed due to missing interest.

If the second point is the case, we would need to look at way more things than pkg-config, if not I don't see the use case yet (for both versions actually) against the missing documentation, well knowing one approach was present before. Please let me know more about the current scenario to let us implement a complete and fully usable solution instead of puzzled pieces.

nmeum:
Ok, if libstlink is not intended to be used by applications anyhow than this can be closed :)

nmeum:
Ok, if libstlink is not intended to be used by applications anyhow than this can be closed :)

Nightwalker-87:
Related to #800.

Nightwalker-87:
I would propose to move the pkg-config settings to a config-textfile in /doc until a proper solution is implemented in form of a documented API. With this the current "puzzle-bit" remains available for developers who fancy to play with it on a pre-alpha-level, just as the `/doc/developer.txt` file and we don't need to worry about anything very rarely used and not currently supported.

Nightwalker-87:
@Vascom: Nevertheless I'll consider your feedback there. I hope this is also an acceptable solution for @nmeum so that we do not need to wipe this issue away. Also we should still make sure both mentioned approaches are compatible.

PR #972, opened 2020-06-01
Fix installation path for desktop-file and icons.

Vascom:
This change fixes installation path for desktop-file and icons.
They should be installed directly to
```
/usr/share/applications/
/usr/share/icons/hicolor/scalable/apps/
```

### DIFF
--- a/src/stlink-gui/CMakeLists.txt
+++ b/src/stlink-gui/CMakeLists.txt
@@ -13,11 +13,11 @@ if (NOT WIN32)
 
         # Install desktop application entry
         install(FILES stlink-gui.desktop
-                DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/applications)
+                DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/applications)
 
         # Install icons
         install(FILES icons/stlink-gui.svg
-                DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/icons/hicolor/scalable/apps)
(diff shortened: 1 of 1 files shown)


src/stlink-gui/CMakeLists.txt
```diff
@@ -13,11 +13,11 @@ if (NOT WIN32)
 
         # Install desktop application entry
         install(FILES stlink-gui.desktop
-                DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/applications)
+                DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/applications)
 
         # Install icons
         install(FILES icons/stlink-gui.svg
-                DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/icons/hicolor/scalable/apps)
+                DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/apps)
```
Nightwalker-87, line 20: I'd set the second path to `/usr/share/icons/stlink/...` to avoid a mixture of files with other applications within that subdirectory. Can you update that please?

Vascom: Why not mixture icons with other applications? All icons should be in one place so in desktop-file you can just write `stlink.svg` instead of full absolute path. It will work in any system.

Nightwalker-87: If this is the only way to achieve this - ok. I just want to make sure we don't run into any clutter when users need to locate files during a manual clean-up without any harm to other applications. That was the main point.

Vascom: It is best way.
```
ls /usr/share/icons/hicolor/scalable/apps/ | wc -l
83
```
on my system.

PR #978, opened 2020-06-04
[refactoring] General maintenance

### LINKED ISSUE #864: Uniformize coding style
Hello folks,

I have found a few code style inconsistencies (whitespaces at the end of the lines, tabs, etc.) during my last contribution.

I would propose to agree on the coding style and document it in a clang-formatter format, format the whole code with it and setup a github action (or integrate the clang-formatter to the Travis CI) to check the coding style on each commit.  I would also recommend to leverage coding style from existing projects to make it easier to define the it. 

I know it will make the git blame harder sometimes in the future, however it is always more pleasurable to work on code which is formatted and indented consistently.

I can help with the github action setup, or even creating the clang formatter file.

### LINKED ISSUE #976: Isn't $(prefix)/include/stm32.h too generic?
I'm not sure, is it bug or "as intended" but previous versions of stlink (1.6.0 and earlier) installed main `include/stlink.h` and several other headers under `include/stlink/`.
1.6.1 installs `include/stlink.h`, `include/stm32.h` and several other headers under `include/stlink/`.
Is it as intended? `stm32.h` looks too generic to be installed into system-wide include directory, isn't it?

Nightwalker-87:
- Reorganised project source directories
- Grouped source files for win32
- [refactoring] Build settings for win32
- Whitespace & code style cleanup
- Added check for cflag -std=gnu18

(Closes #864) (Closes #976)


src/st-util/gdb-server.c
```diff
@@ -30,6 +30,8 @@
 #include "gdb-server.h"
 #include "semihosting.h"
 
+#include "gdb-server.h"
```
Nightwalker-87, line 33: Just noticed this duplicated include. I'll fix it.

slyshykO:
I think we should merge getopt mmap mingw folders to win32

Nightwalker-87:
Good idea. Give me some time, I want to see the CI pass first. I'll let you now as soon as a review makes sense. I just had to fix a mistake.

Nightwalker-87:
Note: mmap is not only used with win32. It should go to `src/` instead.
Also we should not make use of the `unistd.h` file (it does not seem to have been included in recent versions anyway). I run into conflicting types if it is included. If it works without it, we should scrap it, I think. Can you verify this?

slyshykO:
unistd.h is used to build with msvc. So I propose to move it to win32/unistd/unistd.h

slyshykO:
```cmake
if (WIN32 OR MINGW OR MSYS OR MSVC)
    include_directories(src/win32)
    set(STLINK_SOURCE "${STLINK_SOURCE};src/mmap.c;src/win32/mingw.c")
    set(STLINK_HEADERS "${STLINK_HEADERS};src/mmap.h;src/win32/mingw.h")
    if (MSVC)
        include_directories(src/win32/getopt)
        # Use string.h rather than strings.h and disable annoying warnings
        add_definitions(-DHAVE_STRING_H -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_WARNINGS /wd4710)
    endif ()
endif ()
```
This code doesn't lie! Only on windows :)

Nightwalker-87:
This code does not lie either: `src/common.c` :unamused: - Try it out for yourself - whether this is a useful way of implementation - that remains in question.

Nightwalker-87:
> unistd.h is used to build with msvc. So I propose to move it to win32/unistd/unistd.h

We can leave it in `src/win32/`. I only introduced a subfolder for getopt, because it derives from an external source along with a separate license and readme file.

BTW: Why don't we have it integrated here then:

CMakeLists.txt
```diff
@@ -81,50 +81,45 @@ endif ()
...
-    include_directories(src/getopt)
-    # Use string.h rather than strings.h and disable annoying warnings
-    add_definitions(-DHAVE_STRING_H -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_WARNINGS /wd4710)
+    set(STLINK_SOURCE "${STLINK_SOURCE};src/mmap.c;src/win32/mingw.c")
+    set(STLINK_HEADERS "${STLINK_HEADERS};src/mmap.h;src/win32/mingw.h")
+    if (MSVC)
+        include_directories(src/win32/getopt)
+        # Use string.h rather than strings.h and disable annoying warnings
+        add_definitions(-DHAVE_STRING_H -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_WARNINGS /wd4710)
+    endif ()
```
Nightwalker-87, line 122: unistd.h or it's cmake-flag should be included here...

src/st-util/gdb-server.c
Nightwalker-87: :+1:

slyshykO:
>We can leave it in src/win32/
our unistd.h should lay in differ folder than mingw.h cause MinGW include the custom realisation of unistd.h

slyshykO:
> 
> 
> This code does not lie either: `src/common.c` 😒 - Try it out for yourself - whether this is a useful way of implementation - that remains in question.

how to move mmap   -  https://github.com/slyshykO/stlink/commit/cacb493eabf41b41d25e5a83d85f8dd7bc3d5dc6

CMakeLists.txt
```diff
@@ -81,50 +81,45 @@ endif ()
...
 
-if (WIN32 OR MINGW OR MSYS)
-    include_directories(src/mingw)
-    set(STLINK_SOURCE "${STLINK_SOURCE};src/mmap.c;src/mingw/mingw.c")
-    set(STLINK_HEADERS "${STLINK_HEADERS};src/mmap.h;src/mingw/mingw.h")
-endif ()
-
-if (MSVC)
+if (WIN32 OR MINGW OR MSYS OR MSVC)
     include_directories(src/win32)
```
slyshykO, line 115: our custom unistd.h prevoiusly connected here. Only for msvc as is getopt

Nightwalker-87:
> > This code does not lie either: `src/common.c` unamused - Try it out for yourself - whether this is a useful >way of implementation - that remains in question.
> 
> how to move mmap - [slyshykO@cacb493](https://github.com/slyshykO/stlink/commit/cacb493eabf41b41d25e5a83d85f8dd7bc3d5dc6)

Yes, somehow like this. Let me have a closer look first. :wink:

CMakeLists.txt
Nightwalker-87: I got it. Thx.

CMakeLists.txt
Nightwalker-87: This has become obsolete by now.

Nightwalker-87: @slyshykO: Have a look at these lines again now.

slyshykO:
I'd prefer that all conditions were in brackets even if it contains only one line.

prefer 

```cpp
if( x ) {
    printf("%s", "Hello \n");
}
```

over

```cpp
if ( x ) printf("%s", "Hello \n");
```


And it is strange for me that all return values covered in parentheses.
Is it added something to code security?

Nightwalker-87:
Yeah, we can set up bracket for if-statements in general. The return statement can either be with or without brackets as far as i can see, but not only if the value is negative. I'd need to look that up.

Nightwalker-87:
@chenguokai & @slyshykO you may review now.
I'm not planning to add anything else.
This PR should introduce a somehow common and unified code appearance and style.
Any remaining leftovers may be corrected later in time should any be found.

src/win32/win32_socket.h
```diff
@@ -9,19 +9,20 @@
...
 #pragma warning(pop)
 #endif
 
-/* winsock doesn't feature poll(), so there is a version implemented
- * in terms of select() in mingw.c. The following definitions
- * are copied from linux man pages. A poll() macro is defined to
- * call the version in mingw.c.
+/* winsock doesn't feature poll(), so there is a version implemented in terms of select() in mingw.c.
+ * The following definitions are copied from linux man pages.
+ * A poll() macro is defined to call the version in mingw.c.
```
slyshykO, line 25: We have changed the file name. Because of it, it should be also changed in the comment.

src/win32/win32_socket.c
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    LARGE_INTEGER perf_cnt, start, now;
+
+    QueryPerformanceFrequency(&perf_cnt);
+    QueryPerformanceCounter(&start);
+
+    do
+        QueryPerformanceCounter((LARGE_INTEGER*)&now);
+    while ((now.QuadPart - start.QuadPart) / (float)perf_cnt.QuadPart * 1000 * 1000 < waitTime);
```
slyshykO, line 310: Think there is also need brackets for "do while" loop.

src/win32/win32_socket.h
Nightwalker-87: Will be fixed in the next commit.

src/win32/win32_socket.c
Nightwalker-87: Not only there, within this file. Will be fixed in the next commit.

src/win32/win32_socket.h
```diff
@@ -22,7 +22,7 @@
 
 /* winsock doesn't feature poll(), so there is a version implemented in terms of select() in mingw.c.
```
slyshykO, line 23: line 23. mingw.c still exists

Nightwalker-87: Sorry, I missed that one.

PR #993, opened 2020-06-21
Add support for FreeBSD's libusb reimplementation

Crest:
FreeBSD reimplemented the libusb API, but their
implementation of libusb_set_debug() expects an
enum libusb_debug_level instead of an
enum libusb_log_level. This causes excessive
logging on FreeBSD drowning out all expected
output.

To keep the changes to other platforms minimal
the FreeBSD specific code is kept inside an
ifdef block and the enum values are recreated
as magic numbers just like the log levels for
other platforms.

### DIFF
--- a/src/stlink-lib/logging.c
+++ b/src/stlink-lib/logging.c
@@ -66,7 +66,7 @@ int ugly_log(int level, const char *tag, const char *format, ...) {
  *  - LIBUSB_LOG_LEVEL_DEBUG (4)   : debug and informational messages are printed to stderr
  */
 int ugly_libusb_log_level(enum ugly_loglevel v) {
-#if __FreeBSD__
+#ifdef __FreeBSD__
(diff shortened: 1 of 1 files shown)


chenguokai:
If I got this, the previous `#if` failed to set attributes that it intended to set. Am I right?

Crest:
The `__FreeBSD__` define is the official way to check if code is compiled for FreeBSD. It is defined as the target FreeBSD major version (some integer greater than 0), but undefined on other platforms. I used the wrong preprocessor directive to check if the code is compiled for FreeBSD. It works on FreeBSD because `__FreeBSD__` evaluates to 12 which is true, but fails to build on other platforms because `__FreeBSD__` is undefined. The correct condition to check for is `defined(__FreeBSD__)` which can be shortened to `#ifdef __FreeBSD__`.

PR #997, opened 2020-06-25
[feature] Option bytes on the STM32F767 ZIT6 Nucleo-144

rutgerhendriks:
This PR adds --area=option_bytes_boot_add, --area=optcr and --area=optcr1 for the STM32F767 ZI


rutgerhendriks:
Nightwalker-87 it seems I have merged upstream again. Could you have a look at the PR now? It still works here ;-)

src/stlink-lib/flash_loader.c
```diff
@@ -3,213 +3,210 @@
 #include <unistd.h>
 
 #include <stlink.h>
+#include "flash_loader.h"
```
Nightwalker-87: We can leave the brackets < > here, as our makefile explicitly takes care of finding this header file.
Thus it is unified throughout the codebase:

src/st-flash/flash_opts.c
```diff
@@ -2,11 +2,12 @@
 #include <stdio.h>
 #include <string.h>
 
-#include <flash.h>
+#include "flash.h"
```
Nightwalker-87: We can leave the brackets < > here, as our makefile explicitly takes care of finding this header file.
Thus it is unified throughout the codebase.

Nightwalker-87: Correction: My mistake. It's ok like this. Just leave it as it is.

src/stlink-lib/flash_loader.c
Nightwalker-87: Correction: My mistake. It's ok like this. Just leave it as it is.

.github/workflows/ccpp.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: make
+      run: make
```
Nightwalker-87, line 17: I think we don't need this file.

src/st-flash/flash.c
```diff
@@ -25,16 +25,25 @@ static void cleanup(int signum) {
     exit(1);
 }
 
-static void usage(void) {
-    puts("command line:   ./st-flash [--debug] [--reset] [--opt] [--serial <serial>] [--format <format>] [--flash=<fsize>] [--freq=<Hz>] {read|write} <path> [addr] [size]");
+static void usage(void)
+{
```
Nightwalker-87, line 29: Style (minor): Can we move the bracket { behind the previous argument?

src/st-flash/flash.c
```diff
@@ -126,21 +136,20 @@ int main(int ac, char** av) {
         goto on_error;
     }
 
-    if (o.cmd == FLASH_CMD_WRITE) {                                             // write
+    if (FLASH_CMD_WRITE == o.cmd)                                               // write
+    {
```
Nightwalker-87, line 140: Style (minor): Can we move the bracket { behind the previous argument?

src/st-flash/flash.c
```diff
@@ -207,30 +228,64 @@ int main(int ac, char** av) {
...
+                o.size = sl->flash_size;
+            }
+            else if ((o.size == 0) && (o.addr >= sl->sram_base) && (o.addr < sl->sram_base + sl->sram_size)) {
+                o.size = sl->sram_size;
+            }
+            err = stlink_fread(sl, o.filename, FLASH_FORMAT_IHEX == o.format, o.addr, o.size);
 
             if (err == -1) {
-                printf("could not read option bytes (%d)\n", err);
+                printf("stlink_fread() == -1\n");
```
Nightwalker-87, line 241: Maybe a more comprehensible Error-Message?

Nightwalker-87:
@rutgerhendriks: Can you have another look at this as soon as you find time?

src/stlink-lib/usb.c
```diff
@@ -710,6 +710,7 @@ int _stlink_usb_read_mem32(stlink_t *sl, uint32_t addr, uint16_t len) {
 
     cmd[i++] = STLINK_DEBUG_COMMAND;
     cmd[i++] = STLINK_DEBUG_READMEM_32BIT;
+    DLOG("using address %#010x\n", addr);
```
grevaillot, line 713: does that really belong to that commit ? some lowlevel methods could use some better traces, but this is maybe out of that pr scope

src/st-flash/flash.h
```diff
@@ -11,7 +11,7 @@
 
 enum flash_cmd {FLASH_CMD_NONE = 0, FLASH_CMD_WRITE = 1, FLASH_CMD_READ = 2, FLASH_CMD_ERASE = 3, CMD_RESET = 4};
 enum flash_format {FLASH_FORMAT_BINARY = 0, FLASH_FORMAT_IHEX = 1};
-enum flash_area {FLASH_MAIN_MEMORY = 0, FLASH_SYSTEM_MEMORY = 1, FLASH_OTP = 2, FLASH_OPTION_BYTES = 3};
+enum flash_area {FLASH_MAIN_MEMORY = 0, FLASH_SYSTEM_MEMORY = 1, FLASH_OTP = 2, FLASH_OPTION_BYTES = 3, FLASH_OPTION_BYTES_BOOT_ADD = 4, FLASH_OPTCR = 5, FLASH_OPTCR1 = 6};
```
grevaillot, line 14: These enum value names (flash_optcr and optcr1) using registers target register names can be really confusing and likely to clash with actual target register defines.

src/st-flash/flash.c
```diff
@@ -51,13 +60,13 @@ int main(int ac, char** av) {
         return(-1);
     }
 
-    printf("st-flash %s\n", STLINK_VERSION);
+    ILOG("st-flash %s\n", STLINK_VERSION);
```
grevaillot, line 63: printf is used here in that utility, is that change really part of that PR ?

inc/stlink.h
```diff
@@ -120,7 +120,8 @@ enum stlink_flash_type {
     STLINK_FLASH_TYPE_UNKNOWN = 0,
     STLINK_FLASH_TYPE_F0,    // used by f0, f1 (except f1xl),f3. */
     STLINK_FLASH_TYPE_F1_XL, // f0 flash with dual bank, apparently */
-    STLINK_FLASH_TYPE_F4,    // used by f2, f4, f7 */
+    STLINK_FLASH_TYPE_F4,    // used by f2, f4 */
+    STLINK_FLASH_TYPE_F7,
```
grevaillot, line 124: Does F7 really need some specific flash code ? I've not checked again but AFAIR, it was pretty similar to f2/f4 and could be treated as well (same reg base, same bits for base operation). That would avoid some copypasting.

src/st-flash/flash.c
```diff
@@ -51,13 +60,13 @@ int main(int ac, char** av) {
...
 
     sl = stlink_open_usb(o.log_level, 1, (char *)o.serial, o.freq);
 
     if (sl == NULL) { return(-1); }
 
-    if (sl->flash_type == STLINK_FLASH_TYPE_UNKNOWN) {
+    if (STLINK_FLASH_TYPE_UNKNOWN == sl->flash_type) {
```
grevaillot, line 69: Same question regarding PR scope, why that change here ? these are purely debatable style change

Nightwalker-87: Yeah, well true. AFAIK the previous state was the most common type in this toolset.

Nightwalker-87:
@grevaillot: I can remember that you intended or considered to contribute for the H7 from a personal feature branch earlier. So I absolutely don't mind if you take part in this discussion and bring in some helpful suggestions. If you fancy, you can also assign yourself to this topic (again).

Nightwalker-87:
@rutgerhendriks: Can you have another look at this as soon as you find time?

rutgerhendriks:
Yes I will have a look at it this weekend

Op do 2 jul. 2020 22:17 schreef nightwalker-87 <notifications@github.com>:

src/st-flash/flash.h
rutgerhendriks: I'm not really sure what you mean with the names being confusing and likely to clash.

The reason I added OPTCR and OPTCR1, is becase that is the actual name of the registers, so when you have the datasheet at hand, the OPTCR and OPTCR1 is what you see you need to use.
The FLASH_OPTION_BYTES_BOOT_ADD is the functional purpose of those specific option bytes. In the ST-LINK flash GUI you can set them directly using a name like that, so that made the most sense to me here too.

src/st-flash/flash.c
rutgerhendriks: The reason I changed this, is because one of the examples of using st-flash it something like this:
st-flash --area=option read > outputfile
The output file now has the st-flash version number in there which made no sense to me, you can't edit the file and write it back again now... so I removed the printf...

You're right it could be in a separate PR, I'll remove it here

inc/stlink.h
rutgerhendriks: I couldn't find the OPTCR and OPTCR1 registers on the F4, so to distinguish I added a new flash type. If you have another suggestion how to do this, I'm open to it

src/st-flash/flash.c
rutgerhendriks: I can remove these changes, they are not related to this PR, it's just something I'm used to do to prevent compiler warnings

Nightwalker-87: What sort of compiler warnings? It may be something reasonable to consider.

.github/workflows/ccpp.yml
rutgerhendriks: Removed it

src/st-flash/flash.c
rutgerhendriks: Sure, must have slipped in, fixed

src/st-flash/flash.c
rutgerhendriks: fixed

src/stlink-lib/usb.c
rutgerhendriks: Removed

src/st-flash/flash.c
rutgerhendriks: Fixed

src/st-flash/flash.c
rutgerhendriks: Below there is a link to the Travis CI build:
https://travis-ci.org/github/stlink-org/stlink/builds/705314574?utm_source=github_status&utm_medium=notification

It is giving the compiler warnings I was talking about:
[ 55%] Building C object CMakeFiles/st-flash.dir/src/st-flash/flash_opts.c.o
/home/travis/build/stlink-org/stlink/src/st-flash/flash_opts.c:298:29: error: 
      equality comparison with extraneous parentheses
      [-Werror,-Wparentheses-equality]
        } else if ((o->area == FLASH_OTP)) {
                    ~~~~~~~~^~~~~~~~~~~~
/home/travis/build/stlink-org/stlink/src/st-flash/flash_opts.c:298:29: note: 
      remove extraneous parentheses around the comparison to silence this
      warning
        } else if ((o->area == FLASH_OTP)) {
                   ~        ^           ~
/home/travis/build/stlink-org/stlink/src/st-flash/flash_opts.c:298:29: note: 
      use '=' to turn this equality comparison into an assignment
        } else if ((o->area == FLASH_OTP)) {
                            ^~
                            =

If by accident you would write:
FLASH_OTP = o->area (iso FLASH_OTP == o->area), the compiler would be able to catch that since you obviously can't assign a value to a #define. 
As it is now, o->area = FLASH_OTP is valid so the compiler can't help when you accidentally remove a '='.
It takes some getting used to in reading the comparison, but it helps.

Nightwalker-87: I need to look at this again to gain some understanding why this actually happens.

src/st-flash/flash.h
Nightwalker-87: If possible, we should align with the official documentation here.

inc/stlink.h
Nightwalker-87: If possible, we should align with the official documentation here.

src/st-flash/flash.c
Nightwalker-87: Maybe you should remove the inner brackets ( ):
`else if (o->area == FLASH_OTP)`
as this seems to compare (o->area == FLASH_OTP) == TRUE
Can you try this out?

Nightwalker-87:
There is a build failure for clang.

Nightwalker-87:
@rutgerhendriks: There are still some open points to address. Can you have a look?

src/st-flash/flash.h
rutgerhendriks: I'm not sure what you would want to change. The OPTCR and OPTCR1 are part of the official ST documentation and the ST-LINK flash GUI let's you wrote the BOOT_ADD bytes directly. So that's what I had implemented

inc/stlink.h
rutgerhendriks: What do you propose exactly?

src/st-flash/flash.h
Nightwalker-87: Nothing, as long as your proposal for the enum value names matches the ones in the datasheet, everything should be fine. You may resolve this conversation.

inc/stlink.h
Nightwalker-87: This is related to the previous discussion about `src/st-flash/flash.h`.
So if the enum value names don't match between the F7 and F4, and you are sure with it, we need to have specific code.
In this case this should resolve this thread as well.

src/st-flash/flash.c
rutgerhendriks: Thanks, I missed that one, that solved the build warnings

rutgerhendriks:
The build is clean now and I closed the comments. I think it's ready to be merged now, right?

src/st-flash/flash_opts.c
```diff
@@ -233,22 +239,21 @@ int flash_get_opts(struct flash_opts* o, int ac, char** av) {
...
 
             o->cmd = FLASH_CMD_READ;
         } else if (strcmp(av[0], "write") == 0) {
-            if (o->cmd != FLASH_CMD_NONE) { return(-1); }
+            if (FLASH_CMD_NONE != o->cmd) { return(-1); }
 
             o->cmd = FLASH_CMD_WRITE;
         } else if (strcmp(av[0], "reset") == 0) {
-            if (o->cmd != FLASH_CMD_NONE) { return(-1); }
+            if (FLASH_CMD_NONE != o->cmd) { return(-1); }
```
Nightwalker-87, line 256: Just to gain some knowledge out of this:
Does clang still have problems during compilation if you change the order back to the original state now?

Nightwalker-87:
> The build is clean now and I closed the comments. I think it's ready to be merged now, right?

We're almost there, yes. Thx for solving this main compilation issue. :+1:

src/st-flash/flash_opts.c
rutgerhendriks: I reverted those too now, waiting for travis check

Nightwalker-87: The CI build state already confirms that there are no further issues. :tada:

Nightwalker-87:
@rutgerhendriks: Thank you very much for your contribution! :+1:

Issue #468, opened 2016-09-16
STM32F103R8: Problems flashing device - unknown core / chip id! 0


kgoba:
- [x] Programmer/board type: Stlink/v2-onboard (STMF4DISCOVERY board)
- [x] Programmer firmware version: e.g STSW-LINK007 2.27.15
- [x] Operating system: Linux Ubuntu 16.04
- [x] Stlink tools version and/or git commit hash: 7023a9ab62b77102f5d48a34bfb8ad173de22346 and also earlier versions
- [x] Stlink commandline tool name: st-info / st-flash
- [x] Target chip (and optional board): STM32F103R8 ("Blue Pill" board)

I am using stm32f103r8 device (64k flash, 20k RAM) with a ST-LINK/V2 programmer (Discovery board with F407VG). I have multiple boards like these (http://wiki.stm32duino.com/index.php?title=Blue_Pill) and none of them seems to work with stlink. I have connected the GND/SWDIO/SWCLK and RESET pins to the SWD connector. Using this setup has worked for other chips. Problems include:
- not being able to read the flash correctly
- not being able to write the flash at all
- reporting wrong Flash size

Outputs from st-info and st-flash:

```
$ st-info --probe
Found 1 stlink programmers
 serial: 533f6d06493f55573224113f
openocd: "\x53\x3f\x6d\x06\x49\x3f\x55\x57\x32\x24\x11\x3f"
  flash: 26230784 (pagesize: 1024)
   sram: 20480
 chipid: 0x0410
  descr: F1 Medium-density device
```

Stlink recognizes the chip, though somewhat incorrectly, reporting 26230784 bytes of flash. I am not sure about the chipid, as I don't know the correct value. 

```
$ st-flash read orig.bin 0x8000000 0x10000
2016-09-16T11:14:25 INFO /home/karlis/build/stlink/src/common.c: Loading device parameters....
2016-09-16T11:14:25 INFO /home/karlis/build/stlink/src/common.c: Device connected is: F1 Medium-density device, id 0x20036410
2016-09-16T11:14:25 INFO /home/karlis/build/stlink/src/common.c: SRAM size: 0x5000 bytes (20 KiB), Flash: 0x1904000 bytes (25616 KiB) in pages of 1024 bytes
```

Returns almost immediately. The file is the correct size but does not seem to contain valid readout.

```
$ st-flash write orig.bin 0x8000000
[...]

Issue #62, opened 2012-02-10
[reset] st-flash does not work when CPU is in sleep mode


rogerdahl:
In general, the st-flash utility has been working fairly well for me. Thank you for your excellent work, guys. I believe the issues I've had, except for the one I am creating this ticket for, have open tickets already.

The problem I'm having is that I'm planing on creating an interrupt driven design. So I have written an experimental program that sets up an interrupt (connected to the user button) and then puts the CPU to sleep by running the __WFI() function. When I flash a program that runs that function, subsequent flashing with st-flash does not work. To get back in touch with the chip, I have to boot into Windows and use the STM-32 ST-LINK Utility from ST to erase the flash. A clue to what is wrong is that the STM-32 ST-LINK Utility does not work either, unless "Connect under Reset" is enabled in Settings.

After flashing the program that runs the __WFI() function, any attempt to flash causes the following error:

dahl@ubuntu:~/Desktop/3/stlink/example/32l_lcd$ make write
../../flash/flash write lcd.bin 0x08000000
2012-02-09T23:25:54 INFO src/stlink-common.c: Loading device parameters....
2012-02-09T23:25:54 WARN src/stlink-common.c: unknown chip id! 0
stlink_sram_flash() == -1
make: **\* [write] Error 255

Any attempt to erase causes (note how the program displays a different chip id):

make: **\* [write] Error 255
dahl@ubuntu:~/Desktop/3/stlink/example/32l_lcd$ ../../flash/flash erase
2012-02-09T23:28:01 INFO src/stlink-common.c: Loading device parameters....
2012-02-09T23:28:01 WARN src/stlink-common.c: unknown chip id! 0xe0042000
Mass erasing

If I hold the Reset button while flashing:

ahl@ubuntu:~/Desktop/3/stlink/example/32l_lcd$ ../../flash/flash erase
2012-02-09T23:28:28 INFO src/stlink-common.c: Loading device parameters....
2012-02-09T23:28:28 INFO src/stlink-common.c: Device connected is: L1 Med-density device, id 0x10386416
2012-02-09T23:28:28 INFO src/stlink-common.c: SRAM size: 0x4000 bytes (16 KiB), Flash: 0 bytes (0 KiB) in pages of 256 bytes

Even after I erase the flash in Windows, I get the following. This is fixed by pushing the Reset button on the card:

dahl@ubuntu:~/Desktop/3/stlink/example/32l_lcd$ make write
../../flash/flash write lcd.bin 0x08000000
[...]

PR #1051, opened 2020-10-14
st-flash: respect --connect-under-reset option

timothytylee:
The **--connect-under-reset** option is already parsed by **src/st-flash/flash_opts.c**.  Respect this option when opening the USB connection.

### DIFF
--- a/src/st-flash/flash.c
+++ b/src/st-flash/flash.c
@@ -61,7 +61,9 @@ int main(int ac, char** av) {
 
     printf("st-flash %s\n", STLINK_VERSION);
 
-    sl = stlink_open_usb(o.log_level, 1, (char *)o.serial, o.freq);
+    sl = stlink_open_usb(o.log_level,
+            o.connect_under_reset ? 2 : 1,
+            (char *)o.serial, o.freq);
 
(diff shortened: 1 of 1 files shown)


Nightwalker-87:
Can you verify that this closes some related issues to `--connect-under-reset` ? There were some previous approaches to address this issue already. It would be good to know why they obviously failed.

geoffreymbrown:
That looks correct to me.  It looks like the necessary bit of code at the
USB open didn't get committed previously
Geoffrey

Nightwalker-87:
#62, #468 may be related to this issue. Is there any possibility for verification and testing?

timothytylee:
So the reason for this patch, is that I'm reviving the patches from #1033 to add STM32H7 support.  What I have found is that unless I connect to the target board under reset, the chip ID would be read back incorrectly.

I cannot comment on other STM32 targets.

Nightwalker-87:
@timothytylee: Thx for the feedback. Maybe someone else can look at it with the focus on some other targets as well. Thus I'll leave this PR open to feedback for another few days.

Nightwalker-87 (approved):

chenguokai (approved):

PR #1060, opened 2020-10-26
[feature] st-util: Add specialized memory map for STM32H7 devices

cmdrf:
Add a special memory map for STM32H742, STM32H743/753 and STM32H750 (at least). It includes the additional SRAM regions in these chips.

Data taken from linker script generated for H750 by CubeMX software and verified against table 7 of reference manual RM0433 (except for option byte area, which is not documented):
```
MEMORY
{
DTCMRAM (xrw)      : ORIGIN = 0x20000000, LENGTH = 128K
RAM_D1 (xrw)      : ORIGIN = 0x24000000, LENGTH = 512K
RAM_D2 (xrw)      : ORIGIN = 0x30000000, LENGTH = 288K
RAM_D3 (xrw)      : ORIGIN = 0x38000000, LENGTH = 64K
ITCMRAM (xrw)      : ORIGIN = 0x00000000, LENGTH = 64K
FLASH (rx)      : ORIGIN = 0x8000000, LENGTH = 128K
}
```

### DIFF
--- a/src/st-util/gdb-server.c
+++ b/src/st-util/gdb-server.c
@@ -504,6 +504,24 @@ static const char* const memory_map_template_F7 =
     "  <memory type=\"rom\" start=\"0x1fff0000\" length=\"0x20\"/>"         // option byte area
     "</memory-map>";
 
+static const char* const memory_map_template_H7 =
+    "<?xml version=\"1.0\"?>"
+    "<!DOCTYPE memory-map PUBLIC \"+//IDN gnu.org//DTD GDB Memory Map V1.0//EN\""
+    "     \"http://sourceware.org/gdb/gdb-memory-map.dtd\">"
+    "<memory-map>"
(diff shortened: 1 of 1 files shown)


cmdrf:
Looking at https://github.com/stlink-org/stlink/commit/1e2092173aae4ce0c256ff345cf647ec868f5435#diff-44941250a4c135836a4d4d41b9c21282b90e379111bcb9fbab698d66e99872ba I think I might have the option bytes wrong...

Nightwalker-87:
Please verify that your patch applies to the specs and update it by pushing another commit if necessary.

cmdrf:
Updated and confirmed against the manual.

cmdrf:
Noo, something else is wrong now:
> Overlapping regions in memory map: ignoring

Is it overlapping with the peripheral registers?

cmdrf:
I checked the peripheral registers in the manual, and it's correct (except that the length is off by one, but that's the same in all other memory maps, too). So it seems a separate option byte area is not required at all.

Nightwalker-87 (approved):

Issue #136, opened 2013-03-13
[feature]  ITM functionality for ST-Link V2 and STM32Fxx chipsets


obe1line:
I have written a utility that can read ITM trace data using a ST-Link V2. This functionality would be useful to have in st-util if possible. The only problem is when it detects a packet with length 0xF8xx, which I assume is an error code - possibly overrun. It recovers, but loses some of the data. Other than that, I have had it running for a few hours capturing data from an STM32F107Z.

See http://obe1line.github.com/stlink-trace/ for further details.

xor-gate:
Somebody has enhanced your stlink-trace: https://github.com/avrhack/stlink-trace

Nightwalker-87:
Can anybody contribute here?

xor-gate:
@Nightwalker-87 we probably should close this issue in favor of duplicate #187. I also got a personal mail from @simplerobot which implements the ITM functionality in https://github.com/simplerobot/stlink/tree/st-trace but he seems to be unable to open a pull request.

Nightwalker-87:
There should be no limitations on opening PRs. However I actively invited him to contribute to the project. Maybe it helps to solve this issue, as the referenced ticket seems to only partially address this issue. ;-)

simplerobot:
Thanks @Nightwalker-87.  I've submitted a code review.  Here is the error I was originally getting:

An owner of this repository has limited the ability to open a pull request to users that have contributed to this repository in the past.

PR #1072, opened 2020-12-04
[feature] ITM functionality for ST-Link V2 and STM32Fxx chipsets

simplerobot:
This change implements feature #136.

The main changes:
* Adds enable/disable/read trace methods to stlink library.
* Adds flags to stlink and chipid for trace support.
* Adds an additional command line tool st-trace.

I did bench testing while developing this, including testing with incorrect clock speeds and with overruns.  However, this testing has been done with an extremely limited number of only STM32F4 processors.  I could use some help running this on a wider variety of hardware.  I am currently using st-trace in a production environment that automatically runs test suites on multiple hardware devices.

This code review is larger than I would hope for my first to this project.  I've tried to make the code fit in with the project, but please don't hold back if I've missed the mark anywhere.

Thanks,
John


Nightwalker-87 (dismissed):

slyshykO:
Merge with develop branch restored setting of the **has_dual_bank** field for few chips, and it breaks compilation.

Nightwalker-87:
The `has_dual_bank` has been introduced by @Ant-ON just recently and obviously on purpose, so recently involved authors should discuss on how to proceed. I was not aware that there was a functional conflict related to this.

slyshykO:
@simplerobot, there is nothing here you should sorry about. It just working moments.

Ant-ON:
@Nightwalker-87 The `has_dual_bank` flag has already been used before me. I continued to use it.
@simplerobot I think replacing `has_dual_bank` to the flag is a good idea.

Nightwalker-87:
@Ant-ON: I don't mind as long as everyone is satisfied and nothing breaks. Thus I recommend to also discuss with @timothytylee and @slyshykO to find the best possible solution.

slyshykO:
Using flags for me look more appropriate. But bools also acceptable.
The one strong restriction is that the new fields we should add with the care of ABI.
So new fields should be added to the end of the structure.

For convenient work with bit fields usually used a few simple macro like this:
```cpp
#define READ_BIT(REG, BIT)    ((REG) & (BIT))
```

Ant-ON:
@simplerobot I agree with the @slyshykO. Flag checks look simple enough too. They are already used in the other places of project:
```C
if (sl->version.flags & STLINK_F_HAS_GETLASTRWSTATUS2)
```

src/st-trace/trace.c
```diff
@@ -0,0 +1,610 @@
...
+        error = true;
+    }
+
+    if (error && !settings->force)
+        return false;
+
+    return true;
+}
+
+static void ConvertSerialNumberTextToBinary(const char* text, char binary_out[STLINK_SERIAL_MAX_SIZE]) {
```
slyshykO, line 245: all functions have named in _snake_case_, so I think in this file this also should be a rule.

Nightwalker-87:
So we are waiting for our reviewers then. :smiley:

src/st-trace/trace.c
```diff
@@ -0,0 +1,610 @@
...
+    WLOG("Chip ID: 0x%04x\n", stlink->chip_id);
+    WLOG("****\n");
+}
+
+int main(int argc, char** argv)
+{
+    signal(SIGINT, &abort_trace);
+    signal(SIGTERM, &abort_trace);
+    signal(SIGSEGV, &abort_trace);
+    signal(SIGPIPE, &abort_trace);
```
slyshykO, line 522: The signal function doesn't work properly on windows, and even more, SIGPIPE didn't define there so this code can't be compiled for windows.
So, if it not hard for you,  can you change the signal processing in a cross-platform way?
Example : https://github.com/stlink-org/stlink/blob/develop/src/st-util/gdb-server.c#L248

Nightwalker-87:
You may also squash some commits before merging, as there are quite a few already by now.
That would at least help in the context of readability.

src/stlink-lib/chipid.h
```diff
@@ -69,6 +69,10 @@ enum stlink_stm32_chipids {
     STLINK_CHIPID_STM32_WB55             = 0x495
 };
 
+
+#define CHIP_F_HAS_SWO_TRACING  (1 << 0)
```
Ant-ON, line 73: I think it's better to include ```has_dual_bank``` field in the flags too

src/st-trace/trace.c
```diff
@@ -0,0 +1,622 @@
...
+static bool g_abort_trace = false;
+
+static void abort_trace() {
+    g_abort_trace = true;
+}
+
+#if defined(_WIN32)
+BOOL WINAPI CtrlHandler(DWORD fdwCtrlType) {
+    abort_trace();
+    return FALSE;
```
slyshykO, line 154: add (void)fdwCtrlType here to suppress warning about unused parameter

src/st-trace/trace.c
```diff
@@ -0,0 +1,622 @@
...
+    WLOG("Unknown Sources: %s\n", buffer);
+
+    WLOG("Chip ID: 0x%04x\n", stlink->chip_id);
+    WLOG("****\n");
+}
+
+int main(int argc, char** argv)
+{
+#if defined(_WIN32)
+    SetConsoleCtrlHandler((PHANDLER_ROUTINE)abort_trace, TRUE);
```
slyshykO, line 527: Use as handler **CtrlHandler** instead of **abort_trace**.

src/st-trace/trace.c
simplerobot: Done.

src/st-trace/trace.c
simplerobot: Done.

inc/stlink.h
```diff
@@ -83,6 +83,33 @@ enum target_state {
 
 #define STLINK_V3_MAX_FREQ_NB            10
 
+#define STLINK_TRACE_BUF_LEN               2048
+#define STLINK_V2_MAX_TRACE_FREQUENCY   2000000
+#define STLINK_V3_MAX_TRACE_FREQUENCY  24000000
+#define STLINK_DEFAULT_TRACE_FREQUENCY  2000000
+
+/* Cortex Debug Control Block */
+#define DCB_DHCSR 0xE000EDF0
```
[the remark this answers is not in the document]
simplerobot: Done.

src/st-trace/trace.c
simplerobot: Done.

src/st-trace/trace.c
simplerobot: Nice catch.

src/stlink-lib/chipid.h
simplerobot: Done.  I don't have the hardware to test this well.  Could you verify that flash still works with dual bank systems?

Ant-ON: I looked at the changes. Everything should work.

simplerobot: Thanks!

slyshykO:
@Nightwalker-87, this PR looks good to me.

Nightwalker-87 (approved):

Nightwalker-87:
@slyshykO It still fails: https://travis-ci.org/github/stlink-org/stlink/builds/748613319
Unfortunately I can only test when triggering a manual build after merging.
This Travis CI integration has become such a mess! You can't use it before merging any more. :angry:

slyshykO:
@Nightwalker-87 think it is time for Github Actions

slyshykO:
There is #1076, so maybe we beg @Ant-ON to fix one more clang's warning?

Nightwalker-87:
... he seems to be collecting everything up, involuntarily :rofl:

Nightwalker-87:
> @Nightwalker-87 think it is time for Github Actions

At least the recent custom triggered build of the `develop` branch now showed up, so there still is some hope... :hourglass_flowing_sand:

slyshykO:
@Nightwalker-87 Travis CI has changed its pricing model.
There is a blog post - https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing .
So somebody who has access to Travis account should check credits.

Nightwalker-87:
I know, but it is still free for OSS-Projects, so we don't pay anything for it.

slyshykO:
> I know, but it is still free for OSS-Projects, so we don't pay anything for it.

Free, but more limited than before. Looks like we encounter those limits.

Nightwalker-87:
I'll have an eye on it. Should we run into further problems, one may think about appropriate alternatives.
Thanks for the input. :+1:

PR #1076, opened 2020-12-07
Bugfixes for compilation with clang

Ant-ON:
I fixed code compiling by clang. The format of argument problem (linux-mingw) is fix @slyshykO in #1073

PR #1078, opened 2020-12-10
Bugfixes for compilation with clang

simplerobot:
This pull request combines @Ant-On's fix #1076 as well as fixes for two new clang bugs introduced in #1072.

### DIFF
--- a/src/stlink-lib/usb.c
+++ b/src/stlink-lib/usb.c
@@ -1109,7 +1109,7 @@ int _stlink_usb_read_trace(stlink_t* sl, uint8_t* buf, size_t size) {
         return -1;
     }
 
-    size_t trace_count = read_uint16(sl->q_buf, 0);
+    uint16_t trace_count = read_uint16(sl->q_buf, 0);
 
     if (trace_count > size) {
         ELOG("read_trace insufficient buffer length\n");
--- a/src/st-trace/trace.c
+++ b/src/st-trace/trace.c
@@ -410,7 +410,7 @@ static bool read_trace(stlink_t* stlink, st_trace_t* trace) {
 
 static void check_for_configuration_error(stlink_t* stlink, st_trace_t* trace, uint32_t trace_frequency) {
     // Only check configuration one time after the first 10 seconds of running.
-    uint32_t elapsed_time_s = time(NULL) - trace->start_time;
+    time_t elapsed_time_s = time(NULL) - trace->start_time;
     if (trace->configuration_checked || elapsed_time_s < 10) {
         return;
     }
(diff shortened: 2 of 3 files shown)


src/stlink-lib/usb.c
```diff
@@ -1109,7 +1109,7 @@ int _stlink_usb_read_trace(stlink_t* sl, uint8_t* buf, size_t size) {
         return -1;
     }
 
-    size_t trace_count = read_uint16(sl->q_buf, 0);
+    uint32_t trace_count = read_uint16(sl->q_buf, 0);
```
Ant-ON, line 1112: Why limit the bit width of a variable? Can unsigned be used?

slyshykO: It is more easily support stable compilation for different bitness and many compilers & OSes when used values with a defined size.

Ant-ON: My opinion is that shouldn't use bounded types unnecessarily. It is easier for the system to work with native tipe (int).
ps You can ignore my remark. Everything will work well without any changes.

slyshykO: > It is easier for the system to work with native tipe (int)

It is not true anymore. An unsigned has a 4-byte size for 32 and 64-bit platforms. But should be printed with different format flags.
For "real" platform fast you should use something like _int_fast32_t_.

Ant-ON: What are the differences in the flags? I always thought that "%u" it is unsigned.
int_fast32_t on most systems (32-bit and higher) is int. Int should be the most effective.

xor-gate: My suggestion should still use the size_t for unsigned sizes or loop counters. But in this case we do a read_uint16 and should put it in a uint16.

simplerobot: I've updated this to use an unsigned (int).  This is compatible with the timeout parameter to libusb_bulk_transfer, which was breaking the clang build.  However, it is also unsigned with allows the comparison with size_t in the function.

simplerobot: Sorry @xor-gate, I had made this change before seeing your comment.

Should I use an unsigned as suggested by @Ant-ON, or a uint16_t as suggested by @xor-gate?

Nightwalker-87: In my opinion we should use bounded types whenever possible _and_ useful.
Especially on transition from unbounded types to the latter there should be intensive testing on several operating system environments before submitting any PR to rule out regression bugs.

With the next release (simultaneously) I'll further introduce a proper testing/nightly branch as well.
I'm not so happy with the current situation of patching around on `develop` to get it working again (some people want to use it), while CI is buggy/not functional. However this is not meant to blame anyone - it solely describes the current state. :unamused:

simplerobot: Updated to uint16_t.  Tested with gcc 9.3.0 and clang 10.0.0-4.

Nightwalker-87 (approved):

PR #1080, opened 2020-12-19
[feature] Included ITM functionality for building with MSVC

slyshykO:


Nightwalker-87:
@simplerobot: I would have added it already, if it was working properly.

Nightwalker-87 (approved):

PR #1085, opened 2021-01-14
Flash_loader: increase wait rounds for slow boards

seeseemelk:
Certain systems can take a while to boot and flash (because of a slow power supply or because the core is running at a lower clock speed).
Commit b8813fab9734521a1857189aab5dbf2a30d37ea4 broke st-link for several of our boards. Increasing the `WAIT_ROUNDS` define by a little fixes the issues.

This was tested on a custom board with an STM32L471 and a custom board with an STM32L496.

### DIFF
--- a/src/stlink-lib/flash_loader.c
+++ b/src/stlink-lib/flash_loader.c
@@ -368,7 +368,7 @@ int stlink_flash_loader_run(stlink_t *sl, flash_loader_t* fl, stm32_addr_t targe
  * the OS uses, the wait until the error message is reduced to the same order of magnitude
  * as what was intended. -- REW.
  */
-#define WAIT_ROUNDS 30
+#define WAIT_ROUNDS 40
 
     // wait until done (reaches breakpoint)
     for (i = 0; i < WAIT_ROUNDS; i++) {


Nightwalker-87:
I'll probably not change this unless there has been widespread testing with various other board types and / or programmers.
This parameter has been changed several times which not only lead to improvements but also introduced side effects.
I'll leave this open for others in the community to have their say. Should there be a common sense on this, we may change it.

chenguokai:
> Commit b8813fa broke st-link for several of our boards. Increasing the WAIT_ROUNDS define by a little fixes the issues.

I am a bit confused. It seems that this particular commit does not change the wait time ideally, without considering the operating system scheduler and the function call cost which may cause the sleep time to be longer than expected. 

Anyway, since there are some boards that may be affected by this timing issue, I think increasing this period is fine.

Ant-ON (commented):
@seeseemelk I think it's better to rewrite this part to use time_ms(). This will make possible to set the timeout in ms.
https://github.com/stlink-org/stlink/blob/4bfaab0ea993fe86da9f2aef612036adcf8ef58d/src/stlink-lib/helper.h#L4

Nightwalker-87:
@Ant-ON Sounds like a good idea to me. :+1:

Nightwalker-87 (changes requested):
I agree on the idea to rewrite this part in order to have to have a flexible solution.

Nightwalker-87:
@slyshykO: What's your opinion on this?

slyshykO (approved):

slyshykO:
> @slyshykO: What's your opinion on this?

I am Ok with increasing wait time.

Nightwalker-87:
@seeseemelk: Are you going to address the proposal from @Ant-ON ?

seeseemelk:
I'm afraid I don't really have the time to focus on this at the moment, especially given that we are switching from ST-Links to J-Links.

Nightwalker-87 (approved):
I'm going to let this pass, but shall there be any regression bug reports, that point back to this, we should not continue to tune this setting frequently.

Issue #607, opened 2017-06-20
STM32F1/F4/L0/L4: flash loader run error


dougyeager:
**NOTICE: The issue may be closed without notice when not enough information is provided!**

Thank you for giving feedback to the stlink project. Take some time to fill out
 check boxes with a X in the following items so developers and other people can try to
 to find out what is going on. And add/remove what is appropriate to your problem.

When submitting a feature request, try to reuse the list and add/remove what is appropriate.
 Place a `X` between the brackets `[X]` to mark the list item.

- [stlink/v2 ] Programmer/board type: e.g Stlink/v1, Stlink/v2, Stlink/v2-onboard
- [lastest from ST ] Programmer firmware version: e.g STSW-LINK007 2.27.15
- [ mac] Operating system: e.g Linux, Mac OS X, Windows (with specific version)
- [today ] Stlink tools version and/or git commit hash: e.g v1.1.0/git-c722056
- [ st-flash] Stlink commandline tool name: e.g `st-info`, `st-flash`, `st-util`
- [ STM32F446x] Target chip (and optional board): e.g STM32F402VG (STM32Fxxx Discovery)

A as-detailed description possible of the problem with debug output when available.

Output:
Dougs-MacBook-Pro:bin dyeager$ ./st-flash write out.bin 0x8000000
st-flash 1.3.0
2017-06-20T18:28:52 INFO /Users/jerry/Downloads/stlink-master/src/common.c: Loading device parameters....
2017-06-20T18:28:52 INFO /Users/jerry/Downloads/stlink-master/src/common.c: Device connected is: F446 device, id 0x10006421
2017-06-20T18:28:52 INFO /Users/jerry/Downloads/stlink-master/src/common.c: SRAM size: 0x20000 bytes (128 KiB), Flash: 0x80000 bytes (512 KiB) in pages of 131072 bytes
2017-06-20T18:28:52 INFO /Users/jerry/Downloads/stlink-master/src/common.c: Attempting to write 16536 (0x4098) bytes to stm32 address: 134217728 (0x8000000)
Flash page at addr: 0x08004000 erasedEraseFlash - Sector:0x1 Size:0x4000 
2017-06-20T18:28:53 INFO /Users/jerry/Downloads/stlink-master/src/common.c: Finished erasing 2 pages of 16384 (0x4000) bytes
2017-06-20T18:28:53 INFO /Users/jerry/Downloads/stlink-master/src/common.c: Starting Flash write for F2/F4/L4
[...]

PR #1087, opened 2021-01-18
fixes infinite busy loop

bmbove:
I'm not super well-versed in this codebase, but I've been struggling to upload and debug on an STM32L412K8

I've got it working with an instruction modification in the flash loader. Original code:

```
wait:
    ldrh r3, [r10]  ; load flash status register (upper) value into r3
    tst r3, #0x1    ; logical AND flash status register value and literal 0x01 (check FLASH_BSY flag for a 1)
    beq wait        ; if zero flag is set, jump to wait
```

when `FLASH_BSY == 0`, the `tst` instruction should set the zero flag. `beq` takes the branch if the 0 flag is set, so if flash is not busy, it continues to wait (incorrect behavior). changing `beq` to `bne` fixes this.

I'm hung up on this though because it seems like a really naive fix on my part. I'm willing to dig a little deeper if need be. This might help some folks with #607.


bmbove:
actually, I'm going to investigate a little more.

PR #1130, opened 2021-04-16
Test: checkout out the automated build failures. 

rewolff:
DO NOT MERGE. 

I am under the impression that the automated build-checks fail for reasons that are outside the scope of the pull-submitter.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -4182,3 +4182,4 @@ int stlink_fwrite_option_bytes(stlink_t *sl, const char* path, stm32_addr_t addr
     return(err);
 }
 
+


Nightwalker-87:
Automated build checks are part of project maintenance and they are known to be working and have been tested.
Solely the tests for Ubuntu 16.04 do not work currently due to a package issue in the Ubuntu repository.
This is known and will be taken care on the project maintenance side.

In general build-test configuration and GitHub integrations are both strongly related to project organisation and thus not the business of anyone else than members of **stlink-org**.

PR #1145, opened 2021-05-24
Expanded and revised list of chips

Ant-ON:
The list of chips is expanded with missing ones in accordance with the ST documentation
All the names of the chips are brought to the same style
Fixed incorrect size of SRAM of the G0 chips (not critical, but it would be better if it would correspond to the documentation)


Nightwalker-87 (approved):

Nightwalker-87:
@Ant-ON I switched this to testing because I had updated the names and also unified the chip-id defines as part of the chip-config separation from the source code. So one has to resolve the differences: Use of the changed defines + additional changes to the descriptions from your side. Both is useful, but conflicts need to be resolved.

Nightwalker-87:
As I have noticed, some defines need to be updated as well. As currently proposed (see **testing** branch) I'd prefer to use a small "x" for placeholders in MCU names/types to distinguish from the capital "X" used by ST Microelectronics in some MCU part names.

Ant-ON:
@Nightwalker-87 In the `test` branch is missing https://github.com/stlink-org/stlink/commit/eafb4d17925296cf8c31f4309294c424d7174186 and https://github.com/stlink-org/stlink/commit/f02618c53c3d312aad8bb1cf8f0fb369b1af6ba3 commits. Conflicts due to their absence.

Ant-ON:
@Nightwalker-87 In most codding standards the enum names naming according to the rule: "Labels All Upper Case with '_' Word Separators". Using the "x" char is not entirely correct.

Nightwalker-87:
One may also use another symbol like $ etc. as well as long it is part of the ASCII character set.

Nightwalker-87:
@Ant-ON Can we proceed here to allow for the subsequent conclusion of @rewolff's contributon?

Ant-ON:
@Nightwalker-87 Everything is ready here and nothing needs to be changed. In the test branch is lack previous corrections by other authors. "X" is not used in the names of MСU. "X" is used in enum names and this follows their naming standards.

Nightwalker-87:
@Ant-ON No, there were some renamed defines that have already been replaced in other source files. These changes are in **testing** already. Thus this has to be reviewed with care. There are relevant / useful changes from both versions.

Nightwalker-87:
Merged manually preserving the defines from the current branch.
Some more precise descriptions should now be used to allign related defines.
These changes will take place on testing branch soon.

PR #1150, opened 2021-06-06
Fix for 'libusb_devices were leaked' when no STLINK programmer was found

gustavolaureano:
When searching for interfaces, the libstlink is not unreferencing the probed USB devices when no STLINK programmer was found, leading to "libsub_devices were leaked" warnings, as shown below:
image

This warning is evidenced when using the libstlink to poll for ST-Link interfaces on Windows (I am developing an application that waits for a STLINK programmer to be connected)

A description of the problem and solution can be found here:
https://libusb-devel.narkive.com/KfZ3490q/warning-some-libusb-devices-were-leaked

### DIFF
--- a/src/stlink-lib/usb.c
+++ b/src/stlink-lib/usb.c
@@ -1203,7 +1203,7 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, enum connect_type connect,
     libusb_set_option(slu->libusb_ctx, LIBUSB_OPTION_LOG_LEVEL, ugly_libusb_log_level(verbose));
 #endif
 
-    libusb_device **list;
+    libusb_device **list = NULL;
     // TODO: We should use ssize_t and use it as a counter if > 0.
     // As per libusb API: ssize_t libusb_get_device_list (libusb_context *ctx, libusb_device ***list)
     int cnt = (int)libusb_get_device_list(slu->libusb_ctx, &list);
@@ -1232,7 +1232,7 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, enum connect_type connect,
         ILOG("bus %03d dev %03d\n", devBus, devAddr);
     }
 
-    while (cnt--) {
+    while (cnt-- > 0) {
         struct libusb_device_handle *handle;
 
         libusb_get_device_descriptor(list[cnt], &desc);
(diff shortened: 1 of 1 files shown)


src/stlink-lib/usb.c
```diff
@@ -1273,6 +1273,7 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, enum connect_type connect,
 
     if (cnt < 0) {
         WLOG ("Couldn't find %s ST-Link devices\n", (devBus && devAddr) ? "matched" : "any");
+        libusb_free_device_list(list, 1);
```
Ant-ON, line 1276: If we want to use `libusb_free_device_list` on error, then we must force initialize the `list` variable before calling `libusb_get_device_list`.
As I understand it, the bug is a Windows backend. Better to wrap this call in `#ifdef _WIN32 ... #endif` and add a comment that this is a bug fix.

Nightwalker-87:
@gustavolaureano Can you address this?

src/stlink-lib/usb.c
gustavolaureano: Hi @Ant-ON , I am not sure if it is a windows-only bug, from what I understood the correct procedure is to always free the devices (like we do on line 1287, for example)

And about making sure that the list was initialized: it must have been initialized at line 1209, the only way of arriving here without having the list initialized is if the call to libusb_get_device_list at line 1209 returned an error, and in this case we would already have problems starting from the while loop at line 1235, which is not accounting for a negative cnt value coming from the error code,  in my opinion, this is the actual issue we must address

gustavolaureano: Hi @Ant-ON  , I am not sure if it is a windows-only bug, from what I understood the correct procedure is to always free the devices (like we do on line 1287, for example)

And I initialized the list, so it is not a problem if we try to free a list that was not allocated (it would just return [here](https://github.com/libusb/libusb/blob/e782eeb2514266f6738e242cdcb18e3ae1ed06fa/libusb/core.c#L879))

I also added a check for cnt > 0, as libusb_get_device_list (line 1209) could return a negative value in case of an error, leading to out of bounds memory access inside the while loop

Ant-ON: @gustavolaureano I agree, now there should be no errors in the work. Thank you!

Ant-ON (approved):

Nightwalker-87 (approved):

PR #1179, opened 2021-08-15
Fixed few warnings for msvc about type conversion with possible lost data

slyshykO:


Nightwalker-87 (approved):

Nightwalker-87:
I like the attempt to use precise typedefs. However the best solution in the medium term would be to use them in the definition of core variables already... (Which shall not be part of this PR however.) It's a general design issue.

PR #1188, opened 2021-09-08
Improved chipid checks and printouts

andysan:
This pull request addresses the following issues in the chipid code:

- Dumps of the `stlink_chipid_params` struct didn't include the `flash_size_reg` field.
- Comparisons between legacy hard-coded param structs and the new config files was done incorrectly and included a string pointer in the comparison. As a consequence, those comparisons always failed and incorrectly printed an error message.
- `STLINK_FLASH_TYPE_UNKNOWN` wasn't treated as a legal flash type value despite being used by `unknown_device.chip`.

N.B.: None of the chpid configs currently contain a valid `flash_size_reg` field, so comparisons will still fail after this patch.

### DIFF
--- a/src/stlink-lib/chipid.c
+++ b/src/stlink-lib/chipid.c
@@ -916,7 +916,7 @@ void dump_a_chip (FILE *fp, struct stlink_chipid_params *dev) {
     fprintf(fp, "flags %d\n\n", dev->flags);
 }
 
-static int chipid_params_eq(struct stlink_chipid_params *p1, struct stlink_chipid_params *p2)
+static int chipid_params_eq(const struct stlink_chipid_params *p1, const struct stlink_chipid_params *p2)
 {
     return p1->chip_id == p2->chip_id &&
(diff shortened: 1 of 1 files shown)


Nightwalker-87 (approved):

src/stlink-lib/chipid.c
```diff
@@ -915,6 +916,22 @@ void dump_a_chip (FILE *fp, struct stlink_chipid_params *dev) {
     fprintf(fp, "flags %d\n\n", dev->flags);
 }
 
+static int chipid_params_eq(struct stlink_chipid_params *p1, struct stlink_chipid_params *p2)
```
gszy, line 919: Please consider making the function parameters `const`:

```suggestion
static int chipid_params_eq(const struct stlink_chipid_params *p1, const struct stlink_chipid_params *p2)
```

gszy (approved):

Nightwalker-87 (approved):

Issue #1196, opened 2021-11-05
[STM32H735]: Can't access external memory


c-grant:
Issue is similar to issue #1158, except it is in relation the the externally mapped memory busses 

 Programmer/board type: [STLINK V3 on STM32H735-DK]
 Operating system an version: [Windows 10 19042.1083]
 stlink tools version and/or git commit hash: [1.7.0]
 stlink commandline tool name: [st-util]
 Target chip (and board, if applicable): [STM32H735 on STM32H735-DK]

When tryin to read OSPI2 on the STM32H735-DK the memory could not be read.

Investigating the gdb-server.c these regions were also missing.

Pull has been created and HyperRAM on OSPI2 can now be read

c-grant:
Fixed by #1197

PR #1197, opened 2021-11-05
Updated gdb-server.c to allow external memory access on STM32H73xx

c-grant:
This is to fix issue #1196 

This has been build and tested with STM32H735 Discovery kit on the OSPI2 HyperRam

### DIFF
--- a/src/st-util/gdb-server.c
+++ b/src/st-util/gdb-server.c
@@ -518,6 +518,8 @@ static const char* const memory_map_template_H72x3x =
     "    <property name=\"blocksize\">0x%x</property>"
     "  </memory>"
     "  <memory type=\"ram\" start=\"0x40000000\" length=\"0x1fffffff\"/>"   // peripheral regs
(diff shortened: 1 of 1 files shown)


src/st-util/gdb-server.c
```diff
@@ -518,6 +518,8 @@ static const char* const memory_map_template_H72x3x =
     "    <property name=\"blocksize\">0x%x</property>"
     "  </memory>"
     "  <memory type=\"ram\" start=\"0x40000000\" length=\"0x1fffffff\"/>"   // peripheral regs
+    "  <memory type=\"ram\" start=\"0x60000000\" length=\"0x3fffffff\"/>"   // External Memory
+	"  <memory type=\"ram\" start=\"0xC0000000\" length=\"0x1fffffff\"/>"   // External device
```
Ant-ON, line 522: Better to use 4 spaces instead tab

Ant-ON (approved):

Nightwalker-87 (approved):

PR #1201, opened 2021-11-23
Disable extremely verbose function level tracing in FreeBSD's libusb

Crest:
The FreeBSD libusb allows only very limited control over logging. Even log level 1 produces function level tracing output which drowns out any normal output. The only usable log level for normal use is level 0 which only logs errors.


Crest:
Sorry for messing up the branch. I'll clean it up and submit the pull request again.

PR #1202, opened 2021-11-24
Fixed compliation for OpenBSD 7.0

Crest:
On OpenBSD libusb is provided by the libusb1 port and the existing `Findlibusb.cmake` module failed to detect it properly because it installs to `/usr/local` in accordance with the OpenBSD filesystem hierarchy. Teaching `cmake` the OpenBSD file system layout gets us half the way.

The second and last problem I've encountered is that the `MINIMAL_API_VERSION` required under OpenBSD is undeclared causing a compile error in `libusb_settings.h`. I used libusb 1.0.23p2 installed from the official binary package repository, but I don't see any reasons why OpenBSD should require a newer libusb version than other platforms.

### DIFF
--- a/cmake/modules/Findlibusb.cmake
+++ b/cmake/modules/Findlibusb.cmake
@@ -43,6 +43,23 @@ elseif (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")              # FreeBSD; libusb is 
         message(FATAL_ERROR "Expected libusb library not found on your system! Verify your system integrity.")
     endif ()
 
+elseif (CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")              # OpenBSD; libusb-1.0 is available from ports
+    FIND_PATH(
+        LIBUSB_INCLUDE_DIR NAMES libusb.h
+        HINTS /usr/local/include
+        PATH_SUFFIXES libusb-1.0
+        )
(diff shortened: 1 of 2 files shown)


Ant-ON (approved):

Nightwalker-87 (approved):

nbonfils:
I hope it's not to out of topic, but I've been so frustrated with AVR related things on OpenBSD, the fact that  stlink built out of the box without needing to touch anything was so relieving. Literally made my whole weekend better!
I take it's thanks to @Crest and also to the stlink-org team (@Nightwalker-87 and @xor-gate it seems) to have made a software that's not unfriendly to the "fringe" OSes.
Thank you <3

Nightwalker-87:
Thank you very much. It is good to hear such feedback from affected users.
It is indeed a key objective of this toolset to maintain best possible support for a wide range of operating environments (as a far as we can sustain compatibility regarding libraries and packages). As it appears we are on a good way here... :smiley:

May I include your review on our main page to give additional orientation to new users? Please let me know.

nbonfils:
Of course @Nightwalker-87, feel free to do so.
This key objective definitely deserves more appreciation. It might even be worth mentioning in the README as this sort of details usually hints users like me that this software and the maintainers care. We totally need more of that.
Thank you again.

Issue #1075, opened 2020-12-06
Should we care about ABI and semantic versioning for stlink libs?


slyshykO:
I see that at least 2 structures that expose over lib border changed its shape between ver 1.6.0 and 1.6.1 and still changes.
It is struct **stlink_t** and struct **stlink_chipid_params**.
So if we care about ABI we should change their shape in backwards-compatible form.

Nightwalker-87:
Can you explain this issue in more detail and give some examples?

slyshykO:
This is how struct stlink_chipid_params look like in version 1.6.0
```cpp
struct stlink_chipid_params {
	uint32_t chip_id;
	char *description;
	enum stlink_flash_type flash_type;
	uint32_t flash_size_reg;
	uint32_t flash_pagesize;
	uint32_t sram_size;
	uint32_t bootrom_base;
	uint32_t bootrom_size;
};
```
This is how it looks now :
```cpp
struct stlink_chipid_params {
    uint32_t chip_id;
    char *description;
    enum stlink_flash_type flash_type;
    bool has_dual_bank;
    uint32_t flash_size_reg;
    uint32_t flash_pagesize;
    uint32_t sram_size;
    uint32_t bootrom_base;
    uint32_t bootrom_size;
    uint32_t option_base;
    uint32_t option_size;
};
```

As you can see there is bool has_dual_bank was added before  flash_size_reg field.
This is mean that in old structure flash_size_reg have offset 12 bytes. Now an offset for flash_size_reg 16. An old code doesn't know about these changes so it will misread  flash_size_reg   from new has_dual_bank field.

slyshykO:
But, actually, I didn't dig deeper. So it is a question that we haven't similar breakage for previous releases.

So there is a question should we care about it in future?

Nightwalker-87:
I see what you mean. I believe this is also what you meant with the preference to allocate new parameters at the end of structures.

chenguokai:
Maybe this link is somehow useful here:
[https://stackoverflow.com/questions/25947997/how-does-sizeofstruct-help-provide-abi-compatibility](https://stackoverflow.com/questions/25947997/how-does-sizeofstruct-help-provide-abi-compatibility)

I think the intermediate structure may help.

slyshykO:
> I believe this is also what you meant with the preference to allocate new parameters at the end of structures.

Yes, I have ABI in my mind when saying that. But this trick works only if in user code used pointers to structs.

[...]

PR #1210, opened 2022-01-05
Fix for libusb_kernel_driver_active & error handling for st.st_size

### LINKED ISSUE #1211: [refactoring] usb.c: Can we remove some unused code?
In usb.c there are next lines:

```
stlink_t *stlink_open_usb(...) {
line 1144:
    // TODO: Reading a environment variable in a usb open function is not very nice, this should
    // be refactored and moved into the CLI tools, and instead of giving USB_BUS:USB_ADDR a real
    // stlink serial string should be passed to this function. Probably people are using this
    // but this is very odd because as programmer can change to multiple busses and it is better
    // to detect them based on serial.
    char *device = getenv("STLINK_DEVICE");

```
Here supposed that there is an environment variable in user system with name "STLINK_DEVICE" and value "Device_bus:Device_Address". In case of presence such variable in user system function gets device bus and device address from USB devices list and compare them with ones in environment variable. If do not match then proceed with next USB device from list.

There is no such variable in my system. I can not find any place in code which set this variable. Function stlink_open_usb gets serial of ST-LINK in 3rd parameter and uses it to find right USB device. So may it be useful to completely remove all code which works with getenv function and variables device, devBus and devAddr?

### LINKED ISSUE #1214: STM32L-Discovery (STM32L152R8T6) can't write firmware to chip with st-flash on Windows 10
- [x] Programmer/board type: STM32L-Discovery, STLINK /V2-onboard
- [x] Operating system and version: Windows 10 x64
- [x] stlink tools version: 1.7.0, current develop branch
- [x] stlink commandline tool name: `st-flash`
- [x] Target chip and board: STM32L152R8T6, board STM32L-Discovery

Commandline output:

`mmap() size_t overflow for file mb.bin`

Expected/description:

Debugging revealed that problem is in file common.c, function map_file, in code below:
```
  if (sizeof(st.st_size) != sizeof(size_t)) {
    // on 32 bit systems, check if there is an overflow
    if (st.st_size > (off_t)SSIZE_MAX) {
      fprintf(stderr, "mmap() size_t overflow for file %s\n", path);
      goto on_error;
    }
  }
```

Type of _st.st_size_ is _off_t_, which means _long int_. Value is size of firmware file and for me is 602 bytes.
_SSIZE_MAX_ defines somewhere in libc headers. It means maximum size of signed integer. But in my system it compiled in _long long int_. Not 4 bytes _long int_ as off_t, but 8 bytes _long long int_! So there is overflow in casting types without warnings during compile. As a result, instead of:
`if (602 > 2^32-1) {` 
I got:
`if (602 > -1) {` 
And then function returned with error message.

Proposed solution, which works on my computer:
In header section of _common.c_ define macro _OFF_T_MAX_ such a way:

```
#ifndef OFF_T_MAX
#define OFF_T_MAX 1073741824 // long int max value
#endif
```

In map_file function replace _SSIZE_MAX_ by newly defined _OFF_T_MAX_.

Also please note that _SSIZE_MAX_ defined in _unistd.h_ in such a way:
```
#define ssize_t int
#ifndef SSIZE_MAX
#define SSIZE_MAX ((sizeof(ssize_t) == 4) ? 1073741824 : ‭4611686018427387904‬)
#endif
```
There is an error in numbers. _SSIZE_MAX_ is max. value of signed int and must be equal 0x400... . But here it equals 0x7FFF ... in both cases, which is true for UNsigned int. For signed int both values are -1. But in my system _SSIZE_MAX_ defined in libc, so this macro is not used.

hydroconstructor:
According to libusb API documentation, function **_libusb_kernel_driver_active_** is not available on Windows. So I think it will be useful not to compile it on Windows platform.

(Closes #1211) (Closes #1214)


Nightwalker-87 (approved):

slyshykO:
@hydroconstructor, can you provide the direct link to documentation?

hydroconstructor:
There is one more change, typo on line 1293 (see second commit)

Nightwalker-87 (approved):

src/stlink-lib/usb.c
```diff
@@ -1221,6 +1221,8 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, enum connect_type connect,
 
     libusb_free_device_list(list, 1);
 
+// libusb_kernel_driver_active is not available on Windows.
+#if !defined(_WIN32)
```
Ant-ON, line 1193: Theoretically, this can be omitted. The function will simply return LIBUSB_ERROR_NOT_SUPPORTED.
https://github.com/libusb/libusb/blob/1170db79bdc4155c3d5a64787fcc4f8d56b26174/libusb/core.c#L1974-L1989

But let it be a less of unnecessary code)

Ant-ON (approved):

src/stlink-lib/usb.c
```diff
@@ -1134,33 +1134,8 @@ stlink_t *stlink_open_usb(enum ugly_loglevel verbose, enum connect_type connect,
 #endif
 
     libusb_device **list = NULL;
-    // TODO: We should use ssize_t and use it as a counter if > 0.
```
Ant-ON, line 1137: In Linux user space, you can set `export STLINK_DEVICE="..."` before run command

src/stlink-lib/usb.c
Nightwalker-87: yeah, just leave it, it's ok - we have way more serious issues to deal with in other parts of the codebase.

src/stlink-lib/usb.c
Nightwalker-87: Yes, that does indeed make sense.

Nightwalker-87:
Looking at this more closely, I think we can merge it to `develop` as it does not conflict with the `chipid_cleanup` branch.

src/stlink-lib/usb.c
hydroconstructor: @Ant-ON, maybe who writed this code some time ago knows it. But there is no such export anywhere in source files of  project. And there is comment with TODO, why to do so is bad idea (use serial instead of env. variable). So I think it can be done, but should not.

Nightwalker-87 (approved):

src/stlink-lib/usb.c
Nightwalker-87: @hydroconstructor You can safely assume that nobody is aware of it (anymore), and if so, very likely no longer available to be asked. I also believe nobody is willing to crawl the whole long project history to find out about it...

Nightwalker-87 (approved):

src/stlink-lib/usb.c
Nightwalker-87: @hydroconstructor I think this is not what @Ant-ON meant with his comment.
Nevertheless I'll leave a more detailed explanation to him.

src/common.c
```diff
@@ -2200,8 +2204,9 @@ static int map_file(mapped_file_t *mf, const char *path) {
 
   if (sizeof(st.st_size) != sizeof(size_t)) {
     // on 32 bit systems, check if there is an overflow
-    if (st.st_size > (off_t)SSIZE_MAX) {
-      fprintf(stderr, "mmap() size_t overflow for file %s\n", path);
+	if (st.st_size > (off_t)MAX_FILE_SIZE  /*1 GB*/ ) {
```
gszy, line 2207: https://github.com/stlink-org/stlink/issues/1214#issuecomment-1013743705

src/stlink-lib/usb.c
gszy: Wouldn’t dropping support for this environment variable kind of break backward compatibility? It is a feature people _may_ currently use.

src/stlink-lib/usb.c
gszy: I’m in favor of handling `LIBUSB_ERROR_NOT_SUPPORTED` instead of adding this `_WIN32` check.

src/stlink-lib/usb.c
Nightwalker-87: We will likely break backwards compatibility frequently as long as we don't clean-up the codebase from duplicate and what I'd call nested implementations and as long as ABI compatibility is not defined in any form (see #1075).
To achieve the latter one needs a well structured and modular code implementation and design.
So that's why there is no value to it from my perspective at the current point in time, as it would block any further attempts to clean up the codebase. Also the library is not documented well enough yet to serve as a core component for 3rd-party software.

Nightwalker-87: @gszy @hydroconstructor Please proceed with this.

src/stlink-lib/usb.c
Nightwalker-87: @gszy Ok, go for it then... I don't mind.

doc/man/st-util.1
```diff
@@ -19,8 +19,6 @@ option, the default \f[B]4242\f[R] port will be used.
 Stlink version 2 is used by default unless the option
 \f[B]\[en]stlinkv1\f[R] is given.
 .PP
```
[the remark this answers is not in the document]
Nightwalker-87: Correct.

src/common.c
Nightwalker-87: It would be nice if we could include the related necessary changes.

src/stlink-lib/usb.c
gszy: Well, it seems no change is necessary: we only check if `libusb_kernel_driver_active() == 1`, so if the function returned  `LIBUSB_ERROR_NOT_SUPPORTED`, the program would continue in the same way it would do if the function returned `0` (or anything other than `1`). I don’t think we have to check if we’re on WIN32, since libusb already does it.

It may be useful to handle other return values properly, like `LIBUSB_ERROR_INVALID_PARAM`, though it may be out of this PR’s scope.

I understand @hydroconstructor’s change could somewhat reduce stlink binary size on Windows, but I don’t think these additional two lines of code are worth the benefit.

src/stlink-lib/usb.c
gszy: Other than the comment above, feel free to remove this code.

src/common.c
gszy: I’m not sure how to mark it up with GitHub suggestion syntax. I suggest:

- removing the `MAX_FILE_SIZE` `ifdef`/`define`/`endif`,
- replacing the `st.st_size > (off_t)SSIZE_MAX` test with `st.st_size > (intmax_t) SIZE_MAX`,
- perhaps removing the `sizeof(st.st_size) != sizeof(size_t)` test as well.

src/stlink-lib/usb.c
Nightwalker-87: @hydroconstructor Let's remove this check.

Nightwalker-87:
@hydroconstructor Please address the remaining points.

src/common.c
Nightwalker-87: Addressed in commit f6cfd1bfe3eb50b90befe03293808727c66cb8bd.

Nightwalker-87 (approved):

Nightwalker-87:
Thanks @hydroconstructor :+1:

Issue #1199, opened 2021-11-11
Wrong flash size reported for various chips


10110111:
I'm debugging firmware on STM32H725VGT that has some code above 0x8020000. When I try to access this memory e.g. via `disas` command in GDB, I get _"Cannot access memory at address 0x802xxxx"_. I've tracked it down to `info mem` reporting flash region from 0x08000000 to 0x08011c00. The 0x11c00 size is also what I get from `st-info --flash`. This size is wrong: the chip has 1MiB of flash.

I've also tried this on another chip, the one on STM32F407G-DISC1, and `st-info --flash` reports flash size of 0x3fffc00, which would be almost 64MiB! The actual size is still 1MiB.

I'm running stlink from `develop` branch, at e97f438075ff5128630dfa3668f6ac7bd1181c51. OS is Ubuntu 20.04 (Linux 5.4.0-65-generic).

Ant-ON:
@10110111 Can you post output of `st-info --flash`?
Can you post value of 0x1FF1E880 register (read in GDB: `x/1uw 0x1FF1E880`)?

10110111:
The register read returns 1024. The output is below.
```
$ st-info --flash
---------- old ------------
# Chip-ID file for H72x/H73x
#
chip_id 0x483
description H72x/H73x
flash_type  10
flash_size_reg 0x1ff1e880
flash_pagesize 0x20000
sram_size 0x20000
bootrom_base 0x1ff00000
bootrom_size 0x20000
option_base 0x5200201c
option_size 0x2c
flags 2

---------- new ------------
# Chip-ID file for H72x/H73x
#
chip_id 0x483
description H72x/H73x
flash_type  10
flash_size_reg 0x0
flash_pagesize 0x20000
sram_size 0x20000
bootrom_base 0x1ff00000
bootrom_size 0x20000
option_base 0x5200201c
option_size 0x2c
flags 2

0x11c00
```

Ant-ON:
@10110111 Have you edited the configuration files? `flash_size_reg` should be `0x1ff1e880`...

10110111:
No, I don't even know where they are :) . I just did `make install` and went on to run `st-util`.

Ant-ON:
I don't know where they too... Apparently someone broke they :(
Unfortunately, I can only advise you to comment their reading function in the `chipid.c`. Using an array with chip descriptions works stably.

10110111:
[...]

PR #1213, opened 2022-01-06
Erase addr size / section of the flash memory with st-flash

antoinefaure:
Hi,

This PR is to add the possibility to only erase a section of the flash, whereas as of today it is only possible to erase the whole flash (as far as I know).

I've added this functionality in a different function, as I didn't want to mess with the parallelism in `stlink_erase_flash_mass`
Calling erase without specifying an address or size will have the same behavior as before, i.e. erasing the whole flash.

Let me know if you have any remarks.
Thanks!

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -2951,7 +2951,7 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
   return check_flash_error(sl);
 }
 
-int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size) {
+int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size, bool align_size) {
   if (base_addr < sl->flash_base || (base_addr + size) > (sl->flash_base + sl->flash_size)) {
     ELOG("Invalid address or size\n");
     return (-1);
@@ -2971,8 +2971,8 @@ int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size
   do {
     size_t page_size = stlink_calculate_pagesize(sl, addr);
 
-    // Check if we are going further than the requested size
-    if ((addr + page_size) > (base_addr + size)) {
+    // Check if size is aligned with a page, unless we want to completely erase the last page
+    if ((addr + page_size) > (base_addr + size) && !align_size) {
       ELOG("Invalid size (not aligned with a page). Page size at address %#x is %#lx\n", addr, page_size);
       return -1;
     }
@@ -3000,7 +3000,7 @@ int stlink_erase_flash_mass(stlink_t *sl) {
   if (sl->flash_type == STLINK_FLASH_TYPE_L0 ||
       sl->flash_type == STLINK_FLASH_TYPE_WB) {
 
-    stlink_erase_flash_section(sl, sl->flash_base, sl->flash_size);
+    err = stlink_erase_flash_section(sl, sl->flash_base, sl->flash_size, false);
 
   } else {
     wait_flash_busy(sl);
@@ -3525,7 +3525,10 @@ int stlink_write_flash(stlink_t *sl, stm32_addr_t addr, uint8_t *base,
   stlink_core_id(sl);
 
   // Erase this section of the flash
-  stlink_erase_flash_section(sl, addr, len);
+  if (stlink_erase_flash_section(sl, addr, len, true) < 0) {
+    ELOG("Failed to erase the flash prior to writing\n");
+    return (-1);
+  }
 
   if (eraseonly) {
     return (0);
--- a/src/common.c
+++ b/src/common.c
@@ -2952,24 +2952,25 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
 }
 
 int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size) {
-    // erase each page
-    int i = 0, num_pages = (int)(size / sl->flash_pgsz);
-    for (i = 0; i < num_pages; i++) {
-      // addr must be an addr inside the page
-      stm32_addr_t addr =
-          (stm32_addr_t)base_addr + i * (stm32_addr_t)sl->flash_pgsz;
+  stm32_addr_t addr = (stm32_addr_t)base_addr;
 
-      if (stlink_erase_flash_page(sl, addr)) {
-        WLOG("Failed to erase_flash_page(%#x) == -1\n", addr);
-        return (-1);
-      }
+  do {
+    size_t page_size = stlink_calculate_pagesize(sl, addr);
 
-      fprintf(stdout, "-> Flash page at %5d/%5d erased\n", i, num_pages);
-      fflush(stdout);
+    if (stlink_erase_flash_page(sl, addr)) {
+      WLOG("Failed to erase_flash_page(%#x) == -1\n", addr);
+      return (-1);
     }
 
-    fprintf(stdout, "\n");
-    return 0;
+    fprintf(stdout, "-> Flash page at %#x erased (size: %#lx)\n", addr, page_size);
+    fflush(stdout);
+
+    // check the next page is within the range to erase
+    addr += page_size;
+  } while (addr < (base_addr + size));
+
+  fprintf(stdout, "\n");
+  return 0;
 }
 
 int stlink_erase_flash_mass(stlink_t *sl) {
--- a/src/common.c
+++ b/src/common.c
@@ -2952,6 +2952,11 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
 }
 
 int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size) {
+  if (base_addr < sl->flash_base || (base_addr + size) > (sl->flash_base + sl->flash_size)) {
+    ELOG("Invalid address or flash size\n");
+    return (-1);
+  }
+
   stm32_addr_t addr = (stm32_addr_t)base_addr;
 
   do {
--- a/src/common.c
+++ b/src/common.c
@@ -2951,23 +2951,47 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
   return check_flash_error(sl);
 }
 
-int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size, bool align_size) {
-  if (base_addr < sl->flash_base || (base_addr + size) > (sl->flash_base + sl->flash_size)) {
-    ELOG("Invalid address or size\n");
+// Check if an address and size are within the flash
+int stlink_check_address_range_validity(stlink_t *sl, stm32_addr_t addr, size_t size) {
+  if (addr < sl->flash_base || addr >= (sl->flash_base + sl->flash_size)) {
+    ELOG("Invalid address, it should be within 0x%08x - 0x%08lx\n", sl->flash_base, (sl->flash_base + sl->flash_size -1));
     return (-1);
   }
+  if ((addr + size) > (sl->flash_base + sl->flash_size)) {
+    ELOG("The size exceeds the size of the flash (0x%08lx bytes available)\n", (sl->flash_base + sl->flash_size - addr));
+    return (-1);
+  }
+  return 0;
+}
 
-  stm32_addr_t addr = sl->flash_base;
+// Check if an address is aligned with the beginning of a page
+int stlink_check_address_alignment(stlink_t *sl, stm32_addr_t addr) {
+  stm32_addr_t page = sl->flash_base;
+
+  while (page < addr) {
+    page += stlink_calculate_pagesize(sl, page);
+  }
+
+  if (page != addr) {
+    return -1;
+  }
+
+  return 0;
+}
+
+int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size, bool align_size) {
+  // Check the address and size validity
+  if (stlink_check_address_range_validity(sl, base_addr, size) < 0) {
+    return -1;
+  }
 
   // Make sure the requested address is aligned with the beginning of a page
-  while (addr < base_addr) {
-    addr += stlink_calculate_pagesize(sl, addr);
-  }
-  if (addr != base_addr) {
+  if (stlink_check_address_alignment(sl, base_addr) < 0) {
     ELOG("The address to erase is not aligned with the beginning of a page\n");
     return -1;
   }
 
+  stm32_addr_t addr = base_addr;
   do {
     size_t page_size = stlink_calculate_pagesize(sl, addr);
 
@@ -3498,22 +3522,13 @@ int stlink_write_flash(stlink_t *sl, stm32_addr_t addr, uint8_t *base,
   // check addr range is inside the flash
   stlink_calculate_pagesize(sl, addr);
 
-  if (addr < sl->flash_base) {
-    ELOG("addr too low %#x < %#x\n", addr, sl->flash_base);
-    return (-1);
-  } else if ((addr + len) < addr) {
-    ELOG("addr overruns\n");
-    return (-1);
-  } else if ((addr + len) > (sl->flash_base + sl->flash_size)) {
-    ELOG("addr too high\n");
-    return (-1);
-  } else if (addr & 1) {
-    ELOG("unaligned addr 0x%x\n", addr);
+  // Check the address and size validity
+  if (stlink_check_address_range_validity(sl, addr, len) < 0) {
     return (-1);
   } else if (len & 1) {
     WLOG("unaligned len 0x%x -- padding with zero\n", len);
     len += 1;
-  } else if (addr & (sl->flash_pgsz - 1)) {
+  } else if (stlink_check_address_alignment(sl, addr) < 0) {
     ELOG("addr not a multiple of current pagesize (%u bytes), not supported, "
          "check page start address and compare with flash module organisation "
          "in related ST reference manual of your device.\n",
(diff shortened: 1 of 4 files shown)


Nightwalker-87 (approved):

slyshykO:
Hi.
The newly added function 'stlink_erase_flash_section' didn't work in most cases. Because few MCUs have sectors of the same size. Even more, many MCUs have sector sizes that differ from those contained in `sl->flash_pgsz`.

For example, I've tested erase sections with STM32F746G-DISCO.
The first sector of stm32f746 is 16kB. 
But  `sl->flash_pgsz` for it equal only 2kB (I don't know why).
So, I run
```sh
st-flash.exe erase 0x8000000 4096
```

And the application erased 2 sectors with a total volume of 32kB, instead of 1 sector. 

```sh
>st-flash.exe erase 0x8000000 4096
st-flash 1.7.0-130-g42790f3
2022-01-09T16:06:02 INFO common.c: F74x/F75x: 320 KiB SRAM, 1024 KiB flash in at least 2 KiB pages.
EraseFlash - Sector:0x0 Size:0x8000 -> Flash page at     0/    2 erased
EraseFlash - Sector:0x1 Size:0x8000 -> Flash page at     1/    2 erased
```

src/common.c
```diff
@@ -2951,19 +2951,13 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
...
-  int err = 0;
-
-  // TODO: User MER bit to mass-erase WB series.
-  if (sl->flash_type == STLINK_FLASH_TYPE_L0 ||
-      sl->flash_type == STLINK_FLASH_TYPE_WB) {
+int stlink_erase_flash_section(stlink_t *sl, stm32_addr_t base_addr, size_t size) {
     // erase each page
-    int i = 0, num_pages = (int)(sl->flash_size / sl->flash_pgsz);
-
+    int i = 0, num_pages = (int)(size / sl->flash_pgsz);
```
slyshykO, line 2956: You cannot calculate the number of memory segments by simple division. In reality, many MCUs have sectors of different sizes.

slyshykO (changes requested):

Nightwalker-87:
I could have spotted that indeed... Good that we usually have at least two reviewers.

antoinefaure:
Hi,
Thanks for your feedback, that makes sense. 
I'll update the PR to use `stlink_calculate_pagesize` to get the size of each page that's going to be erased, does this seem right to you ?

slyshykO:
@antoinefaure 
Yes, the calculation of the size of all sections to be erased is the correct way. But also, you should check user input in `st-flash`.
The address should be a flash address and you should not try to erase absent sectors.

Nightwalker-87:
@antoinefaure Please read the respective value from the .chip files located in `/config/chips` along with `struct stlink_chipid_params` in `chipid.h`.
However we should keep in mind that changes will apply to this file soon, but as long as we stay on the `develop` branch there should be no conflict with current achievements. Most important is **not** to make use of the old chip definitions in `chipid.c`. This would lead to additional efforts when merging the current feature branch. It's probably best if you have a look at branch `chipid_cleanup` to get an idea on what is going on in the background.

antoinefaure:
I've just pushed 3 new commits:
- [d3bf145](https://github.com/stlink-org/stlink/pull/1213/commits/d3bf1453fd8c2f57faaa91bec01f24f0b00af347) to use ` stlink_calculate_pagesize` to get the page size, instead of relying on `sl->flash_pgsz`
- [8a1535e](https://github.com/stlink-org/stlink/pull/1213/commits/8a1535e2d2589a6594026b07ee3f4490533de692) to check the validity of the size and address in `stlink_erase_flash_section` instead of `main`. That way, if an invalid size or address is given, it will return an error instead of erasing the whole flash (which was the behavior in my first commit).
- [f277fdb](https://github.com/stlink-org/stlink/pull/1213/commits/f277fdb67750033716101ec0d0c48add46c5e075) adds a few safety checks, to make sure the address and size are aligned with a page. It will therefore return an error if an user tries to erase half a page for example.

@slyshykO 
I'm not sure what you had in mind when saying this : 
> you should not try to erase absent sectors

Is there something else that needs to be checked ?

@Nightwalker-87 
What should I read in the chips files ? It seems I have everything I need in `sl` and with `stlink_calculate_pagesize`.
Am I missing something ?

Nightwalker-87:
@antoinefaure Take a look at `chipid.h` where we have `struct stlink_chipid_params`. This is the format structure of the previously mentioned chip-files. You need to ensure that `flash_pagesize` is read for the current chip on which you try to erase a flash section.

Nightwalker-87:
Ah, I see - this is still an old implementation from `common.c`... All addresses are hardcoded for a second time (it did not even read from the old chipid data-base in `chipid.c`). OMG what a mess! :disappointed:
Yeah, use it for now, but we should have that re-coded as well to read the flash pagesizes from the chip-files instead.

antoinefaure:
That's what I understood when going through `chipid.h` and `common.c`, hence my confusion.
Thanks for checking!

slyshykO:
@Nightwalker-87 
Maybe we should wait with reviewing this PR till `chipid_cleanup` is being merged?

src/common.c
```diff
@@ -2951,30 +2951,57 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
...
+    return (-1);
+  }
+
+  stm32_addr_t addr = sl->flash_base;
+
+  // Make sure the requested address is aligned with the beginning of a page
+  while (addr < base_addr) {
+    addr += stlink_calculate_pagesize(sl, addr);
+  }
+  if (addr != base_addr) {
```
slyshykO, line 2966: This line denies me to erase any sector except the first.

slyshykO (changes requested):

Nightwalker-87:
@slyshykO I thought about that also at first, but actually as I'm not planning to change anything in `common.c` yet, it should be ok I think. It may be better to start a new approach for the latter (otherwise the changes become to large for a merge).

src/common.c
Nightwalker-87: Yeah, can we find a proper solution for that?

antoinefaure: Are you sure the address you used was aligned with the beginning of a page ?
On my H7 (with 128kB sectors) I can erase `0x08000000`, `0x08020000`, `0x08040000` and so on. Is `stlink_calculate_pagesize` supporting your flash size ?

Nightwalker-87: IMHO alignment is even a basic requirement for some memory regions according to common ARM reference guides (e.g. The Definitive Guide to ARM® Cortex®-M3 and Cortex®-M).
I found this as a direct reference as well: https://wiki.segger.com/Open_Flashloader#Example

antoinefaure: Yeah but that's what I'm doing here: I'm iterating through the pages to find the one we want to erase, and I check if the address is equal to the first address of the page.
Hence the question to @slyshykO regarding the parameters / chip he used, because I've tested this on my H7 (the only board I have at the moment) and I was able to erase starting from the 2nd, 3rd etc page.

antoinefaure: e.g. : 
```
./build/Release/bin/st-flash --connect-under-reset --reset --area=main erase 0x08060000 128k
st-flash 1.7.0-133-gf277fdb
2022-01-12T09:31:57 INFO common.c: H74x/H75x: 128 KiB SRAM, 2048 KiB flash in at least 128 KiB pages.
-> Flash page at 0x8060000 erased (size: 0x20000)
```

```
./build/Release/bin/st-flash --connect-under-reset --reset --area=main erase 0x08080000 256k
st-flash 1.7.0-133-gf277fdb
2022-01-12T09:32:17 INFO common.c: H74x/H75x: 128 KiB SRAM, 2048 KiB flash in at least 128 KiB pages.
-> Flash page at 0x8080000 erased (size: 0x20000)
-> Flash page at 0x80a0000 erased (size: 0x20000)
```

Ant-ON:
@antoinefaure In my opinion, this changes does not check the correctness of entering the starting address of the first page. I would add checks like these:
https://github.com/stlink-org/stlink/blob/7cc1fdae6b0c48f96f4f3155786305ac0cde0d87/src/common.c#L3472-L3514

It would also be nice to fix `stlink_write_flash` and add a call to `stlink_erase_flash_section` there instead of the erase code.

antoinefaure:
@Ant-ON Correct me if I'm wrong, but I think most of the checks you've linked are wrong:

- `} else if ((addr + len) < addr) { ` cannot happen as `len` is a `uint32_t` (or in my case a `size_t`). It can therefore never be `< 0`, so this test should always pass.
- `} else if (addr & 1) { ` doesn't check that an address is aligned with a page, it just checks if the last bit is a 0. e.g. `0x08000010 & 0x01 = 0x00` so the test will pass, but it is not aligned with a page. 
-  `} else if (len & 1) { ` same problem, it doesn't make sure `len` is aligned with a page
-  `} else if (addr & (sl->flash_pgsz - 1)) { ` unfortunately we've seen before that we can't use `sl->flash_pgsz` as some flash have different page sizes depending on the address (first few pages are 32kB, the next ones 128kB for example).

That leaves us with 2 tests : 
- `if (addr < sl->flash_base) { `
- ` } else if ((addr + len) > (sl->flash_base + sl->flash_size)) { `

Both of them are done here: 
https://github.com/stlink-org/stlink/pull/1213/files#diff-f6c5883a603e481644b9ef028b158acace9eabf59598f164dbec7b97d8da231fR2954-R2958

The alignment checks are done here : 
https://github.com/stlink-org/stlink/pull/1213/files#diff-f6c5883a603e481644b9ef028b158acace9eabf59598f164dbec7b97d8da231fR2963-R2969

And here : 
https://github.com/stlink-org/stlink/pull/1213/files#diff-f6c5883a603e481644b9ef028b158acace9eabf59598f164dbec7b97d8da231fR2963-R2969

Unless I missed something, I think we are all good with the checks.


> It would also be nice to fix `stlink_write_flash` and add a call to `stlink_erase_flash_section` there instead of the erase code.

That's a good catch, I'll fix this!

Ant-ON:
@antoinefaure 
- `} else if ((addr + len) < addr) {` happened in case of overflow. For example, when the address is `0xFFFFFF10` and the len is `0xFF`. Yes, an unobvious check and redundant.
- `} else if (addr & 1) { ` simple checks of address typo. I think this can be redundant.
- `} else if (len & 1) { ` this check is needed only for the write operation. Due to its limitations.
- `} else if (addr & (sl->flash_pgsz - 1)) { ` yes, may write better. This is the simplest test.

Sorry. At a cursory glance, I missed these checks. It is better to separate the checks so that the user understands the problem better. 
```c
  if (base_addr < sl->flash_base || base_addr >= (sl->flash_base + sl->flash_size)) {
    ELOG("Address is out of the flash (0x%08X - 0x%08X)\n", sl->flash_base, sl->flash_base + sl->flash_size);
    return (-1);
  }
  if ((base_addr + size) > (sl->flash_base + sl->flash_size)) {
    ELOG("The size exceeds the size of the flash (%u bytes available)\n", sl->flash_base + sl->flash_size - base_addr);
    return (-1);
  }
```
There is still some redundancy.
https://github.com/stlink-org/stlink/blob/7cc1fdae6b0c48f96f4f3155786305ac0cde0d87/src/common.c#L3472-L3499
This code can be shortened to (all checks done in the `stlink_erase_flash_section`):
```c
  if (len & 1) {
    WLOG("unaligned len 0x%x -- padding with zero\n", len);
    len += 1;
  }
```

src/common.c
Nightwalker-87: Ping @slyshykO

Nightwalker-87:
@antoinefaure @Ant-ON Thanks for the comments. Both looks good to me.
It would be indeed very helpful to have some of the explanations from above as inline comments as well.

antoinefaure:
Hey @Ant-ON @Nightwalker-87 
I've moved the checks to 2 separate functions, that way `stlink_write_flash` also benefits from the updated checks, and from a proper alignment test.

Testing this has been a bit frustrating, as after running `make install` I ended up with this issue : https://github.com/stlink-org/stlink/issues/1199.
But I've covered all the cases that came to mind : invalid address, invalid (address + size), address not aligned

Ant-ON (approved):

Ant-ON:
@antoinefaure Thank you! Everything looks very good!
Yes, the bug with configs has not been fixed yet...(

Nightwalker-87 (approved):

Nightwalker-87:
@antoinefaure Very good and valuable contribution. Thank you. :1st_place_medal: 
Technically everything seems to be addressed by now.
Yes, we're working on #1199 don't bother too much about it.
Now we are only waiting to hear @slyshykO and then this can go for a merge.

src/common.c
```diff
@@ -2951,30 +2951,81 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
   return check_flash_error(sl);
 }
 
+// Check if an address and size are within the flash
+int stlink_check_address_range_validity(stlink_t *sl, stm32_addr_t addr, size_t size) {
+  if (addr < sl->flash_base || addr >= (sl->flash_base + sl->flash_size)) {
+    ELOG("Invalid address, it should be within 0x%08x - 0x%08lx\n", sl->flash_base, (sl->flash_base + sl->flash_size -1));
```
slyshykO, line 2957: Please, add explicit conversation to `unsigned long` for `(sl->flash_base + sl->flash_size -1)` .

```c
(unsigned long)(sl->flash_base + sl->flash_size -1)
```

src/common.c
```diff
@@ -2951,30 +2951,81 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
...
+    return (-1);
+  }
+  if ((addr + size) > (sl->flash_base + sl->flash_size)) {
+    ELOG("The size exceeds the size of the flash (0x%08lx bytes available)\n", (sl->flash_base + sl->flash_size - addr));
```
slyshykO, line 2961: Please, add explicit conversation to `unsigned long` for `(sl->flash_base + sl->flash_size - addr)`

src/common.c
```diff
@@ -2951,30 +2951,81 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
...
+      ELOG("Invalid size (not aligned with a page). Page size at address %#x is %#lx\n", addr, page_size);
+      return -1;
+    }
+
+    if (stlink_erase_flash_page(sl, addr)) {
+      WLOG("Failed to erase_flash_page(%#x) == -1\n", addr);
+      return (-1);
+    }
+
+    fprintf(stdout, "-> Flash page at %#x erased (size: %#lx)\n", addr, page_size);
```
slyshykO, line 3009: Please, add explicit conversation to `unsigned long` for `page_size`

src/common.c
```diff
@@ -2951,30 +2951,81 @@ int stlink_erase_flash_page(stlink_t *sl, stm32_addr_t flashaddr) {
...
+    return -1;
+  }
+
+  stm32_addr_t addr = base_addr;
+  do {
+    size_t page_size = stlink_calculate_pagesize(sl, addr);
+
+    // Check if size is aligned with a page, unless we want to completely erase the last page
+    if ((addr + page_size) > (base_addr + size) && !align_size) {
+      ELOG("Invalid size (not aligned with a page). Page size at address %#x is %#lx\n", addr, page_size);
```
slyshykO, line 3000: Please, add explicit conversation to `unsigned long` for `page_size`

slyshykO (changes requested):

src/common.c
Nightwalker-87: Addressed by https://github.com/stlink-org/stlink/pull/1213#issuecomment-1008361996 and the following discussion.

src/common.c
Nightwalker-87: @slyshykO Does this work now for you with aligning?

Nightwalker-87:
@slyshykO No, we shouldn't continue with this practise of continued explicit conversions.
It should be fixed at the root of the problem and not in in the further course.

src/common.c
Nightwalker-87: see comment below

src/common.c
Nightwalker-87: see comment below

src/common.c
Nightwalker-87: see comment below

src/common.c
Nightwalker-87: see comment below

antoinefaure:
Hey guys, thanks for your help and for getting this merged in!
Cheers

slyshykO:
> @slyshykO No, we shouldn't continue with this practise of continued explicit conversions. It should be fixed at the root of the problem and not in in the further course.

The root of the problem is C-lang itself and its many realizations for numerous platforms. You will never guess which base type equal `ssize_t` unsigned or unsigned long. Wherein, C doesn't expand printf-like format.

Nightwalker-87:
A solution to this is to work with fixed-length typedefs like int16_t, int32_t, etc.
The described issue is one of the reasons why they were defined.

Have you tested [this](https://github.com/stlink-org/stlink/pull/1213#discussion_r785240746) in the meanwhile?
If not, please do so to allow for final closure of this PR.

slyshykO:
> A solution to this is to work with fixed-length typedefs like int16_t, int32_t, etc. The described issue is one of the reasons why they were defined.

Please, hear me. This is no solution at all when we deal with printf format lines cause its syntax to know nothing about sized types. 
Even more, the same uint32_t for GCC is unsigned long but for clang, it is just unsigned. So what format we should use %lx or %x ?

gszy:
> Please, hear me. This is no solution at all when we deal with printf format lines cause its syntax to know nothing about sized types. Even more, the same uint32_t for GCC is unsigned long but for clang, it is just unsigned. So what format we should use %lx or %x ?

`PRIu32`, defined in `inttypes.h`, see <https://en.cppreference.com/w/c/types/integer>.

Nightwalker-87:
Please open a new PR for this issue and reference the respective comment above.
This PR is now closed as the original issue has been addressed.

PR #1221, opened 2022-01-31
Set C standard through cmake variables

slyshykO:


CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
+set(CMAKE_C_STANDARD 11)
```
Nightwalker-87, line 12: Note: As soon as we reach `cmake` v3.21 as the minimum required version in the future, dependent on supported operating systems support we may set this to "17". For now we are fine with "11".

Nightwalker-87 (approved):

PR #1234, opened 2022-03-08
Optimisation for processing chipid files

Sylensky:
Hi everyone,

during my work i made some changes to the stlink library which i thought of would benefit the overall project.
These changes include optimizations in the processing of chipfiles and in a seperate PR the addition of option byte descriptions.

- [x] printing processed chipfiles only when it makes sense (not in st-info --version)
- [x] moved the dump chip printout from stderr to stdout
- [x] the chipfile read buffer was quite large for a rather small file and since its processed line per line it doesn't make sense to define the buffer that large.  (maybe i made it too small when someone adds a rather big description in the chipfile?)
- [x] sscanf already looks for spaces and seperates them so there was no need to explicitly look for spaces in my opinion 
- [x] the filedescriptor wasn't closed after the file has been processed. - fixed that

### DIFF
--- a/src/st-info/info.c
+++ b/src/st-info/info.c
@@ -29,7 +29,7 @@ static void stlink_print_version(stlink_t *sl) {
 }
 
 static void stlink_print_info(stlink_t *sl) {
-
+    const struct stlink_chipid_params *params = NULL;
     if (!sl) { return; }
 
     printf("  version:    "); stlink_print_version(sl);
@@ -37,6 +37,9 @@ static void stlink_print_info(stlink_t *sl) {
     printf("  flash:      %u (pagesize: %u)\n", (uint32_t)sl->flash_size, (uint32_t)sl->flash_pgsz);
     printf("  sram:       %u\n", (uint32_t)sl->sram_size);
     printf("  chipid:     0x%.3x\n", sl->chip_id);
+
+    params = stlink_chipid_get_params(sl->chip_id);
+    if (params) { printf("  dev-type:   %s\n", params->dev_type); }
 }
 
 static void stlink_probe(enum connect_type connect, int freq) {
--- a/src/st-info/info.c
+++ b/src/st-info/info.c
@@ -29,7 +29,6 @@ static void stlink_print_version(stlink_t *sl) {
 }
 
 static void stlink_print_info(stlink_t *sl) {
-    const struct stlink_chipid_params *params = NULL;
 
     if (!sl) { return; }
 
@@ -38,9 +37,6 @@ static void stlink_print_info(stlink_t *sl) {
     printf("  flash:      %u (pagesize: %u)\n", (uint32_t)sl->flash_size, (uint32_t)sl->flash_pgsz);
     printf("  sram:       %u\n", (uint32_t)sl->sram_size);
     printf("  chipid:     0x%.3x\n", sl->chip_id);
-
-    params = stlink_chipid_get_params(sl->chip_id);
-    if (params) { printf("  dev-type:   %s\n", params->dev_type); }
 }
 
 static void stlink_probe(enum connect_type connect, int freq) {
@@ -68,6 +64,8 @@ static int print_data(int ac, char **av) {
         printf("v%s\n", STLINK_VERSION);
         return(0);
     }
+
+    init_chipids(ETC_STLINK_DIR);
 
     for (int i=2; i<ac; i++) {
         
@@ -134,8 +132,6 @@ int main(int ac, char** av) {
         usage();
         return(-1);
     }
-
-    init_chipids (ETC_STLINK_DIR);
 
     err = print_data(ac, av);
 
--- a/src/stlink-lib/chipid.c
+++ b/src/stlink-lib/chipid.c
@@ -10,28 +10,28 @@
 
 static struct stlink_chipid_params *devicelist;
 
-void dump_a_chip (FILE *fp, struct stlink_chipid_params *dev) {
-    fprintf(fp, "# Device Type: %s\n", dev->dev_type);
-    fprintf(fp, "# Reference Manual: RM%s\n", dev->ref_manual_id);
-    fprintf(fp, "#\n");
-    fprintf(fp, "chip_id 0x%x\n", dev->chip_id);
-    fprintf(fp, "flash_type %d\n", dev->flash_type);
-    fprintf(fp, "flash_size_reg 0x%x\n", dev->flash_size_reg);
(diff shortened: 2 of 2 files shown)


src/stlink-lib/chipid.c
```diff
@@ -55,29 +55,23 @@ void process_chipfile(char *fname) {
...
-        for (p = buf; isspace (*p); p++);
+    while (fgets(buf, sizeof(buf), fp) != NULL) {
 
-        if (!*p) {
-            continue; // we hit end-of-line with only whitespace
-        }
-
-        if (*p == '#') {
+        if(strncmp(buf, "#", strlen("#")) == 0)
             continue; // ignore comments
```
gszy, line 61: I don’t like the redundant braces, but let’s follow the existing style:

```suggestion
        if (strncmp(buf, "#", strlen("#")) == 0) {
            continue; // ignore comments
        }
```

src/stlink-lib/chipid.c
```diff
@@ -55,29 +55,23 @@ void process_chipfile(char *fname) {
 
     ts = calloc(sizeof(struct stlink_chipid_params), 1);
 
-    while (fgets(buf, 1024, fp) != NULL) {
-        for (p = buf; isspace (*p); p++);
+    while (fgets(buf, sizeof(buf), fp) != NULL) {
 
-        if (!*p) {
-            continue; // we hit end-of-line with only whitespace
-        }
```
gszy, line 63: If there’s an empty line between two valid lines, wouldn’t we process the first one twice?

src/stlink-lib/chipid.c
```diff
@@ -30,8 +30,8 @@ struct stlink_chipid_params *stlink_chipid_get_params(uint32_t chip_id) {
     struct stlink_chipid_params *params = NULL;
     for (params = devicelist; params != NULL; params = params->next)
         if (params->chip_id == chip_id) {
-            fprintf(stderr, "\ndetected chip_id parametres\n\n");
-            dump_a_chip(stderr, params);
+            fprintf(stdout, "\ndetected chip_id parametres\n\n");
+            dump_a_chip(stdout, params);
```
gszy, line 34: :unicorn: IMHO think this function should not print this stuff at all. If it does, let’s keep it at the diagnostic output.

src/st-info/info.c
```diff
@@ -38,9 +37,6 @@ static void stlink_print_info(stlink_t *sl) {
     printf("  flash:      %u (pagesize: %u)\n", (uint32_t)sl->flash_size, (uint32_t)sl->flash_pgsz);
     printf("  sram:       %u\n", (uint32_t)sl->sram_size);
     printf("  chipid:     0x%.3x\n", sl->chip_id);
-
-    params = stlink_chipid_get_params(sl->chip_id);
-    if (params) { printf("  dev-type:   %s\n", params->dev_type); }
```
gszy, line 43: I wouldn’t delete this, see :unicorn: below.

gszy (commented):
Hi @Sylensky, thanks for your contribution. I’ve left a few comments in the changes.

src/stlink-lib/chipid.c
Sylensky: Just to clarify: You mean something like this?
```
# Chip-ID file for STM32F401xD/xE device
#
dev_type STM32F401xD_xE

ref_manual_id 0368
chip_id 0x433                // STM32_CHIPID_F4_DE
flash_type F2_F4
...
```

Sylensky: If that is the case you meant it processes the file properly after testing that case.

src/stlink-lib/chipid.c
Sylensky: My suggestions would be to move the `detected chip_id parametres` and the `dump_a_chip` to the debug log and keeping the below removed lines.

src/stlink-lib/chipid.c
gszy: Well, it seems it processes the properly, but it seems I was right that every blank line causes it to process the last non-blank line again. I think we it would be better to avoid such behavior.

Sylensky: What about an additional check if the line is empty? 3b328b26bd5eb0e07f78b8ba48e10c2581a1239a

gszy:
I’m sorry for the delay, I hope I’ll be able to review this next week.

Nightwalker-87:
Yes, there is nothing else I can report from my side actually... :pensive: 
@Sylensky Please stay tuned - it will remain on our ToDo list.

gszy:
Well, no luck, I won’t have enough time. Just one more thing I’ve spotted: the new small buffer size may cause issues on lines having long comments.

Sylensky:
I thought about that aswell. Making it double should totally be enough. Or do you see a case where such long comments might still be useful?

gszy:
I’m not sure about its usefulness, but FWIW here `F76x_F77x.chip` has quite a long line:
https://github.com/stlink-org/stlink/blob/8d96e769f32d1c85be913f77e6aab4b1d84b80b1/config/chips/F76x_F77x.chip#L12

Sylensky:
I checked the other files and the `F76x_F77x.chip` has the longest line so far. Choosing a buffer size of `256` should be in my opinion enough even if there is going to be more description on a line.

src/stlink-lib/chipid.c
Nightwalker-87: Yes, empty lines should be detected and simply ignored from my point of view.

Nightwalker-87: This point appears to remain unresolved yet...

src/st-info/info.c
Nightwalker-87: Yes, I agree - please keep these lines.

src/stlink-lib/chipid.c
Nightwalker-87: Yes, using the the debug log here does make sense to me.

src/stlink-lib/chipid.c
Sylensky: Do you mean even after 3b328b2? Can you point into the direction where i missed the point?

Nightwalker-87: No, that should be fine, as long as it finds it's way into this PR.

Nightwalker-87 (approved):

Nightwalker-87:
Waiting for @gszy now...

Nightwalker-87:
Ping @gszy

PR #1235, opened 2022-03-08
Added parametres option_base, option_size for F401xD_xE

Sylensky:
Hi everyone,

heres the PR i was talking about in #1234.
It simply adds the option byte base address for STM32F401VDT and other STM32F401xD devices.

### DIFF
--- a/config/chips/F401xD_xE.chip
+++ b/config/chips/F401xD_xE.chip
@@ -9,6 +9,6 @@ flash_pagesize 0x4000        // 16 KB
 sram_size 0x18000            // 96 KB
 bootrom_base 0x1fff0000
 bootrom_size 0x7800          // 30 KB
-option_base 0x0
-option_size 0x0
+option_base 0x40023C14
+option_size 0x4
 flags swo


config/chips/F401xD_xE.chip
```diff
@@ -9,6 +9,6 @@ flash_pagesize 0x4000        // 16 KB
 sram_size 0x18000            // 96 KB
 bootrom_base 0x1fff0000
 bootrom_size 0x7800          // 30 KB
-option_base 0x0
-option_size 0x0
+option_base 0x40023C14
+option_size 0x10
```
gszy, line 13: I must be missing something obvious, sorry for the noise. At 0x40023c14 there is the `FLASH_OPTCR` register, which is used for option byte access. But it seems this just a single register, so shouldn’t we have four bytes there?

Sylensky: I agree on the size. The `FLASH_OPTCR` register has 32 Bits where 24 are user programmable. So the overall size of 4 Byte is the correct one.

Sylensky:
Sorry for the additional force push. I accidently rebase this branch on the other pull request branch.

gszy (approved):

Nightwalker-87 (approved):

PR #1250, opened 2022-05-20
Added support for option bytes to F1xx_XLD (GD32F30x)

simon-wh:
This PR defines the `option_base` and `option_size` for F1xx_XLD which so far seems to be just used by GD32F30x chips which use the same option bytes configuration as STM32F1.

I also included the chip I've been using in `devices_boards.md`, it's pretty much the same as the existing GD32F303VGT6 chip mentioned there, mainly just a smaller package type.


Nightwalker-87:
We do not support non-STM devices officially.
Thus this change will only be approved after successful testing on respective STM32F1 boards.
We faced interoperability issues in this field within this project before.

Ant-ON (approved):

Nightwalker-87 (approved):

PR #1252, opened 2022-05-23
[st-trace] clock issues

simplerobot:
Fixed register read to use the right register for TPI_ACPR.
Updating to expect frequency units for clock and trace flags.
Making sure trace frequency is not less than 1/5 of the system frequency.

### DIFF
--- a/src/st-trace/trace.c
+++ b/src/st-trace/trace.c
@@ -98,11 +98,50 @@ static void usage(void) {
   puts("  -V, --version         Print this version");
   puts("  -vXX, --verbose=XX    Specify a specific verbosity level (0..99)");
   puts("  -v, --verbose         Specify a generally verbose logging");
-  puts("  -cXX, --clock=XX      Specify the core frequency in MHz");
-  puts("  -tXX, --trace=XX      Specify the trace frequency in Hz");
+  puts("  -cXX, --clock=XX      Specify the core frequency, optionally followed by");
+  puts("                        k=kHz, m=MHz, or g=GHz (eg. --clock=180m)");
+  puts("  -tXX, --trace=XX      Specify the trace frequency, optionally followed by");
+  puts("                        k=kHz, m=MHz, or g=GHz (eg. --trace=2m)");
   puts("  -n, --no-reset        Do not reset board on connection");
   puts("  -sXX, --serial=XX     Use a specific serial number");
   puts("  -f, --force           Ignore most initialization errors");
+}
+
+static bool parse_frequency(char* text, uint32_t* result)
+{
+  if (text == NULL) {
+    ELOG("Invalid frequency.\n");
+    return false;
+  }
+
+  char* suffix = text;
+  double value = strtod(text, &suffix);
+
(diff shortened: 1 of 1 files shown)


src/st-trace/trace.c
```diff
@@ -98,13 +98,53 @@ static void usage(void) {
...
+  if (value == 0.0) {
+    ELOG("Invalid frequency.\n");
+    return false;
+  }
+
+  for (int i = 0; suffix[i] != 0; i++)
+    suffix[i] = tolower(suffix[i]);
+
+  double scale = default_scale;
+  if (strcmp(suffix, "hz") == 0)
```
Nightwalker-87, line 127: One should use the correct upper/lower case orthography here: Hz, kHz, MHz and GHz

Nightwalker-87 (changes requested):
See inline comment.

src/st-trace/trace.c
simplerobot: Thank you for the review @Nightwalker-87.
Just to be clear, you expect st-trace to give an error if the user enters the units using the wrong case?

Nightwalker-87: No, not necessarily, but the implementation should interpret it correctly and ensure that e.g. "khz" and "kHz" is recognised as an input, but always displayed as "kHz" in any output wherever it may occur.

src/st-trace/trace.c
```diff
@@ -98,13 +98,53 @@ static void usage(void) {
   puts("  -V, --version         Print this version");
   puts("  -vXX, --verbose=XX    Specify a specific verbosity level (0..99)");
   puts("  -v, --verbose         Specify a generally verbose logging");
-  puts("  -cXX, --clock=XX      Specify the core frequency in MHz");
-  puts("  -tXX, --trace=XX      Specify the trace frequency in Hz");
+  puts("  -cXX, --clock=XX      Specify the core frequency");
```
Nightwalker-87, line 101: Another (more favourable) option would be to keep the unit Hz here and make solely k, M, G part of the input string.
Though not technically relevant here, one should keep in mind that "m" stands for "milli" and "M" for "mega" so looking at this it would give me a better feeling to follow the common convention here.

simplerobot: I prefer this too.  I'll make this change.  If the user does not include a unit, the system clock should default to Hz, but this would break backwards compatibility where it is assumed to be MHz.  What would you recommend in this case?

Nightwalker-87: There is not much backwards compatibility to care about. The st-trace tool is not in the codebase for very long.
Just don't bother about it - we'll add the change to the changelog and that should do. :wink:

simplerobot: Sounds good.  I'll have the fix for you tomorrow.

Nightwalker-87: Thanks, but take your time - we're not in a hurry. :wink:

Nightwalker-87 (approved):

PR #1258, opened 2022-07-11
increase option size to include write protection

Sylensky:
This is a rather odd pull request to support flashing of option byte binaries greater than 4 bytes for the G0Bx series.
Simply because there is currently no option to set/clear the write protection bits accordingly.
image

I would rather suggest what a proper solution might look like: An extra field in every `.chip` marking the area of the write protection bits and their sizes. If i would have enough time i would have proposed a solution for that but iam currently unable to do so.


Sylensky:
Sorry i have to revoke the PR because it doesn't work as intended.

PR #1259, opened 2022-07-14
[doc] Added `make install` to the macOS compiling instructions

jdbruijn:
This is required to get the `chips` directory in place. References #1237.

### DIFF
--- a/doc/compiling.md
+++ b/doc/compiling.md
@@ -120,12 +120,12 @@ or execute (Debian-based systems only): `apt-get install gcc build-essential cma
 1. Change into the project source directory: `cd stlink`
 2. Run `make clean` -- required by some linux variants.
 3. Run `make release` to create the _Release_ target
-4. Run `make install` to full install the package with complete system integration
+4. Run `make install` to full install the package with complete system integration. This might require sudo permissions.
 5. Run `make debug` to create the _Debug_ target (_optional_)<br />
    The debug target is only necessary in order to modify the sources and to run under a debugger.
 6. Run `make package`to build a Debian Package. The generated packages can be found in the subdirectory `./build/dist`.
 
-As an option you may also install to an individual user-defined folder e.g `$HOME` with `make install DESTDIR=$HOME`.
+> **Note** As an option you may also install to an individual user-defined folder e.g `$HOME` with `make install DESTDIR=$HOME`.
 
 #### Removal:
 
@@ -218,6 +218,8 @@ To do this with only one simple command, type:
 4. Run `make install` to full install the package with complete system integration. This might require sudo permissions.
 5. Run `make debug` to create the _Debug_ target (_optional_)<br />
    The debug target is only necessary in order to modify the sources and to run under a debugger.
+
+> **Note** As an option you may also install to an individual user-defined folder e.g `$HOME` with `make install DESTDIR=$HOME`.
 
 ## Build options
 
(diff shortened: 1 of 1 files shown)


doc/compiling.md
```diff
@@ -215,7 +215,8 @@ To do this with only one simple command, type:
 1. Change into the project source directory: `cd stlink`
 2. Run `make clean` to clean remnants of any previous builds.
 3. Run `make release` to create the _Release_ target
-4. Run `make debug` to create the _Debug_ target (_optional_)<br />
+4. Run `make install` to full install the package with complete system integration. This might require sudo permissions.
+5. Run `make debug` to create the _Debug_ target (_optional_)<br />
    The debug target is only necessary in order to modify the sources and to run under a debugger.
```
gszy, line 221: Consider adding the same comment as is under the Linux installation instructions:

```suggestion

As an option you may also install to an individual user-defined folder e.g `$HOME` with `make install DESTDIR=$HOME`.

```

doc/compiling.md
```diff
@@ -215,7 +215,8 @@ To do this with only one simple command, type:
 1. Change into the project source directory: `cd stlink`
 2. Run `make clean` to clean remnants of any previous builds.
 3. Run `make release` to create the _Release_ target
-4. Run `make debug` to create the _Debug_ target (_optional_)<br />
+4. Run `make install` to full install the package with complete system integration. This might require sudo permissions.
```
gszy, line 218: In case of Linux, superuser permissions may be needed as well. Consider adding this note there.

Perhaps the build instructions for the these operating systems could be merged?

gszy (approved):

doc/compiling.md
jdbruijn: Common requirements etc. are different for Linux and macOS, so I think merging the installation instructions section make things more difficult to follow/find. Did add the node for linux though!

doc/compiling.md
```diff
@@ -120,12 +120,12 @@ or execute (Debian-based systems only): `apt-get install gcc build-essential cma
...
 2. Run `make clean` -- required by some linux variants.
 3. Run `make release` to create the _Release_ target
-4. Run `make install` to full install the package with complete system integration
+4. Run `make install` to full install the package with complete system integration. This might require sudo permissions.
 5. Run `make debug` to create the _Debug_ target (_optional_)<br />
    The debug target is only necessary in order to modify the sources and to run under a debugger.
 6. Run `make package`to build a Debian Package. The generated packages can be found in the subdirectory `./build/dist`.
 
-As an option you may also install to an individual user-defined folder e.g `$HOME` with `make install DESTDIR=$HOME`.
+> **Note** As an option you may also install to an individual user-defined folder e.g `$HOME` with `make install DESTDIR=$HOME`.
```
gszy, line 128: I think we should not use blockquotes where asides would be more appropriate. (/me flees)

Nightwalker-87: FMPOV this additional note is not important enough to qualify for such highlighting.

doc/compiling.md
Nightwalker-87: I would not merge both either. The separation was also intended originally to guide users to the specific instruction they need for their type of system, in relation to the required setup, as @jdbruijn realised already.

gszy (approved):

Nightwalker-87 (approved):

Issue #1270, opened 2022-09-25
Need Fixing on Debian Packaging


reharish:
- [x] I made serious effort to avoid creating duplicate or nearly similar issue

In order to allow developers to isolate and target your respective issue, please take some time to select the check boxes below and fill out each of the following items appropriate to your specific problem.

- [x] Programmer/board type: v3(e.g STLINK /V1, /V2, /V2-onboard, /V2-clone, /V3)
- [x] Operating system an version: Linux (e.g Linux, macOS, Windows)
- [x] **stlink tools version** and/or git commit hash: c2db68e0df (e.g v1.6.1/git-d0416149)
- [x] stlink commandline tool name: stlink-gui (e.g `st-info`, `st-flash`, `st-trace`, `st-util`)
- [x] Target chip (and board, if applicable): nucleo with LoraWAN (e.g STM32F103C8T6 (NUCLEO-F103RB))

Further we kindly ask you to describe the detected problem as detailed as possible and to add debug output if available, by using the following template:

Commandline output:

```
Failed to load UI file: stlink-gui.ui

```

Expected/description:

* Expected to open stlink-gui

Fixed:

* Error shows there is no `stlink-gui.ui` file
* moved `stlink-gui` bin to `/usr/share/stlink/`
* then, executed  from the dir. It Works.

PR #1271, opened 2022-09-25
Fixed stlink-gui binary issue

reharish:
https://github.com/stlink-org/stlink/issues/1270
	modified:   cmake/packaging/deb/postinst


Nightwalker-87 (changes requested):
No, we shouldn't fix this outside of cmake which used to take care of this.

Issue #1276, opened 2022-11-04
[compilation] Correct path to stlink/chips subdirectory


letshin:
Hi - I am not sure if these are two separate issues but after compiling from this repo I get the following error with a E73-2G4M08S1EX (nRF52833) connected.
Commandline input:
```
st-info --probe
```
Commandline output:
```
/usr/share//stlink/chips: No such file or directory
Found 1 stlink programmers
  version:    V2J29S7
  serial:     50001000030000544334544E
  flash:      0 (pagesize: 0)
  sram:       0
  chipid:     0x000
```
Is it normal for "/usr/share//stlink/chips: No such file or directory" to appear?

Not sure if its related but while running PYOCD I get the following

Commandline input:
```
pyocd load s140_nrf52_7.2.0_softdevice.hex --target nrf52833
```

Commandline output:
```
0001139 C STLink error (9): Get IDCODE error [__main__]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/pyocd/__main__.py", line 161, in run
    status = cmd.invoke()
  File "/usr/local/lib/python3.8/dist-packages/pyocd/subcommands/load_cmd.py", line 96, in invoke
    with session:
  File "/usr/local/lib/python3.8/dist-packages/pyocd/core/session.py", line 391, in __enter__
    self.open()
  File "/usr/local/lib/python3.8/dist-packages/pyocd/core/session.py", line 529, in open
    self._board.init()
  File "/usr/local/lib/python3.8/dist-packages/pyocd/board/board.py", line 139, in init
    self.target.init()
  File "/usr/local/lib/python3.8/dist-packages/pyocd/core/soc_target.py", line 147, in init
    seq.invoke()
  File "/usr/local/lib/python3.8/dist-packages/pyocd/utility/sequencer.py", line 213, in invoke
    resultSequence.invoke()
  File "/usr/local/lib/python3.8/dist-packages/pyocd/utility/sequencer.py", line 208, in invoke
    resultSequence = call()
  File "/usr/local/lib/python3.8/dist-packages/pyocd/coresight/dap.py", line 439, in _connect
    probe_conn.connect(self._protocol)
  File "/usr/local/lib/python3.8/dist-packages/pyocd/coresight/dap.py", line 172, in connect
    self._connect_probe(protocol)
  File "/usr/local/lib/python3.8/dist-packages/pyocd/coresight/dap.py", line 189, in _connect_probe
    self._probe.connect(protocol)
[...]

PR #1279, opened 2022-11-14
[compilation] Corrected path to stlink/chips subdirectory

weiminshen99:
Fixes Issue #1276.


Nightwalker-87 (changes requested):
Please don't change the standard directory for installations on unix-based systems.
You may add an if-clause for identifying macOS systems instead.

weiminshen99:
Hi, I have added the if-clause for macOS as you suggested. Please review again. Thanks.

Nightwalker-87 (approved):

Issue #1272, opened 2022-09-28
STM32F407: st-trace can't reconnect


jared52005:
**Thank you for giving feedback to the stlink project.**

---

**NOTE: In order to offer sufficient and the best possible support, please read /CONTRIBUTING.md and follow the given instructions _before_ submitting a ticket.**

**Bug reports and/or feature requests will be deleted, if they violate our contribution guidelines and if no issue-template is used!** Thank you for your support.

---

- [x] I made serious effort to avoid creating duplicate or nearly similar issue

In order to allow developers to isolate and target your respective issue, please take some time to select the check boxes below and fill out each of the following items appropriate to your specific problem.

- [x] Programmer/board type: [STLINK V2J29S7]
- [x] Operating system an version: [Windows 11 21H2 22000.978]
- [x] **stlink tools version** and/or git commit hash: [v1.7.0 Release date: 2021-04-25]
- [x] stlink commandline tool name: [`st-trace`]
- [x] Target chip (and board, if applicable): [STM32F407VET6]

Further we kindly ask you to describe the detected problem as detailed as possible and to add debug output if available, by using the following template:

**Step 1:**  
Begin reading a trace and then end the application using Ctrl+C
```
C:\STLink Tests>st-trace.exe
libusb: info [winusb_get_device_list] The following device has no driver: 'USB\VID_8087&PID_0025\6&1143EA6&0&4'
libusb: info [winusb_get_device_list] libusb will not be able to access it
2022-09-28T17:42:11 INFO common.c: F4xx: 192 KiB SRAM, 1024 KiB flash in at least 16 KiB pages.
2022-09-28T17:42:11 INFO trace.c: Trace Port Interface configured to expect a 982 MHz system clock.
2022-09-28T17:42:11 INFO trace.c: Trace frequency set to 2000000 Hz.
2022-09-28T17:42:11 INFO trace.c: Reading Trace
Application is starting
^C
```
**Step 2:**  
 Reconnect. Sometimes `st-trace` is able to get through and it will reconnect correctly, usually I will end up with `LIBUSB_ERROR_TIMEOUT` error and then `LIBUSB_ERROR_PIPE` when I will pull USB cable out of ST-Link device
```
C:\STLink Tests>st-trace.exe
libusb: info [winusb_get_device_list] The following device has no driver: 'USB\VID_8087&PID_0025\6&1143EA6&0&4'
libusb: info [winusb_get_device_list] libusb will not be able to access it
[...]

PR #1292, opened 2023-01-02
Fixed st-trace reconnect on Windows

simplerobot:
This should fix issue #1272.

The windows signal handling was returning FALSE which caused the default handler to be called.  The default handler called ExitProcess before our process had a chance to cleanup.  While the code looks similar, st-util does not have the same issue since it actually cleans up its resources in the signal handler.


Nightwalker-87 (changes requested):
I'd suggest to fix this in `gdb-server.c` as well, as mentioned before, so we can include some testing there too. :wink:

Nightwalker-87 (approved):

Nightwalker-87:
Now awaiting testing by @jared52005.

PR #1293, opened 2023-01-04
Add new chip IDs for G0B0, G0B1, G0C1, G051, G061

ghent360:
New chip IDs fetched from STMProgrammer 2.12. RAM according to datasheet/rm0444.
Verified on G0B1.


Nightwalker-87 (changes requested):
@ghent360 We can't merge this in the current state.
You must work with the current `develop` branch as a base for your proposed changes.

PR #1306, opened 2023-04-22
Fixed unbounded write and check return values of sscanf

szsam:
Format string "%s" that does not control the length of data written may overflow.

### DIFF
--- a/src/st-util/gdb-server.c
+++ b/src/st-util/gdb-server.c
@@ -160,8 +160,10 @@ int parse_options(int argc, char** argv, st_state_t *st) {
 
             break;
         case 'p':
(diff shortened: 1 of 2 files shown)


Nightwalker-87 (approved):

gszy (approved):

Machiry:
This looks like a security vulnerability. Should there be a CVE associated with this?

Nightwalker-87:
@Machiry I'm not sure how relevant this is with respect to the use case of this software.
As we will not patch older versions due to compatibility issues in the codebase, it appears more reasonable to move forward towards a new release which includes this fix.

PR #1325, opened 2023-08-15
CMake: Avoid hard-wired /usr/local/share

gollux:
Instead of defining own CMAKE_INSTALL_SHAREDIR, use existing CMAKE_INSTALL_FULL_DATADIR.

This way, CMAKE_INSTALL_PREFIX is respected.

### DIFF
--- a/src/stlink-gui/CMakeLists.txt
+++ b/src/stlink-gui/CMakeLists.txt
@@ -16,19 +16,19 @@ if (NOT WIN32 AND NOT CMAKE_CROSSCOMPILING)
 
         # Install desktop application entry
         install(FILES stlink-gui.desktop
-                DESTINATION ${CMAKE_INSTALL_SHAREDIR}/applications)
+                DESTINATION ${CMAKE_INSTALL_FULL_DATADIR}/applications)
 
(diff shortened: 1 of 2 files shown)


src/stlink-gui/CMakeLists.txt
gszy, line 1: Please restore spaces-based indentation.

gszy (changes requested):
Looks OK except an indentation issue.

src/stlink-gui/CMakeLists.txt
gollux: Sorry. Fixed.

gszy (approved):

Nightwalker-87:
@gollux: Thanks for your contribution. The change looks fine to me.
However we should add a note in our tutorial on how to select a custom install path (e.g. link to https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) to point users to it, in case they are not familiar with CMake.

Nightwalker-87 (approved):

Nightwalker-87:
I'll update the documentation afterwards.

PR #1333, opened 2023-09-04
Fix printf format to print 0 correctly

fliiiix:
Otherwise 0 is printed as 0.

```
 // works
 printf("Command 0x%02x \n", 0x0);
 
 // broken for 0
 printf("Command %#02x \n", 0x0);
 printf("Command %#02x \n", 0x00);
 printf("Command %#02x \n", 0xFF);
```

https://godbolt.org/z/roGYqde3a


gszy (approved):
I’d reword the commit message to something less about fixing correctness and more about improving consistency; both `0` and `0x00` are correct. Perhaps something like "Prefix all CDB bytes with 0x, including zeros".

fliiiix:
fair enough updated

Nightwalker-87 (approved):

Issue #1352, opened 2023-11-07
[feature] OTP Read/Write


mirkomatonti:
- [x] I made serious effort to avoid creating duplicate or nearly similar issue
- [x] Programmer/board type: [STLINK V2]
- [x] Operating system an version: [Linux]
- [x] **stlink tools version** and/or git commit hash: [ v1.7.0]
- [x] stlink commandline tool name: [st-flash]
- [x] Target chip (and board, if applicable): [STM32L4*]

Hello,

I need to read and write the otp area for a project of mine. I noticed that this feature has not being implemented yet so I did a quick implementation and it works. I know that OTP area is not standard across the STM32 line so how would that work in case I want to do a PR?
Can I just add support to L4x? (which is pretty easy as the OTP area is treated as FLASH) or do I need to do an implementation for all the supported boards?

My implementation is something along the line:

`--area=otp read <output.bin> <len>` //Read from the OTP start address to len
`--area=otp write <input.bin> <add>` //Write from a specific OTP address

To keep compatibility with the other boards I basically added two new fields (otp_base and otp_size) to the .chips file which in case they are not set the OTP read/write request returns a Not Implemented error. I suppose this will be helpful in case someone in the future wants to extend support to other boards.

PR #1353, opened 2023-11-08
First Implementation of the OTP Read/Write function

mirkomatonti:
Hi,

I need to use the OTP area so I decided to provide an implementation for those boards that treat the OTP area as another FLASH region. 

Here is the basic structure of my implementation:
--area=otp read <output.bin> <len> //Read from the OTP start address to len
--area=otp write <input.bin> <add> //Write from a specific OTP address

Please note that I have only been able to test my implementation on the l452 and l4a6 boards as they are the only ones I have access to.

[Transcript](https://gist.github.com/mirkomatonti/a7a784d4d460f95e0cf2da4b5b31292d)

[Link to discussion](https://github.com/stlink-org/stlink/issues/1352)


mirkomatonti:
Forgot to include a file :(

Nightwalker-87 (approved):

Nightwalker-87:
@Ant-ON: Waiting for second review now.

src/stlink-lib/common_flash.c
```diff
@@ -1310,9 +1310,16 @@ int32_t stlink_fwrite_flash(stlink_t *sl, const char *path, stm32_addr_t addr) {
    * If the file is identified to be all-empty and four-bytes aligned,
    * still flash the whole file even if ignoring message is printed.
    */
-  err = stlink_write_flash(sl, addr, mf.base,
+
+  /* In case the address is within the OTP area we use a different flash method */
+  if(addr >= sl->otp_base && addr <= sl->otp_base + sl->otp_size) {
```
Ant-ON, line 1315: The condition should look like:
```c
if(addr >= sl->otp_base && addr < sl->otp_base + sl->otp_size) {
```

Nightwalker-87: indeed...

mirkomatonti: Yeah indeed. Will fix that tomorrow 👍

Nightwalker-87 (approved):

PR #1355, opened 2023-11-14
Fixed support for STM32U5 chips

marcelobarrosufu:
I have been using the stm32U535 and I was getting an error when trying to write or erase parts of the flash with the st-flash.exe program. Analyzing the code I noticed that:

1) There was only support for chip ID 0x482. I added the others with the appropriate flash/ram specs.

0x455: STM32U535/545
0x476: STM32U5Fx/5Gx
0x481: STM32U59x/5Ax
0x482: STM32U575/585

2) It was not possible to delete a page from the flash because the original test was based only on the STM32L5x2xx (with one or two banks and 2k pages) while the U5 family has 8k pages and two banks.

As this is my first contribution I'm not sure if the way I solved the problems was the best. I remain at your disposal for further clarification.

### DIFF
--- a/config/chips/U535_U545.chip
+++ b/config/chips/U535_U545.chip
@@ -11,4 +11,4 @@ bootrom_base 0x0bf90000
 bootrom_size 0x8000          // 32 KB
 option_base 0x0
 option_size 0x0
-flags none
+flags swo dualbank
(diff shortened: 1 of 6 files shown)


rcubee:
Shouldn't `flags none` inside the .chip files be set to `flags swo dualbank`?

marcelobarrosufu:
I didn't know that was one of the uses. If this is the recommendation, it might be interesting to make the code clearer. Any tips for filling out the rest of the file? I believe the option bytes region is missing too.

Nightwalker-87 (approved):

Ant-ON (approved):

Issue #1313, opened 2023-05-14
Stlink v3 MINI Debugger: Mac M2 ERROR usb.c: Failed to open USB device 0x0483:0x374e, libusb error: -99)


KoSv:
Hi there,

I hope getting my debugger up and running...

but I'm getting this error:

```
st-info --probe   
```

```
ERROR usb.c: Failed to open USB device 0x0483:0x374e, libusb error: -99)
```
-----

``` 
lsusb -vvv -d 0x0483:0x374e
```

gives me:

```
USB:

    USB 3.1 Bus:

      Host Controller Driver: AppleT8112USBXHCI

        STLINK-V3:

          Product ID: 0x374e
          Vendor ID: 0x0483  (STMicroelectronics)
          Version: 1.00
          Serial Number: 0017000A5553500F20393256
          Speed: Up to 480 Mb/s
          Manufacturer: STMicroelectronics
          Location ID: 0x01100000 / 2
          Current Available (mA): 500
          Extra Operating Current (mA): 0

    USB 3.1 Bus:

      Host Controller Driver: AppleT8112USBXHCI

    USB 3.1 Bus:

      Host Controller Driver: AppleT8112USBXHCI
```

MacOS Ventura v13.3.1(a)
MacBook Pro M2 14'' 2023
 
libs:
```
st-info v1.7.0

libusb v1.0.26
open-ocd v0.12.0
```

what can I do?

thanks and cheers!!

Konstantin

Nightwalker-87:
MacOS is no longer supported by this toolset.
See also Commit: 61ff09e5274d46a46ae58bc4ffe44fe90a887ea6

Issue #1296, opened 2023-01-12
Documentation bug? Installation on a Mac (M1)


timregan-cri:
In the main [README.md](https://github.com/stlink-org/stlink/blob/6bc70f776ab9f52bb9c11c57eeef399e33765850/README.md?plain=1#L87) it says

> MacOS: Users can open a terminal window and then follow the same procedure as for installing on Linux

But [the instructions for installing on Linux](https://github.com/stlink-org/stlink/blob/6bc70f776ab9f52bb9c11c57eeef399e33765850/README.md?plain=1#L72) start

> We recommend to install `stlink-tools` from the package repository of the used distribution

and then lists the linux distributions.

> - Debian Linux: [(Link)](https://github.com/stlink-org/stlink/releases)  
> - Ubuntu Linux: [(Link)](https://github.com/stlink-org/stlink/releases)
> - Arch Linux: [(Link)](https://www.archlinux.org/packages/community/x86_64/stlink)
> - Alpine Linux: [(Link)](https://pkgs.alpinelinux.org/packages?name=stlink)
> - Fedora: [(Link)](https://src.fedoraproject.org/rpms/stlink)
> - Gentoo Linux: [(Link)](https://packages.gentoo.org/packages/dev-embedded/stlink)

None of those look right for the Mac (M1). What is the correct way to install on a Mac, and should that be covered in the installation section of the README?

(N.B. there is a [brew package](https://formulae.brew.sh/formula/stlink), perhaps that is the preferred route?)

gszy:
See 0d96e362696c5b9be0f98ea3187bbfcca12dfd66 where the `README` was updated.

It [seems](https://github.com/Homebrew/homebrew-core/blob/master/Formula/stlink.rb) the Homebrew package follows upstream releases, so I think it should work for you. Note that version 1.7.0 is quite old and does not support for some newer MCUs.

Issue #1280, opened 2022-11-20
Build of stlink on macOS fails: 


fxcoudert:
- [X] I made serious effort to avoid creating duplicate or nearly similar issue

In order to allow developers to isolate and target your respective issue, please take some time to select the check boxes below and fill out each of the following items appropriate to your specific problem.

- [x] Programmer/board type: none, it's a compilation issue
- [x] Operating system an version: macOS 13 (Ventura)
- [x] **stlink tools version** and/or git commit hash: 1.7.0
- [x] stlink commandline tool name: no tool, it's a compilation issue
- [x] Target chip (and board, if applicable): none, it's a compilation issue

----

Compilation with `cmake` on macOS 13, using Xcode 14.1 (system compiler), leads to the following build failure:

```
Undefined symbols for architecture x86_64:
    "_SecTaskCopyValueForEntitlement", referenced from:
        _darwin_detach_kernel_driver in libusb-1.0.a(darwin_usb.o)
    "_SecTaskCreateFromSelf", referenced from:
        _darwin_detach_kernel_driver in libusb-1.0.a(darwin_usb.o)
```

These should be found in the Security framework, which should be linked with `-framework Security`.

It looks like this was seen before, but was specific (at the time) to macOS 10.15 https://github.com/stlink-org/stlink/issues/1269 . I can confirm that this behavior now occurs with macOS 13.

amiiit:
On the same combination: Mac OS 13.0.1 I get the following error:

```
CMake Error at cmake_install.cmake:69 (file):
  file INSTALL cannot copy file
  "/Users/amit/workspace/webrelay/stlink/build/Release/lib/libstlink.a" to
  "/usr/local/lib/libstlink.a": Permission denied.
```

I did then try with `sudo make install` and I am getting now the following error:

```
CMake Error at cmake_install.cmake:137 (file):
  file cannot create directory: /usr/share/stlink/chips.  Maybe need
  administrative privileges.
```

Any idea if this is okay for the installer to create a directory over there?

Nightwalker-87:
@fxcoudert Please see https://github.com/stlink-org/stlink/issues/1269#issuecomment-1288165069.
Due to several restrictions introduced by Apple since 10.15 support for macOS will be dropped completely with the next release. Please refer to the official STMCube-Programmer by ST-Microelectronics instead.

Issue #1269, opened 2022-09-14
Build fails on Mac OSX Catalina (10.15)


kgoba:
Build on Mac OSX 10.15 fails, possibly due to changes in the framework environment:
```
Undefined symbols for architecture x86_64:
  "_SecTaskCopyValueForEntitlement", referenced from:
      _darwin_detach_kernel_driver in libusb-1.0.a(darwin_usb.o)
  "_SecTaskCreateFromSelf", referenced from:
      _darwin_detach_kernel_driver in libusb-1.0.a(darwin_usb.o)
```

Looks like this can be solved by adding the Security framework to library link requirements in `CMakeFiles.txt`:
```
@@ -187,7 +187,8 @@ if (APPLE)     # ... with Apple macOS libraries
     find_library(ObjC objc)
     find_library(CoreFoundation CoreFoundation)
     find_library(IOKit IOKit)
-    target_link_libraries(${STLINK_LIB_SHARED} ${LIBUSB_LIBRARY} ${SSP_LIB} ${ObjC} ${CoreFoundation} ${IOKit})
+    find_library(Security Security)
+    target_link_libraries(${STLINK_LIB_SHARED} ${LIBUSB_LIBRARY} ${SSP_LIB} ${ObjC} ${CoreFoundation} ${IOKit} ${Security})
```
(and the same on lines 226/227)

Nightwalker-87:
Remaining support for macOS 10.15 will end with release `v1.7.1`.
Newer versions will not be supported due to the discontinuation of KEXTs by Apple.
Thus there will be no attempt to fix this.

PR #1373, opened 2024-02-17
fix stlink 1.8.0 osx build

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/161497


Nightwalker-87:
macOS is no longer supported by this toolset.
See #1269, #1280, #1296 and #1313 for further reference.

Nightwalker-87 (commented):
macOS is no longer supported by this toolset.
See #1269, #1280, #1296 and #1313 for further reference.

Issue #1294, opened 2023-01-06
STM32H7A3: $Verification of Flash Failed


defyingphysics:
Hello, 

I have been trying to use the st-flash to upload a binary, but I am getting `ERROR common.c: Verification of flash failed at offset: 6144`. 
If I use Eclipse IDE to build and program, take the built .elf, convert to to bin using `arm-none-eabi-objcopy ` and then try to flash it works. But if I make a new build and try to flash that "raw" it fails every time. 

- [x] I made serious effort to avoid creating duplicate or nearly similar issue
- [x] Programmer/board type: [STLINK-V3E ] 
- [x] Operating system an version: [Linux] 
- [x] **stlink tools version** and/or git commit hash: [v1.7.0]
- [x] stlink commandline tool name: [st-flash write]
- [x] Target chip (and board, if applicable): [NUCLEO-H7A3ZI-Q] 

Further we kindly ask you to describe the detected problem as detailed as possible and to add debug output if available, by using the following template:

Commandline output:
Here is the debug output from a project after it has been built using Cube IDE. 

`arm-none-eabi-objcopy -O binary blinky.elf blinky.bin`
`st-flash --debug write blinky.bin 0x8000000`
```
st-flash 1.7.0
2023-01-06T22:15:27 INFO usb.c: Unable to match requested speed 1800 kHz, using 1000 kHz
2023-01-06T22:15:27 INFO common.c: H7Ax/H7Bx: 128 KiB SRAM, 2048 KiB flash in at least 8 KiB pages.
file blinky.bin md5 checksum: 8643ac75929d77f7548faabfc45914ab, stlink checksum: 0x002011c3
2023-01-06T22:15:27 INFO common.c: Attempting to write 21356 (0x536c) bytes to stm32 address: 134217728 (0x8000000)
2023-01-06T22:15:27 INFO common.c: Flash page at addr: 0x08000000 erased
2023-01-06T22:15:27 INFO common.c: Flash page at addr: 0x08002000 erased
2023-01-06T22:15:27 INFO common.c: Flash page at addr: 0x08004000 erased
2023-01-06T22:15:27 INFO common.c: Finished erasing 3 pages of 8192 (0x2000) bytes
2023-01-06T22:15:27 INFO common.c: Starting Flash write for H7
21356/21356 bytes written
2023-01-06T22:15:27 INFO common.c: Starting verification of write complete
2023-01-06T22:15:27 ERROR common.c: Verification of flash failed at offset: 6144
[...]

Issue #1249, opened 2022-05-18
STM32H7A3ZI: Flash verification fails if UART/USART interface is enabled


EmanueleGiacomini:
- [x] Programmer/board type: [STLINK V3]
- [x] Operating system an version: [Linux]
- [x] **stlink tools version** and/or git commit hash: [v1.7.0-195-g951859c]
- [x] stlink commandline tool name: [st-flash]
- [x] Target chip (and board, if applicable): [STM32H7A3ZI (Nucleo-H7A3ZI-Q)] 

Further we kindly ask you to describe the detected problem as detailed as possible and to add debug output if available, by using the following template:

Commandline output:

```
$st-flash --reset write build/h7a3zi_test.bin 0x08000000

st-flash 1.7.0-195-g951859c
Failed to parse flash type or unrecognized flash type
2022-05-18T11:23:19 INFO common.c: STM32H7Ax_H7Bx: 128 KiB SRAM, 2048 KiB flash in at least 8 KiB pages.
file build/h7a3zi_test.bin md5 checksum: f56fd507ca1329599bac3afb7c645, stlink checksum: 0x00152999
2022-05-18T11:23:19 INFO common_flash.c: Attempting to write 13832 (0x3608) bytes to stm32 address: 134217728 (0x8000000)
-> Flash page at 0x8000000 erased (size: 0x2000)
-> Flash page at 0x8002000 erased (size: 0x2000)

2022-05-18T11:23:19 INFO flashloader.c: Starting Flash write for H7
13832/13832 bytes written
2022-05-18T11:23:19 INFO common_flash.c: Starting verification of write complete
2022-05-18T11:23:19 ERROR common_flash.c: Verification of flash failed at offset: 6144
stlink_fwrite_flash() == -1

```

Expected/description:
While testing the board, we noticed that st-flash fails during the verification phase if a UART/USART interface is enabled through STM32CubeMX.
By flashing the same build with STM32CubeProgrammer, no issues appears during the write and the verification phases.

The code is fully generated by STM32CubeMX with a small addition by ours to transmit a sample text on USART3 every 100 ms 

We replicated this behavior on two different machines (Pop!_OS 20.04 on Intel i7-9850H and Ubuntu 20.04.4 on Intel i7-7700HQ)

Ant-ON:
@EmanueleGiacomini  Can you try use connect under reset?
`st-flash --connect-under-reset --reset write build/h7a3zi_test.bin 0x08000000`

[...]

PR #1397, opened 2024-05-14
[feature] --mass-erase for st-flash write commands

### LINKED ISSUE #1366: [feature] Caching and diffing bin/hex
In most of the cases when you are programming you only change small pieces of code. It means that the bin/hex is changed only  slightly based on the previous one. When you program the device, all blocks are programmed again. It means that even when you change a small string value, all 40 or 50 memory blocks are re-programmed again and you have to wait all the times for 40sec because the max programming speed is 4Mbps. Should it be possible to check which blocks have been changed based on the previous bin/hex and only program those which are different? It could speedup the development cycle a lot. It could be an option to provide --non-caching or something like that to reprogram the whole flash.

hannesweisbach:
Hello,

I have an STM32H7 where flash verification fails (probably somehow related to #1249 and/or #1294). #1238 indicates that a `--mass-erase`-flag for `st-flash write` would be welcome.

Open points:

* ~I use `_Bool` to store flags, instead of `int32_t`, which conveys better the meaning of the variable. However, I have no issue changing to `int32_t` or something similar to keep the codebase uniform.~ Changed to int32_t.
* ~Should passing `--mass-erase` to `read`, `erase`, and `reset` commands issue an error? As it has no meaning nor impact on those commands, it is currently silently ignored.~ `stlink --mass-erase erase` implemented.
* ~This PR is against the `develop` branch, as indicated by the Contributing Guidelines. Is this correct?~ Base branch changed to `testing`.

Best regards,
Hannes

(Closes #1238) (Closes #1366)


Ant-ON:
@hannesweisbach

My thoughts on open points:
* it's better to use int32_t or enum (to be more informative) instead _Bool.
* It may not be processed. But it would be nice to add a new masserase command. Now, to do this, you need to call erase command with zero address or zero size, which is not obvious.

In general, I think it’s not worth adding mass_erase field to struct _stlink. It would be nice to add a erase type parameter to the stlink_write_flash and stlink_mwrite_flash functions.

hannesweisbach:
I changed the the type of the `mass_erase` member in `flash_opts` to `int32_t`. I also introduced an `enum erase_type_t` in `flash_common.h`, which now gets passed to `stlink_fwrite_flash`, `stlink_mwrite_flash`, and `stlink_write_flash`.

I think it's a bit unclean in the regard that calling `stlink_fwrite_flash` for OTP-programming ignores this argument alltogether, but oh well, you can't win them all ;)

With the introduction of the `erase_type-flag, I also removed the `mass_erase` field from `struct stlink`.

Regarding the `masserase` command: I found `stlink erase` quite intuitive.

Do you imagine a command like: `stlink mass_erase` (i.e. completely new command), or `stlink --mass-erase erase [...]` (i.e. if the flag is given, the erase command ignores additional arguments and performs a mass erase)?

Ant-ON:
@hannesweisbach 
Hmm, you're right. With OTP calls it really looks a little illogical. We can replace `MASS_ERASE` to something like `NO_ERASE`, then everything will look more logical.
I imagined `stlink --mass-erase erase` without address and size.

hannesweisbach:
I changed the `erase` command such that the `--mass-erase` is observed. I added it as separate commit only for review reasons. I'll squash the commit before merging.

I have NO_ERASE yet to implement.

Nightwalker-87:
@hannesweisbach Thank you very much for the update. There is no need to squash the commits. An upcoming merge will add all separate changes at once and basically has the same effect seen from the perspective of the destination branch.

hannesweisbach:
Yes, I know. I'm always unhappy with the resulting commit message though. Atleast on GitLab. I haven't done merge squashes with the UI on Github.

Ant-ON (approved):

Nightwalker-87 (approved):

hannesweisbach:
Sorry for the delay. I squashed the commits and cleaned the commit message up (no code changes).

Nightwalker-87:
> Yes, I know. I'm always unhappy with the resulting commit message though. Atleast on GitLab. I haven't done merge squashes with the UI on Github.

@hannesweisbach: I've noticed that you force pushed in order to squash. It shall be ok here, but please take into account  for future contributions that this is not considered a good practise for technical reasons: https://docs.github.com/de/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests.

PR #1411, opened 2024-07-08
[feature] Improvements for stlink-gui

hyves42:
Hello,
I'm using stlink-gui a lot these days and decided I could help by fixing a few things that I miss in the UI. 

This PR contains 4 commits with 4 features independent from each other:
* [When opening hex files, start dialog from the last used location](https://github.com/stlink-org/stlink/commit/7b55ed933efd762ae76a393603293b155e0bc744) By default the `Open file` dialog always starts from the home directory. With this commit, if a file has been open since program start, the dialog will start from that directory. 
* [In 'Flash device' dialog, put focus on the OK button](https://github.com/stlink-org/stlink/commit/1405e58059609e0b7cd295cf9613b6cbcc6a25ac) When flashing, we always have to move the mouse to the `OK` button to start programming. With this commit we only have to press `space` or `return`
* [Add a 'Reset' button to gui](https://github.com/stlink-org/stlink/commit/3d5b49b042d3e9017471d4a802f5d67ce7e76dbb) Add a button to the right of `Flash` button to reset the target board
* [Add command line arguments to open a file at startup](https://github.com/stlink-org/stlink/commit/896a77a696130caf303ed0f3fc114b8a859cff4a) Add support for common command line arguments like `--help`,  `--version` and especially to load a .hex file at startup. Also update the .desktop file to let nautilus open .hex files with `stlink-gui`

If you prefer, I can split this into 4 separate PRs.
Thanks for your feedback

### DIFF
--- a/src/stlink-gui/gui.c
+++ b/src/stlink-gui/gui.c
@@ -50,6 +50,17 @@ static void stlink_gui_init(STlinkGUI *self) {
     self->file_mem.base   = 0;
 }
 
+static void help(void)
+{
+    puts("usage: stlink-gui [options] file\n");
+    puts("options:");
+    puts("  --version/-v           Print version information.");
+    puts("  --help/-h              Show this help.");
+    puts("");
+    puts("examples:");
+    puts("  stlink-gui path/to/file");
+}
+
 static gboolean set_info_error_message_idle(STlinkGUI *gui) {
     if (gui->error_message != NULL) {
         gchar *markup;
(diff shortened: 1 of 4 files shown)


Nightwalker-87:
> If you prefer, I can split this into 4 separate PRs.

Hi. Thanks for contributing. Splitting is neither necessary not desired. There is one common topic, so we shall be fine.
I'll have a look at this PR as soon as I find some time.

Nightwalker-87 (approved):

PR #1431, opened 2024-09-30
[feature] Prevent hardcoded install

### LINKED ISSUE #1430: [feature] Prevent installing of hardcoded modprobe config and udev rules
When working with isolated build environments (docker container or local user builds) under Linux it is not usable if the cmake build forces an installation to install files to hardcoded /etc/{modprobe.d,udev} locations. A option like "PREVENT_HARDCODED_INSTALL" would be desirable.

e.g.:
make -j1 CMAKEFLAGS="-DPREVENT_HARDCODED_INSTALL=1" release install

See PR #1431

sansyse:
- Introduced an environment variable that prevents the CMake installation script from installing modprobe config and udev rules to /etc/.

(Closes #1430)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -348,7 +348,7 @@ install(TARGETS st-trace DESTINATION ${CMAKE_INSTALL_BINDIR})
 # Device configuration (Linux only)
 ###
 
-if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND NOT PREVENT_HARDCODED_INSTALL)
     ## Install modprobe.d conf files to /etc/modprobe.d/ (explicitly hardcoded)
     set(STLINK_MODPROBED_DIR "/etc/modprobe.d" CACHE PATH "modprobe.d directory")
     install(FILES ${CMAKE_SOURCE_DIR}/config/modprobe.d/stlink_v1.conf DESTINATION ${STLINK_MODPROBED_DIR})


CMakeLists.txt
```diff
@@ -348,7 +348,7 @@ install(TARGETS st-trace DESTINATION ${CMAKE_INSTALL_BINDIR})
 # Device configuration (Linux only)
 ###
 
-if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND NOT PREVENT_HARDCODED_INSTALL)
```
Nightwalker-87, line 351: It appears that this is a self-defined `cmake` command. It needs to be documented before having it added to the code. Also this topic should be discussed in general.

Nightwalker-87 (changes requested):

CMakeLists.txt
sansyse: It's just an environment variable that, if defined non-null, prevents this install. It works.

In my opinion, it is not a good practice, if a build process of a development tool needs root privileges and forces modifications on the host system. Due to security reason, I never ever run such a build process with root privileges, in particular if you have not reverse-engineered the installation process and know exactly which modifications are done. I'm working on several different projects on the same development host and need to track modification manually. On the other hand a good practice is, if a project can installed relatively for packatizing, archiving, in-docker installation and/or project-local use. The modification on the host system is only done once (fully intentional), but the build process and debugging is done regularly and should never be done with root privileges.
In my environment the stlink stuff are archived in the project together with sources and other resources and the build and installation process of stlink tools are part of the project build. It is installed locally only. 

STLINK_INSTPREFIX       = $(abspath host)
STLINK_CMAKEFLAGS       = -DCMAKE_INSTALL_FULL_DATADIR=$(STLINK_INSTPREFIX) -DCMAKE_INSTALL_PREFIX=$(STLINK_INSTPREFIX) -DPREVENT_HARDCODED_INSTALL=1

... make -j1 CMAKEFLAGS="$(STLINK_CMAKEFLAGS)" release install

I agree, that a hobbyist or standard-user may do not need such a feature and for him it would be safer to install everthing in one automated installation process. But there are also experienced users with non-standard use cases.
A compromise and maybe better solution would be, to use the sysconfig_dir feature in you installation script to use it like this:

STLINK_INSTPREFIX       = $(abspath host)
STLINK_CMAKEFLAGS       = -DCMAKE_INSTALL_SYSCONFDIR=$(STLINK_INSTPREFIX)/etc -DCMAKE_INSTALL_FULL_DATADIR=$(STLINK_INSTPREFIX) -DCMAKE_INSTALL_PREFIX=$(STLINK_INSTPREFIX)

... make -j1 CMAKEFLAGS="$(STLINK_CMAKEFLAGS)" release install

Nightwalker-87: Based on a very similar argumentation, I would not mind either, if a full install would run without root privileges, but in practise we find that this is not possible on linux (and also on windows) for every step of the process due to predefined system default RWX directory permissions.

Nightwalker-87:
@Ant-ON: What's your opinion on this approach?

Ant-ON:
@Nightwalker-87 I think this modification is unnecessary. An experienced user who needs a custom installation can manually place executable and configuration files. There's not much that needs to be done by hand. This modification raises more questions than it provides in functionality.

Issue #1399, opened 2024-05-24
STLINK/V3-SET: If the device fails to enter SWD mode, the usb device is not released


TheFissk:
- [X] I made serious effort to avoid creating duplicate or nearly similar issue

---

- [X] Programmer/board type: STLINK/V3-SET
- [X] Operating system an version: Linux (Ubuntu 22.04 LTS + a really out of date version of Debian)
- [X] **stlink tools version** and/or git commit hash: v1.8.0
- [X] stlink commandline tool name: Tested against st-flash and st-info

---
When there is no board connected to the programmer causing it to fail to enter SWD mode subsequent invokations of the tool hang before failing.

Commandline output:
These commands are separated by several seconds, I'm not machinegunning the tool expecting it to work. The first command returns immediately, the second hangs for ~10s
``` 
$ st-flash reset
st-flash 1.8.0
Failed to enter SWD mode
Failed to connect to target
Failed to parse flash type or unrecognized flash type //This line went to stderr rather than stdout
$ st-flash reset
st-flash 1.8.0
2024-05-24T10:41:08 ERROR usb.c: GET_VERSION send request failed: LIBUSB_ERROR_TIMEOUT
2024-05-24T10:41:11 ERROR usb.c: GET_CURRENT_MODE send request failed: LIBUSB_ERROR_TIMEOUT
2024-05-24T10:41:14 ERROR usb.c: GET_COM_FREQ send request failed: LIBUSB_ERROR_TIMEOUT
2024-05-24T10:41:17 ERROR usb.c: GET_CURRENT_MODE send request failed: LIBUSB_ERROR_TIMEOUT
2024-05-24T10:41:20 ERROR usb.c: ENTER_SWD send request failed: LIBUSB_ERROR_TIMEOUT
Failed to parse flash type or unrecognized flash type
Failed to enter SWD mode
Failed to connect to target
2024-05-24T10:41:23 ERROR usb.c: DEBUG_EXIT send request failed: LIBUSB_ERROR_TIMEOUT
```

I also tested it against st-info --chipid and it works in much the same way, but it suppresses the stderr output, so that the program hangs for a ~10 seconds before outputting:
```
$ st-info --chipid
Failed to enter SWD mode
0x0000
````

Plugging a board into the programmer once its entered this degraded state does not resolve the issue, it will continue to output the same errors

`st-info --probe` works as expected returning instantly regardless of whether or not the device was able to enter SWD mode in previous invokations of `st-info --probe`

[...]

PR #1437, opened 2024-10-21
[withdrawn] New logging system in preparation for an intended stlink library approach

a-michelis:
> **This Pull Request is purely for overview purposes.**

This PR attempts to give STLink project a proper logging system in preparation for a library approach, to enable users incorporate stlink-lib into their projects. 

## Proposal: Main changes (in a nutshell)

- The project now uses [libusb-cmake](https://github.com/libusb/libusb-cmake) project as its _libusb_ provider, which in turn allows for custom-built _libusb_, tailored to each system's needs.
- All included headers have been placed into appropriate `stlink` _sub-folders_ that are included into the project as system includes, to allow for system-like includes. This change required the change of all inclusions in the project to change from `#include "%HEADER%.h"` to `#include <stlink/%HEADER%.h>`
- Several headers have exposed functionality which could be used by custom software, using `extern "C"` blocks.

## Much needed prior additions that are missing (this PR):
1) Changing the logging infrastructure to allow programmatically managed logging, instead of STDOUT logs.
2) Help to accommodate the appropriate cmake modifications for other BSD/*NIX based systems, as well as verify/implement a proper handling of pre-installed `libusb` followed by exhaustive testing
3) Documentation of the exported functions (and of course evaluate whether the overall exposure is excessive or too limited).

A.M.
(updated by Nightwalker-87)


Nightwalker-87:
There are a few header dependencies that have not been converted to the new scheme. Thus, as it appears, this results in compilation errors. However I believe this can be easily fixed. I'll continue reviewing afterwards. Yet there are quite a few attempts and improvements I consider very useful.

Nightwalker-87:
I'd prefer to keep source and header files together in the same directory respectively, unless there are superior header files without a matching source file, but this should be rather a stylistic issue than a technical one...

a-michelis:
> I'd prefer to keep source and header files together in the same directory respectively, unless there are superior header files without a matching source file, but this should be rather a stylistic issue than a technical one...

I tend to divide the big projects into separate, smaller, target-centric sub-projects, included into the root `CMakeLists.txt` via `add_subdirectory(${PROJECT_SOURCE_DIR}/src/target_name)`. This usually helps with projects like `stlink` that generate a toolset rather than one entity -allows for separation of concerns and enables two developers work in two different tools at the same time while avoiding conflicts.

Additionally, I usually separate headers and source files (despite being harder to manage) because it makes it easier to prefix different versions differently if ever needed, or avoid name conflicts, like with Windows and existing `libusb.h`. 
AFAIK, there is no mechanism in CMAKE that allows for "virtual prefixing" of headers, which in turn makes using `<stlink/header.h>` rather than `"header.h"` in our sources and headers impossible.

That said, the project's structure is already in place and I'm in no position to enforce such changes, unless they're wanted -in which case I can pay closer attention on what goes where, how and why. :)

---

Seems like separating sources and headers makes it harder for static analyzers such as `cppcheck` to properly analyze the project, which is also something to consider in my approach.

a-michelis:
Fixing the non-prefixed `usb.h` made both workflows succeed - i shall remove my branch from the workflows, tho, since it will mess up the main project.

That said, I also added windows (MSVC) on the workflow and that passed as well!

Nightwalker-87:
> Fixing the non-prefixed `usb.h` made both workflows succeed - i shall remove my branch from the workflows, tho, since it will mess up the main project.
> 
> That said, I also added windows (MSVC) on the workflow and that passed as well!

No, it won't mess up the project in any kind, as long as the merge of a test-failing PR does not take place. Indeed it is even better to have them here to detect possible issues before merging into the `testing` branch.

Nightwalker-87:
Further I've converted this PR into a draft as long as proposed changes are still in discussion.
Remember: Haste makes waste - we are not in a rush. Approaches and ideas should be well considered by taking into account relevant aspects. This does not imply that there is no view to improvements at all.

Nightwalker-87:
Any update here?

a-michelis:
Hello again, 

Unfortunatelly IRL is being crazy for the past month. 

Let's formulate a plan as to what needs to be implemented and at which order, so I can allocate time implementing, instead of theorizing about potential features!

**To my experience, making libstlink a standalone library requires:**

1) a bulletproof method of linking to either existing or downloaded/compiled LIBSTLINK. Currently, this is not present for Windows/MSVC. 

2) complete separation of the library source/headers and the rest of the toolset (as in subproject, not necessarily a dedicated repo)

3) a bulletproof method of providing bindings to the library (ie `FindStlink.cmake`, pkgconfig etc)

**About these points:**

(1) even though we have a provider for Windows/MSVC LIBSTLINK, it seems to require some manual labor which we shall take into account:
- Having the LIBUSB downloaded/built for Win/MSVC requires from STLINK to "install" it together with its own binaris -which counts as clear redistribution and I'm not aware of its legal implications.
- Requiring it to be pre-installed might be a solution, but it limits the libraries portability -which i'm not aware whether it's a wanted quality.

(2) To add to this single-responsibility fashion, I'd suggest each tool to be contained within its own subproject for better maintainability and proper linking.

(3) This is also related to note (1), and requires some research and further testing on how this library is supposed to be used. If portability is on the table, the methodology must also contain the discovery logic (such as `StlinkConfig.cmake` ). If not, a simple `FindStlink.cmake` with some options available about the requested library type could be suffice. 

All these said, I can only lay down some ideas and contribute to their implementation to the best of my ability! I wish you find this textwall usefull and it allows you to cherrypick the next actions! 

Cheers,
A.M.

Nightwalker-87:
> Hello again,
> 
> Unfortunatelly IRL is being crazy for the past month.
> 
> Let's formulate a plan as to what needs to be implemented and at which order, so I can allocate time implementing, instead of theorizing about potential features!
> 
> **To my experience, making libstlink a standalone library requires:**
> 
>     1. a bulletproof method of linking to either existing or downloaded/compiled LIBSTLINK. Currently, this is not present for Windows/MSVC.
> 
>     2. complete separation of the library source/headers and the rest of the toolset (as in subproject, not necessarily a dedicated repo)
> 
>     3. a bulletproof method of providing bindings to the library (ie `FindStlink.cmake`, pkgconfig etc)
> 
> 
> **About these points:**
> 
> (1) even though we have a provider for Windows/MSVC LIBSTLINK, it seems to require some manual labor which we shall take into account:
> 
>     * Having the LIBUSB downloaded/built for Win/MSVC requires from STLINK to "install" it together with its own binaris -which counts as clear redistribution and I'm not aware of its legal implications.
> 
>     * Requiring it to be pre-installed might be a solution, but it limits the libraries portability -which i'm not aware whether it's a wanted quality.
> 

I can't spot a problem here, as `libusb` is an open project as well - I mean it's a library _to be used_ with other software, so what's the point about it? The only thing one should take into account FMPOV is that redistribution occurs under the same licensing terms of the original project for the redistributed part.

> 
> (2) To add to this single-responsibility fashion, I'd suggest each tool to be contained within its own subproject for better maintainability and proper linking.
> 

A separate subdirectory structure should do. As this solely applies to Windows anyway, things should go to `/stlink/src/win32/` which appears to be the appropriate place.

> (3) This is also related to note (1), and requires some research and further testing on how this library is supposed to be used. If portability is on the table, the methodology must also contain the discovery logic (such as `StlinkConfig.cmake` ). If not, a simple `FindStlink.cmake` with some options available about the requested library type could be suffice.
> 

The latter seems to be the most sufficient approach.

> All these said, I can only lay down some ideas and contribute to their implementation to the best of my ability! I wish you find this textwall usefull and it allows you to cherrypick the next actions!
> 

I consider it a good approach to start with the PR you recently opened, tailoring it down to Windows-specific changes in the context of the given comments from the review today.
In a second step I'd like to address the cross-building capability under the changed circumstances.

After that it may turn out to useful to strip this Draft-PR down to the remaining changes in order to continue the review.
I'd not change the structure of the project yet, as I see this as a separate, independent issue, which may follow in a third step.

Kind regards
Nightwalker-87

Nightwalker-87:
@a-michelis: It might be a good idea to do some refactoring related to STLINK commands and related code on STLINK programmer APIs (see #1399) in the sense of a library approach, before continuing with more general restructuring of the code base. As you may have already noticed, I pushed this Draft-PR to the v1.8.2 milestone in the meanwhile.

a-michelis:
> @a-michelis: It might be a good idea to do some refactoring related to STLINK commands and related code on STLINK programmer APIs (see #1399) in the sense of a library approach, before continuing with more general restructuring of the code base. As you may have already noticed, I pushed this Draft-PR to the v1.8.2 milestone in the meanwhile.

I had the same concept in my mind, initiated from a different prespective:

I started playing around with the logging facility of the project. Specifically I'm trying to create the logging interface I was talking about in https://github.com/stlink-org/stlink/pull/1437#issue-2602369275 . The approach I'm working on is:

1. 4 Distinct log types:
   - ERROR
   - WARNING
   - INFO
   - DEBUG
   
   Debug category is a _phony_ for Tracing. 10 individual trace levels will be used, from 0 to 9.
   ```c
   enum log_type {
       LT_NONE = 0, // default value, no log
       LT_ERR = 1,
       LT_WRN = 2,
       LT_INF = 3,
       LT_DBG = 4, // this is a trace. tracelevel = value - 4, valid tracelevel values are 0-9. everything above that will be clipped and handled as trace lvl 9
       // anything above that is clipped to 13 and handled as level 9 trace
   };
   ```
2. Log and trace level is defined programmatically (and can be exported in flags): 
   - One can define the max logging level as before
   - If `LT_DEBUG` is selected, user can also select the Debug logs verbosity (trace level) from 0 to 9.
   
3. User of the library can:
    - disable logging altogheter or (re)enable it at any given point. (`stlink_nolog()` / `stlink_yeslog()`)
    - define a custom log handling function (ie for when making a GUI tool and want to export the logs to the tool, or when they want to stream the logs within a trace facility via sockets or something) 
    - change the log handling function at any given point in execution time (`stlink_loghandler_custom(void (*handler)(uint8_t lvl, const char* sender, const char* msg))`)
   
   Logging disabling can be also usefull to the tools, for a convinient `-q` / `--quiet` flag.
   Even when using custom log handling, the function will receive only the logs qualified by the loglevel and tracelevel filters.

4. There are three in-house (aka default) log handling functions.
   - STDOUT logging (as before), with format:
      ```
      [LOGTYPE] TIMESTAMP | SENDER_FUNCTION | MESSAGE
      ```
   
   - Color-enabled STDOUT logging: same format as simple stdout, but with different colored trace levels (ADHD fellas like me will find it way easier to read)
     
   - CSV logging : Exports the logs to a CSV file with `logtype`, `timestamp`, `sender` and `message` columns

   All of these can be selected by using flags (`--color`, `--csv $FILEPATH`) in the tools, or programmatically, by using `stlink_loghandler_default()` / `stlink_loghandler_default_color()` `stlink_loghandler_default_csv(const char* filepath)`. Any of the handler changing functions will first check if a csv file is opened by a previously used CSV handler and close it.
   
   The log type codes will be `ERR`, `WRN`, `INF` and `TR0`-`TR9`
### Reasoning

I strongly believe that having a robust logging system will allow us to better understand an issue, while letting us write a more structured code - and thus maintain it easier (_want to understand what a function does? Simply read its `log`/`trace` commands_). Also, having a documented train of thought via the log/trace entries, will facilitate API documentation writeups later on, when the Library approach comes to life.

All that said, If you find a part in the codebase that needs immediate care, please share proposed approaches/changes so i can start with them instead!

### Off topic

On another note, today is my bday 🎉 , so my presence will be somewhat limited -I'll start working on it the lib more intensively the following week's aftenoons.

Nightwalker-87:
Happy birthday :tada: :smiley:. Have a nice day.

Thanks for sharing your ideas which were interesting to read.
Here are my thoughts, attempts and findings:

a) To ensure proper logging is very useful and not only convenient. Surely we can put that up on the agenda. Currently there are two kinds of logging available. The standard one is based on `uglylogging` and the second one (optional) is `spdlog` for which a wrapper file (`logging_spdlog_wr.h`) is present in the sources. The latter can be enabled in `logging.h`.

b) As you can see, I've reallocated and renamed some header files in the last commit to improve the current structure. Also the source files `common.c` and `sg.c` have been renamed, along with their respective header files, to highlight the oldest and not yet restructured, reimplemented or optimised parts of the codebase.

c) In the meanwhile I found that st-info currently outputs
`st-info: error while loading shared libraries: libstlink.so.1: cannot open shared object file: No such file or directory` though the build test all pass successfully. Returning to older commits, for which this still worked in the past, give the same result. Can you try to reproduce this on your side as soon as you find time for that? I just want to make sure that we do not accidentally introduce a regression bug. Maybe something just messed up on my side.

Looking at this, I'd suggest to line that up in a c) --> b) --> a) approach, which should still go to the v1.8.1 milestone, especially c) and b).

a-michelis:
about `c`

Windows, after install:

image

Linux comming soon

Nightwalker-87:
I found the reason for this. As it appears, there is no regression.
`sudo ldconfig` did the job. 
<del>I'll add a HowTo-note about it to our tutorial soon.</del>
Edit: In fact this was already covered by the compiling instructions. :eyes: :see_no_evil:

Nightwalker-87:
Having sorted quite a few things out in the codebase, I think we should now be ready for logging improvements.
I tried to implement the recent changes carefully and step by step followed by a few very basic tests afterwards.
Some quick interim testing of main tasks and functions of somewhat daily use on another system could turn out helpful though.

a-michelis:
I'm sorry for being MIA, 
Im struggling to find some Free time lately... That said I started implementing a new logging system from scratch, but it's still in progress, that's why i Haven't pushed anything to my fork yet

a-michelis:
Hello again, after all this time!

I managed to test my approach using a "hijack-ey" method, just to try things out. Seems Simple and Helpful.

image

Next step is modifying the rest of the library to use this 😝

Nightwalker-87:
Hi there. Is this topic still alive?

a-michelis:
Hello again, it's been a while... 

Since my wedding is approaches i have less and less time to give to myself (and, thus, the project). I promise to start actively working on it after after the commotion ends (starting early July) and keeping you posted with frequent updates. 

In general I'm interested in maintaining the project more holistically and not just provide "satelite features". That said, I do believe that some structural changes must take place first to allow for modularity and possibly better maintainability.

The hardest issue so far is understanding the inner code workings, so, proper (probably in-code) documentation of the implemented functions would be helpfull...

When the time comes (aka. when the wedding is done), If you could give me _a tour to the codebase_ it would be godsent!

Cheers,
AM.

a-michelis:
I'm officially married! Starting from the following weekend, I'm planning to start diving into the codebase and write some notes about how it works!
If there's any resource on the matter, or even suggestions on how to tackle it, I'd gladly take it.

Quick questions: 
- I'm thinking in-code documentation with some 3rd party tool to extract it and create proper documentation out of it. Any ideas, restrictions or prefferences on that? I've used in-code documentation for my C# Projects before (DocFX) but not for C/C++ libraries.  
- How welcome is C++ code on the codebase? If applicable, what are the restrictions/requirements for C++ code?
- Do you have a preference on what should be implemented when (in terms of wanted tasks and in terms of order)?
- Do you prefer this thread for further discussions/chit-chat or some other way is more appropriate?

Cheers!
A.M.

Nightwalker-87:
@a-michelis: Hi. Good to hear from you.
Congratulations, all the best wishes and may there be many happy returns of this special day. 

Regarding the topic above there has been no further action yet as I could not find any spare time in the recent weeks and months to invest in this project. :disappointed: 

Regarding you questions:

- Maybe it's a good idea for now to solely focus on the logging feature you originally proposed.
- We can keep this place for further conversation, but I'd suggest to update this PR to the latest commit of the `testing` branch, so it can continue to be actively used.
- C++ Code is indeed welcome, especially in places where it helps to improve structuring (understandability), code safety and maintenance of the code base. Assembly for very hardware-near implementations has already has found it's way into the code in some parts.
- We at least used to have an option to use `pandoc`. I'm not sure though, if this option it is still completely functional. `doxygen` may also be worth a discussion later on.
- In-code-documentation: In the end I'd love to have a common source file header formatting as I got to know from other projects characterising the respective source and/or header file, which can be then automatically detected by such tools as mentioned above. Functions should be introduced by a short description together with tags like "@ param", "@ return".

I hope this draws a first idea in which direction one may proceed step by step.

Regards
nightwalker-87

a-michelis:
I've started my logging approach from scratch - since I've done some reading on spdlog - to properly solidify the situation.

Some key factors are: 
- The approach now have three main "logging init" functions, representing 3 different spdlog sinks:
  - `logger_init_stderr` : logs into STDERR with colors (custom STDERR sink with color support)
  - `logger_init_file` : logs into a specified file (default spdlog file sink)
  - `logger_init_callback` : this won't be used by the existing tools, but can later be used by external software which uses stlink library to get structured logging entries. it's a custom sink that, upon receiving a new log, it triggers the callback specified during initialization (which uses simple enough argument types to support interoperability with other languages such as C#)

- All logs contain:
  - Timestamp
  - LOC (File, Function, Line)
  - Log Level
  - Message
    
- the step i'm currently working on is removing ugly logging completely and, since I realized that the log levels seem to correspond to stlink's log levels, to bridge the value gap the same way you did with libusb-stlink log compatibility. 

- Next step is to provide each device with its own logger (and, thus, its own max logging level), so we can instatiate multiple devices at once without issues and without a common logging level. This needs carefull planning, since it requires structural changes into the stlink_t struct.


I suppose that, since I re-iterated the logging system, there now is a new branch which requires a new PR. If i'm wrong, please advice on how to proceed!

If you want to see the current state of logging, I'm about to push the first changes into [my forked repo](https://github.com/a-michelis/libstlink/tree/feature/amichelis/logging).

Cheers,
A.M.

a-michelis:
Side note: I'd gladly convert the whole codebase in C++17 or C++11, but that'll take time.
That's something I plan to do for fun with my fork at some point either way, so, if it's something welcome to the original project, I guess it wouldn't hurt to make a PR or even start the project as a separate repo in the org.

Nightwalker-87:
C++ 17 and C 17 are preferred. GCC 8.1 or later should be for compilation in terms of upcoming system requirements.
A Feature-/Maintenance-PR should be fine.

Nightwalker-87:
The above discussion is going to be continued in issue #1474.
This PR is closed instead as it is outdated in relation to the current state of code development.

Nightwalker-87 (commented):
This contribution needs to be re-addressed in the current issue #1474.

PR #1440, opened 2024-11-21
[feature] libusb-cmake as libusb provider and added support for MSVC

### LINKED ISSUE #1424: [feature] Proper MSVC Building without needing MINGW
Hello, everyone,

I attempted to build the project for native MSVC without success -as expected.
That said, I discovered that the only reason for this is the lack of proper `MSVC` handling on `cmake/modules/Findlibusb.cmake`.

In more detail, the original snippet for MSVC handling is the following:

```cmake
        else (MSVC)
            # libusb library
            set(LIBUSB_NAME libusb-1.0)
            find_library(LIBUSB_LIBRARY
                NAMES ${LIBUSB_NAME}
                HINTS ${LIBUSB_WIN_OUTPUT_FOLDER}/MinGW${ARCH}/dll
                NO_DEFAULT_PATH
                NO_CMAKE_FIND_ROOT_PATH
                )
        endif()
```

If handled correctly, STLink library, as well as all tools are built properly (with some warnings, ofcourse) -at least with VS2022 which i tried.

```diff
index 15afb00..e14993b 100644
--- a/cmake/modules/Findlibusb.cmake
+++ b/cmake/modules/Findlibusb.cmake
@@ -107,12 +107,28 @@ elseif (WIN32 OR (MINGW AND EXISTS "/etc/debian_version"))      # Windows OR cro
         else (MSVC)
             # libusb library
             set(LIBUSB_NAME libusb-1.0)
+
+            if(140 EQUAL ${MSVC_TOOLSET_VERSION})
+                set(VSVERSION "2015")
+            elseif(141 EQUAL ${MSVC_TOOLSET_VERSION})
+                set(VSVERSION "2017")
+            elseif(142 EQUAL ${MSVC_TOOLSET_VERSION})
+                set(VSVERSION "2019")
+            elseif(143 EQUAL ${MSVC_TOOLSET_VERSION})
+                set(VSVERSION "2022")
+            else()
+                set(VSVERSION "2013")
+            endif()
+
             find_library(LIBUSB_LIBRARY
                 NAMES ${LIBUSB_NAME}
-                HINTS ${LIBUSB_WIN_OUTPUT_FOLDER}/MinGW${ARCH}/dll
+                HINTS ${LIBUSB_WIN_OUTPUT_FOLDER}/VS${VSVERSION}/MS${ARCH}/static
                 NO_DEFAULT_PATH
                 NO_CMAKE_FIND_ROOT_PATH
                 )
+
+            #fix for ill-defined ssize_t in basetsd.h
+            add_compile_definitions(_SSIZE_T_DEFINED ssize_t=int64_t)
         endif()
     endif()
```

The structure of 3rd party `.7z` contains different releases for different versions of VS. For that reason, the patch discovers the proper release to use (see the big if-elseif-else cascade).

Since it seems perfectly doable but I believe I lack basic cmake knowledge to consider the patch above decent, I share it with you to further discuss it and potentially merge it to the code!

Cheers, 
A.M.

a-michelis:
## Overview
This pull request alters `Findlibusb.cmake` in order to add proper Windows MSVC handling. It discards the old MINGW approach and uses the `libusb-cmake` repository as the main provider instead **for both Windows and Linux**. As it stands now, if `libusb` is not installed for these two OSes, the [libusb-cmake](https://github.com/libusb/libusb-cmake) repository is being cloned and incorporated into the build. This way,`stlink` has a freshly built `libusb`, tailored to the system's configuration and needs, installed into the same `INSTALL_PREFIX` as `stlink`.

Additionally, this pull request adds Windows/MSVC support into the c/c++ github workflow.

## Potential issues
1. In Windows, there is a possibility that the installation path for all `LIBRARY` items must be the same as the one of `RUNTIME` items, as opposed to the Linux paradime, where `LIBRARY` and `ARCHIVE` items are the ones that share location. This needs further investigation and potentialy a complementary fix, as I currently have no equipment and a clean environment to test it. I'll get back to it once I manage to prepare the proper testing ground.

2. One very important note on this PR is that it probably drops support for MINGW. If mingw is still needed I'm going to need assistance, Since I've never worked with it in the past and i do not know how to properly validate such builds.

(Closes #1424)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -2,7 +2,7 @@ name: C/C++ CI
 
 on:
   push:
-    branches: [master, develop, testing, feature/libusb_cmake]
+    branches: [master, develop, testing]
   pull_request:
     branches: [master, develop, testing]
 
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -2,7 +2,7 @@ name: C/C++ CI
 
 on:
   push:
-    branches: [master, develop, testing]
+    branches: [feature/libusb_cmake]
   pull_request:
     branches: [master, develop, testing]
 
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -203,7 +203,6 @@ jobs:
           cmake --build . --target ALL_BUILD
       - name: make release
         run: |
-          mkdir build
           mkdir build\\release
           cd build\\release
           cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Release"
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -2,7 +2,7 @@ name: C/C++ CI
 
 on:
   push:
-    branches: [feature/libusb_cmake]
+    branches: [master, develop, testing, feature/libusb_cmake]
   pull_request:
     branches: [master, develop, testing]
 
@@ -188,6 +188,40 @@ jobs:
         run: sudo make package
       - name: sudo make uninstall
         run: sudo make uninstall && sudo make clean
+
+  job_windows_msvc:
+    name: windows MSVC
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: debug
+        run: |
+          mkdir build
+          mkdir build\\debug
+          cd build\\debug
+          cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Debug"
+          cmake --build . --target BUILD_ALL
+      - name: make test
+        run: |
+          cd build\\debug
+          rm -Recurse -Force .\\*
+          cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Debug"
+          cmake --build . --target test
+      - name: make release
+        run: |
+          mkdir build
+          mkdir build\\release
+          cd build\\release
+          cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Release"
+          cmake --build . --target BUILD_ALL
+      - name: sudo make install
+        run: |
+          cd build\\release
+          cmake --build . --target install
+      - name: sudo make uninstall
+        run: |
+          cd build\\release
+          cmake --build . --target uninstall
 # Linux MinGW cross compliation
 
 # job_linux_22_04_cross:
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -200,7 +200,7 @@ jobs:
           mkdir build\\debug
           cd build\\debug
           cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Debug"
-          cmake --build . --target BUILD_ALL
+          cmake --build . --target ALL_BUILD
       - name: make test
         run: |
           cd build\\debug
@@ -213,11 +213,11 @@ jobs:
           mkdir build\\release
           cd build\\release
           cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Release"
-          cmake --build . --target BUILD_ALL
+          cmake --build . --target ALL_BUILD
       - name: sudo make install
         run: |
           cd build\\release
-          cmake --build . --target install
+          cmake --build . --target INSTALL
       - name: sudo make uninstall
         run: |
           cd build\\release
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -201,12 +201,6 @@ jobs:
           cd build\\debug
           cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Debug"
           cmake --build . --target ALL_BUILD
-      - name: make test
-        run: |
-          cd build\\debug
-          rm -Recurse -Force .\\*
-          cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Debug"
-          cmake --build . --target test
       - name: make release
         run: |
           mkdir build
@@ -214,7 +208,7 @@ jobs:
           cd build\\release
           cmake ..\\.. -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE="Release"
           cmake --build . --target ALL_BUILD
-      - name: sudo make install
+      - name: make install
         run: |
           cd build\\release
           cmake --build . --target INSTALL
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,7 @@ endif()
 ###
 
 find_package(libusb REQUIRED)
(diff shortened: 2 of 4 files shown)


.github/workflows/c-cpp.yml
```diff
@@ -2,7 +2,7 @@ name: C/C++ CI
 
 on:
   push:
-    branches: [master, develop, testing]
+    branches: [master, develop, testing, feature/libusb_cmake]
```
Nightwalker-87: We don't need this here. The workflow is called upon merging into testing and the `testing` branch may break, that's totally acceptable for it's dedicated purpose. :wink:

cmake/modules/Findlibusb.cmake
```diff
@@ -137,8 +108,29 @@ else ()                                                         # all other OS (
 
     FIND_PACKAGE_HANDLE_STANDARD_ARGS(libusb DEFAULT_MSG LIBUSB_INCLUDE_DIR LIBUSB_LIBRARY)
     mark_as_advanced(LIBUSB_INCLUDE_DIR LIBUSB_LIBRARY)
-
     if (NOT LIBUSB_FOUND)
-        message(FATAL_ERROR "libusb library not found on your system! Install libusb 1.0.x from your package repository.")
+        message(STATUS "No libusb-1.0 not installed into your system. Downloading and building it from source")
```
Nightwalker-87: This is not a good practise from my point of view.
All unix-based systems should continue to install and use `libusb `provided from their respective package repositories. It must be considered a very bad attempt to mix locally installed newer library versions with probably already existing dependencies used by other (system) packages originating from the package system, as this opens the door for incompatibilities and functional failures.

Nightwalker-87 (changes requested):
Please solely introduce changes affecting the windows operating system due to the package-based nature of unix-based systems to ensure the verification of system integrity.

.github/workflows/c-cpp.yml
a-michelis: Yes, Correct -I forgot to remove it, Will update ASAP!

cmake/modules/Findlibusb.cmake
a-michelis: Ok, no problem, Will revert to the old practice! 

As for windows, Will do some checks for dll locations compatibility later today or tmr, so we can be sure for this as well

a-michelis:
Tested the MSVC Build and it needs some additional root cmake modifications to work, but totally easy to do. Will come back with them sometime later today. 

As for the library part, This is something that cannot be brought to this PR, since it requres restructuring in both code and CMAKE: 

- Since windows have their own `usb.h`, it conflicts with `stlink`'s header and, thus, it's required to put the headers within a prefixed subfolder and use them as such (ie `#include <stlink/usb.h>`). This requires code modifications, as all headers that have `#include <someStlinkHeader.h>` must be converted to `#include <stlink/someStlinkHeader.h>`.

- Windows require `__declspec(dllexport/dllimport)` to each desired function, in order to properly export functionality. This is solvable by creating an `stlink/api.h` header with the well-known macro definition (let's call it `STLINK_API`), that prepends dllexport when building and dllimport when using (via compiler definition). That's easy to address- essentially a templated copy-paste-edit snippet.

CMakeLists.txt
```diff
@@ -262,15 +263,32 @@ set_target_properties(${STLINK_LIB_SHARED} PROPERTIES
...
 # Link shared library
 if (WIN32)
     target_link_libraries(${STLINK_LIB_SHARED} ${LIBUSB_LIBRARY} ${SSP_LIB} wsock32 ws2_32)
 else ()
     target_link_libraries(${STLINK_LIB_SHARED} ${LIBUSB_LIBRARY} ${SSP_LIB})
 endif()
 
-install(TARGETS ${STLINK_LIB_SHARED} ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
-
+# For windows, shared libraries go to `bin` and their implibs go to `lib`
```
Nightwalker-87: Documentaton: Replace "implibs" with "imported libraries (implibs)" in comment for clarification.

cmake/modules/Findlibusb.cmake
```diff
@@ -51,75 +52,42 @@ elseif (CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")                   # OpenBSD; libus
         message(FATAL_ERROR "No libusb-1.0 library found on your system! Install libusb-1.0 from ports or packages.")
     endif()
 
-elseif (WIN32 OR (MINGW AND EXISTS "/etc/debian_version"))      # Windows OR cross-build with MinGW-toolchain on Debian
```
Nightwalker-87, line 54: Would it be possible to leave
`elseif (MINGW AND EXISTS "/etc/debian_version") # Cross-build with MinGW-toolchain on Debian`

including the following associated code block for now, until cross building via MVSC on Linux has also been tested and verified? The idea is to drop this part as well later on, if everything works well. I'd just prefer to keep at least one working approach for now ...

Nightwalker-87 (changes requested):
I consider this a good and very useful contribution, with only a few remaining FMPOV accountable remarks that should be easily addressable. Thank you very much. :+1:

CMakeLists.txt
a-michelis: On it! Will change it to "import Libraries (implibs)" since that's the full name for windows - these `.lib` files are not static libraries, they're _symbol containers_ and generally instruct MSVC linker how to link to the `.dll` .

cmake/modules/Findlibusb.cmake
a-michelis: I believe this is possible. That said, some additional things must be performed within the `CMakeLists.txt`, to further distinguish between MSVC and MINGW/other.

Nightwalker-87 (approved):

Nightwalker-87:
@a-michelis: Sorry for the delay. Having swapped hardware recently, I needed some time last WE to finalise the setup of my working environment...
What came to my mind is that we should have a small description in our compiling manual (doc/compiling.md), but I can take cake of that afterwards. However some notes on how to proceed in the original issue could be helpful, I guess.

a-michelis:
> What came to my mind is that we should have a small description in our compiling manual (doc/compiling.md), but I can take cake of that afterwards. However some notes on how to proceed in the original issue could be helpful, I guess.

@Nightwalker-87: I totally agree, It won't take much to add some compilation guidance to compiling.md.
I believe I can allocate some time at 26/12 and have it committed by the end of that day!

Small Context & suggestion:
As it is now, Compilation for MSVC uses solely CMAKE commands (CMAKE uses the installed msvc compiler).
It is my understanding that the provided makefile is a wrapper to the appropriate CMake Commands. If my understanding is correct, unifying the testing methodology in all OSes (in C/C++ Workflow) to use cmake would be a good practice imho. That would be a separate PR of course, but I leave it here as food for thought.

a-michelis:
I'm sorry for being late on my promise, Holidays got the better of me! Added the compilation manual, as discussed!

Nightwalker-87 (approved):

Nightwalker-87:
> It is my understanding that the provided makefile is a wrapper to the appropriate CMake Commands. If my understanding is correct, unifying the testing methodology in all OSes (in C/C++ Workflow) to use cmake would be a good practice imho. That would be a separate PR of course, but I leave it here as food for thought.

True, surely a good attempt. Maybe one may update and reuse the other Draft-PR for this.

Thanks again for your valuable contribution @a-michelis. I believe we are now ready to merge.

Nightwalker-87 (approved):

PR #1449, opened 2025-01-10
[feature] Corrected and unified GitHub-Actions C/C++ CI workflow

### LINKED ISSUE #1446: [feature] GitHub-Actions C/C++ CI: All tests are currently performed on GCC 9.4.0
As I mentioned on https://github.com/stlink-org/stlink/pull/1440#issuecomment-2561168962 , I started looking at the C/C++ Workflow in order to attempt and unify the approach. While doing so, I discovered that all of the CLANG tests were NOT using clang, and used GCC instead.

The reason behind this is that all ubuntu workers on GH are equipped with both gcc-10 and clang-12. Installation of these two was doing essentially nothing, and since GCC was the preffered by cmake (unless specifically requested by setting env variables), all clang tests are running with gcc.

**Solution:**
For clang tests, using `env` section in `.yaml` to define `CC` and `CXX`  variables solves the issue.

```yaml
  job_linux_22_04_64_clang:
    name: ubuntu-22.04 clang
    runs-on: ubuntu-22.04
    env:
      CC: clang-12
      CXX: clang++-12
```

Continuing further, I discovered that, when using CLang in release, CMake appends some integer conversion warning-as-error flags, which are not possible to be removed unless we interfere directly with the cmake files (possibly). That said, the only issue here was a simple implicit casting on libstlink's `helper.cpp` which, when explicitly casted, stopped producing the error

`a/src/stlink-lib/helper.h:29`:
```c
value = strtol(str, &tail, 10);
```

`b/src/stlink-lib/helper.h:29`:
```c
value = (int32_t)strtol(str, &tail, 10);
```

After that I discovered another major issue-duo:
In addition to CLang tests, GCC tests were not using the correct GCC version as well - again solved by defining CC and CXX env variables. BUT, this time it was obvious that `M32` flags were not being registered as well, which led me to define them using the _workflow-intended_ way (using `env` section within the step). This led to the 2nd major issue which is: The compilation fails when using `m32` in both gcc-10 and clang-12, and I still can't figure out why, except the fact that it is related to the linking process.

Ideas? Pointers? Suggestions?
A.M.

a-michelis:
As there where many problems with the previous C-Cpp Github actions workflow, a few changes were made:

- All checkout steps were bumped from `v2` to `v4`, as v2 is about to be deprecated
- Instead of individual jobs , two matrices were implemented: one for x32 compilations and one for x64
- Compilation flags and compiler definitions are now properly defined using `env` keys within the workflow
- All `make` calls were replaced with their equivalent `cmake` counterparts, to provide a unified methodology of testing across all platforms and OSes.
- Many more OS-Compiler variants were added -all of which are used for both x32 and x64 compilation. The list now consists of:
  - Ubuntu 20.04:
    - GCC-9
    - GCC-10
    - CLang 10
    - CLang 12
  - Ubuntu 22.04:
    - GCC-10
    - GCC-11
    - GCC-12
    - CLang 11
    - CLang 12
    - CLang 13
    - CLang 14
    - CLang 15
  - Ubuntu 24.04:
    - GCC-10
    - GCC-11
    - GCC-12
    - GCC-13
    - CLang 14
    - CLang 15
    - CLang 16
    - CLang 17
    - CLang 18

This PR Closes #1446 .

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -13,11 +13,23 @@ jobs:
       fail-fast: false
       matrix:
         config:
+          - name: Ubuntu20.04_x64 gcc9
+            os : ubuntu-20.04
+            cc : gcc-9
+            cxx : g++-9
+            packages: gcc-9 g++-9
+
           - name: Ubuntu20.04_x64 gcc10
             os : ubuntu-20.04
             cc : gcc-10
             cxx : g++-10
             packages: gcc-10 g++-10
+
+          - name: Ubuntu20.04_x64 clang10
+            os : ubuntu-20.04
+            cc : clang-10
+            cxx : clang++-10
+            packages: gcc-9 g++-9 clang-10
 
           - name: Ubuntu20.04_x64 clang12
             os : ubuntu-20.04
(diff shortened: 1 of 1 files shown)


Nightwalker-87 (approved):

Nightwalker-87:
Thank you very much for this valuable contribution. :1st_place_medal:

PR #1466, opened 2025-04-16
[feature] STM32 flash type implementation for WB05, WB06/07, WB09, WL3x

dbeinder:
I have implemented support for a new flash type, tentatively named `WB0`. These devices were originally called BlueNRG, but a year ago ST has [renamed them](https://blog.st.com/stm32wb09-stm32wb05-stm32wb06-stm32wb07-stm32wb05xn/) and now markets them under the STM32 brand. Probably because they were not originally intended to integrate into the STM32 ecosystem, they are a bit quirky:

- Flash starts at `0x10040000`
- Completely different `FLASH` peripheral. No unlocking scheme, writing is done by filling data registers and triggering a write command.
- No `IDCODE` register, however there is a `JTAG_ID` register which has a `PART_NUMBER` field. STM32CubeProg uses this as the DeviceID, so I've followed their example: 
grafik
- No `DBGMCU` block to gate watchdogs conditionally when a debugger is attached.
- No option bytes - there is a protection mechanism, but I haven't added this yet.
- Writing to the OTP block is different from normal flash write.

I have written a fast flashloader and direct support for byte-granular writes to OTP.

Apparently [STM32WL3x](https://www.st.com/en/microcontrollers-microprocessors/stm32wl3x.html) also shares this architecture, but for lack of a documented family name, I have stuck to `WB0` as the `flash_type`.

Tested only on `STM32WB07CC` so far, with STLINK-V3MINIE and the on-board `STLINK-V3EC` on the `NUCLEO-WB07CC` dev board.

# References
[STM32WB05 (BlueNRG-LPS) RM0491](https://www.st.com/resource/en/reference_manual/rm0491-the-bluenrglps-arm-cortex-m0based-stmicroelectronics.pdf)
[STM32WB06/07 (BlueNRG-LP) RM0530](https://www.st.com/resource/en/reference_manual/rm0530--stm32wb07xc-and-stm32wb06xc-ultralow-power-wireless-32bit-mcus-armbased-cortexm0-with-bluetooth-low-energy-and-24-ghz-radio-solution-stmicroelectronics.pdf)
[STM32WB09 (BlueNRG-LPF) RM0505](https://www.st.com/resource/en/reference_manual/rm0505-stm32wb09xe-ultralow-power-wireless-32bit-mcu-armbased-cortexm0-with-bluetooth-low-energy-and-24-ghz-radio-solution-stmicroelectronics.pdf)
[STM32WL3x RM0511](https://www.st.com/resource/en/reference_manual/rm0511-stm32wl33xx-armbased-wireless-mcus-with-subghz-radio-solution-stmicroelectronics.pdf)
ST's OpenOCD implementation: [bluenrg-x.c](https://github.com/STMicroelectronics/OpenOCD/blob/0de861e21c12fd39fdce7b6296994d89eca6146f/src/flash/nor/bluenrg-x.c)


dbeinder:
The reason I changed `write_buffer_to_sram()` to handle padding is because the approach used in the existing code seems to introduce out-of-bounds reads from the mmap'd file buffer?
https://github.com/stlink-org/stlink/blob/b149e2448e30a3fbf68ebd7adef9e08818977c56/src/stlink-lib/flash_loader.c#L752-L755

Nightwalker-87 (approved):

PR #1483, opened 2025-10-29
[feature] Make st-info --probe run probes in parallel

CreaxxOG:
This makes st-info --probe run each probe in parallel and makes the entire command much faster. We use this for our internal projects and it has improved the performance drastically.

It builds & passes all tests locally.


CreaxxOG:
@Nightwalker-87 If you could please review that would be great!

Nightwalker-87 (approved):

PR #1496, opened 2026-07-02
[feature] Added st-server (and remote backend) to support remote control of an STLINK programmer

jameswalmsley:
I got frustrated at having my nucleo board dangling from my laptop, and our internal tooling is built around the st-link flashing tools. Even though there's a gdb-server with remote access, I wanted to be able to st-flash etc remotely.

Adding remote support turned out to be rather neat, by adding a backend to support remote access to a new tool  (st-server). The client simply forwards all backend requests to the server. The integration with the current code-base is minimal and completely orthogonal to the local st-link use case.

Once the backend was added, adding support for all tools was easy, and I have tried and tested st-info, st-flash, st-util and st-trace (with STM32H5 and STM32F4 series).


Nightwalker-87:
@jameswalmsley: Thank you very much for this valuable contribution! :1st_place_medal:

Nightwalker-87 (approved):

PR #1502, opened 2026-07-16
Explicit typing for main header files

AlexandreBlais42:
I made the conversion explicit instead of relying on undefined behavior.

### DIFF
--- a/inc/stlink.h
+++ b/inc/stlink.h
@@ -35,37 +35,37 @@ enum target_state {
     TARGET_DEBUG_RUNNING = 4,
 };
 
-#define STLINK_CORE_RUNNING             0x80
-#define STLINK_CORE_HALTED              0x81
+#define STLINK_CORE_RUNNING                0x80
+#define STLINK_CORE_HALTED                 0x81
 
 /* STLINK modes */
-#define STLINK_DEV_DFU_MODE             0x00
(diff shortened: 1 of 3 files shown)


Nightwalker-87 (changes requested):
@AlexandreBlais42: I see the point.
However we try to avoid **int** in favour of fixed type definitions **int$_t** wherever possible.
Also `0xA05F` is a positive value and the official, constant write key for the Debug Core Register (DHCSR). Thus we should make use of **uint32_t** here. Could you consider this?

AlexandreBlais42:
I see that macros in `stm32_registers.h` are often used in the data argument of
```
int32_t stlink_write_debug32(stlink_t *sl, uint32_t addr, uint32_t data);
```
and thus should all be `uint32_t` but are defined with `(1 << n)` or similar which makes them coerce to `int`. Is this a concern?

Nightwalker-87:
@AlexandreBlais42: We need to check on that.

Nightwalker-87:
Yes, indeed we should stick to this approach (e.g.):

```
#define FLASH_CR_PER (1U << 2)
#define FLASH_CR_STRT (1U << 3)
#define FLASH_CR_PSIZE_8 (1U << 4)
// ...
```

Nightwalker-87 (approved):

Issue #1482, opened 2025-10-29
[STM32L4R5ZIT6]: STM will halt unless flashed even number of times


Robin-OhPL:
## Summary

* Flashing with `st-flash` or GDB `load` can halt the chip on odd-numbered flashes.
* Re-flashing or power cycling recovers the chip. Pressing reset does not help.
* STM32CubeProgrammer (GUI or CMD) does not show this issue.
* Core ID discrepancy suggests a potential STLink detection problem

---

Typically, we flash the new program onto the STM32L4R5ZIT6 with either a STLINK-V3SET or a STLINK-V3MINIE using one of two methods on a computer running native Ubuntu 22.04 :

1. **Using `make flash`**  
   This is a wrapper that internally runs:  
   ```bash
   st-flash --connect-under-reset --reset write $(BUILD_DIR)/$(TARGET).bin 0x8000000
   ```

2. **Using GDB**
   Start the debugging session with:

   ```bash
   st-util
   gdb-multiarch build/firmware.elf -ex="tar extended-remote :4242"
   ```

   Then, run `load` inside GDB to program the chip.

---

## Observed Behavior

Whenever we use either of the above methods, the chip always halts if programmed an **ODD number of times**..

In GDB, the sequence looks like this:

```gdb
Reset_Handler () at startup_stm32l4r5xx.s:91
91        str  r3, [r2]
(gdb) load
Loading section .isr_vector, size 0x1c0 lma 0x8000000
Loading section .text, size 0x16f10 lma 0x80001c0
Loading section .rodata, size 0x7a0 lma 0x80170d0
Loading section .ARM.extab, size 0x308 lma 0x8017870
Loading section .ARM, size 0x4f8 lma 0x8017b78
Loading section .init_array, size 0x18 lma 0x8018070
Loading section .fini_array, size 0x8 lma 0x8018088
Loading section .data, size 0x880 lma 0x8018090
Start address 0x08015cc8, load size 100624
Transfer rate: 27 KB/sec, 7740 bytes/write.
(gdb) c
Continuing.

Program received signal SIGTRAP, Trace/breakpoint trap.
0x1fff51da in ?? ()
(gdb) bt
#0  0x1fff51da in ?? ()
#1  0xfffffffe in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) start
The program being debugged has been started already.
Start it from the beginning? (y or n) y
[...]

PR #1503, opened 2026-07-23
Fix: disable halting debug and disarm fault catches on detach after a reset

dakejahl:
## Summary
After `st-flash reset` (or `write --reset`), the tool detaches leaving the target with `DHCSR.C_DEBUGEN` set and the vector catches armed at connect (`VC_HARDERR`, `VC_BUSERR`) still live. With halting debug enabled and no debugger attached, the target's next hard or bus fault halts the core silently instead of running its fault handler — the board looks bricked until a power cycle. This extends the recent `VC_CORERESET` detach cleanup to the debug-enable itself and the remaining connect-time catches.

## Problem
Connecting arms halt-on-fault state: `stlink_soft_reset()` writes `DEMCR = TRCENA | VC_HARDERR | VC_BUSERR (| VC_CORERESET)` and sets `DHCSR.C_DEBUGEN`. These live in the debug power domain, survive every system reset, and are cleared only by a power-on reset.

`stlink_exit_debug_mode()` skips its `DHCSR = DBGKEY` debug-disable whenever `core_stat == TARGET_RESET` — which is exactly the final state of `st-flash reset` and `write --reset`: `stlink_reset()` sets it, and `stlink_run()` never updates it. The recent `VC_CORERESET` cleanup runs despite that skip and fixed the halt-on-`NVIC_SystemReset()` case, but the target still detaches with `C_DEBUGEN = 1` and `VC_HARDERR | VC_BUSERR` armed, so a later fault halts the core at the fault vector with nothing attached to resume it. Fault-logging firmware and watchdog recovery are silently bypassed.

Observed in production flashing of STM32H743 flight controllers: on 1.8.0-era code (before the `VC_CORERESET` cleanup), `st-flash reset` after a `--connect-under-reset` flash left every board halting silently at its own next software `reboot`. Reading the registers back over `--hot-plug` while one sat hung showed `DHCSR = 0x00030003` (`C_DEBUGEN | C_HALT | S_REGRDY | S_HALT`) and `DEMCR = 0x01000501` — exactly the connect-time arming. Current `testing` fixes the reset-catch half; the fault-catch residue reproduces the same way.

## Solution
In `stlink_exit_debug_mode()`, once the target is identified: write `DHCSR = DBGKEY` unconditionally, dropping the `TARGET_RESET` skip — that state is precisely the reset-last-command case that needs the cleanup, and the existing `VC_CORERESET` cleanup already performs debug-register writes in that same state, so the access is proven safe there. Extend the DEMCR cleanup to disarm all three connect-time vector catches (`VC_CORERESET | VC_HARDERR | VC_BUSERR`), still preserving unrelated bits (`TRCENA`, `MON_EN`).

### DIFF
--- a/src/stlink-lib/common_legacy.c
+++ b/src/stlink-lib/common_legacy.c
@@ -455,25 +455,30 @@ int32_t stlink_exit_debug_mode(stlink_t *sl) {
   }
 
   if(sl->flash_type != STM32_FLASH_TYPE_UNKNOWN) {
-    if(sl->core_stat != TARGET_RESET) {
-      // stop debugging if the target has been identified
-      stlink_write_debug32(sl, STM32_REG_DHCSR, STM32_REG_DHCSR_DBGKEY);
-    }
+    // stop debugging if the target has been identified. This must also run
+    // when the last command was a reset (core_stat == TARGET_RESET, e.g.
+    // `st-flash reset` or `write --reset`, since stlink_run() does not update
+    // core_stat): C_DEBUGEN survives system resets, and left set it keeps any
+    // armed vector catch live, halting the target with no debugger attached.
+    stlink_write_debug32(sl, STM32_REG_DHCSR, STM32_REG_DHCSR_DBGKEY);
 
-    // Disarm vector-catch-on-reset before detaching. --connect-under-reset
-    // arms DEMCR.VC_CORERESET (halt-on-reset); it lives in the debug power
-    // domain and survives NRST, so if left armed the core halts on every reset
-    // -- including NVIC_SystemReset() from firmware -- and the board looks
-    // bricked until a power cycle. This must run even when core_stat is
-    // TARGET_RESET, which is exactly the connect-under-reset case that would
-    // otherwise skip the cleanup. Only the read failing (dead debug link)
-    // stops us, and we preserve the other DEMCR bits.
+    // Disarm the vector catches armed at connect before detaching.
+    // stlink_soft_reset() arms DEMCR.VC_CORERESET (halt-on-reset) and
+    // VC_HARDERR/VC_BUSERR (halt-on-fault); they live in the debug power
+    // domain and survive NRST, so if left armed the core halts on every reset
+    // -- including NVIC_SystemReset() from firmware -- or silently at its next
+    // hard/bus fault instead of running the fault handler, and the board looks
+    // bricked until a power cycle. Only the read failing (dead debug link)
+    // stops us, and we preserve the other DEMCR bits (TRCENA, MON_EN).
+    const uint32_t vector_catches = STM32_REG_CM3_DEMCR_VC_CORERESET |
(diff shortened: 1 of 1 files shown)


Nightwalker-87 (approved):

Nightwalker-87:
@dakejahl Is there any (partial) relation to issue #1482? Please check and discuss.

jameswalmsley (approved):

dakejahl:
Not related — different root cause, and this PR won't fix #1482.

**#1482 is a boot-source problem.** The transcripts there show the core executing at `0x1fff51da` and `0x1fff34c2`, the second after a plain Ctrl-C. That's system memory, so the part is running the ROM bootloader instead of the application that was just written. Boot source is decided by BOOT0 / nSWBOOT0 / nBOOT0 / nBOOT1 and the L4 empty-check flag — none of those live in the debug power domain, and no DHCSR or DEMCR bit can steer them. A `VC_CORERESET` halt would stop at the application's reset vector, a `VC_HARDERR` halt at the faulting instruction; neither gives two different addresses inside ROM.

**What the two do share is the fingerprint:** NRST doesn't recover it, a power cycle does, and CubeProgrammer is unaffected. Both the debug power domain and the L4 empty-check latch are cleared only by a power-on reset, which is why they look alike from the outside.

There is one place st-flash could produce #1482's symptom, but it's a separate bug and unproven: `st-flash erase` calls `stlink_reset()` immediately after the mass erase (`src/st-flash/flash.c`), so the target leaves reset with `0x08000000` reading `0xffffffff` — the condition the L4/G0/G4/WB/WL empty check latches on. Their posted commands don't include `erase`, so it stays a hypothesis. I'd rather post a probe on #1482 that reads the boot source and the relevant flash and option bits after each flash, so the odd/even parity can be pinned to a specific bit, than guess at a fix here. We own no L4 part to confirm it on.

**On this PR:** validated on an STM32H743 with an STLINK-V3, against current `testing` as the control. After `--connect-under-reset reset`, `reset`, and `write --reset`, `testing` detaches with `DEMCR=0x01000500` and skips the DHCSR disable; with this PR it's `DEMCR=0x01000000` and the disable is written.

One result is worth calling out because it argues for this PR specifically. On a board where **NRST is not wired**, `stlink_reset(RESET_AUTO)` falls back to `stlink_soft_reset(sl, 0)`, which rewrites DEMCR *without* `VC_CORERESET` — so #1500's cleanup effectively self-disarms there, and what remains armed after detach is `VC_HARDERR | VC_BUSERR`. I then had the firmware take a deliberate bus fault after detaching: with current `testing` the board never came back, because the core halted at the fault and the fault handler never ran. With this PR it ran its handler, rebooted, and came back.

Happy to attach the probe scripts if they're useful.

Nightwalker-87:
Hi @dakejahl. Thx for looking into this! :rocket:
It would be great if you could share your probe scripts on pastebin together with your related findings in #1482.
As I have a STM32L476RG Nucleo-Board here we can use it for testing.

I'll merge this PR soon then.

PR #1501, opened 2026-07-10
Fixes for MinGW cross compilation on Linux to generate Windows binaries (Part 2) 

Nightwalker-87:
This is a subsequent PR to #1497.

PR #1497, opened 2026-07-05
Fixes for MinGW cross compilation on Linux to generate Windows binaries (Part 1)

jameswalmsley:

PR #1504, opened 2026-08-01
Fixes for MinGW cross compilation on Linux to generate Windows binaries (Part 3) and MSVC compatibility

### LINKED ISSUE #1472: [STM32F746G DISCOVERY]: st-util.exe doesn't run
---

- [x] I made serious effort to avoid creating duplicate or nearly similar issue
- [x] Programmer/board type: ST-LINK/V2-1
- [x] Operating system an version: Windows 11 Pro
- [x] **stlink tools version** and/or git commit hash: v1.8.0
- [x] stlink commandline tool name: `st-util`
- [x] Target chip (and board, if applicable): 32F746GDISCOVERY

I download Release v1.8.0 and extract it.
Then in powershell:
```
> cd .\stlink-1.8.0-win32\bin\
> st-util.exe
>
```

There is no output. When I try to launch any of the tools, they close printing nothing. The previous version v1.7.0 works as expected.

### LINKED ISSUE #1478: Windows binaries
- [x] Programmer/board type: STLINK/V2-1
- [x] Operating system an version: Win10 IoT LTSC
- [x] **stlink tools version** and/or git commit hash: 1.8.0 (Latest Windows Binary Release)
- [x] stlink commandline tool name: All
- [x] Target chip (and board, if applicable): STM32F401RE (NUCLEO-F401RE)

When I attempt to flash in VSCodium:

https://pastebin.com/Wk4HVQwB

ERROR:

> [build] C:/Program Files (x86)/stlink/config/chips: No such file or directory

It seems that the location to the config folder is somehow fixed. I ve searched through MyProject, a project that uses stlink, and found nothing related to this fixed location.

Nightwalker-87:
**Restored cross-compiling for Windows binaries (i686 and x86-64) & major refactoring of `cmake` build configuration**

1) Major refactoring for `cmake` build configuration
- New subfolder `stlink-env` for system-specific sources, headers and libraries
- System-specific definitions, header files and libraries relocated and condensed in subordinate `CMakeLists.txt`
- Simplified environment-specific compilation of function `init_chipids( )`
- Corrected system-specific include in `map_file.h`
- Raised min. version for cmake to v3.21
- Updated version_support.md

2) Fixes for cross-compiling and native compiling on Windows (WIN32, MSVC, MINGW)
- Fixes for `cmake` build configuration
- Fix for `getopt`: compile-time constants
- Fixed filepath issue for chipid files
- Added missing include guard for `win32_socket.h`
- Fixed includes in `trace.h`, `remote.h`, `gdb-remote.h`
- Minor formatting fixes
- Use vcpkg with MSVC (contribution by [slyshykO])
- Resolved winsock.h conflict
- Replaced pthread usage with a platform thread wrapper

3) Fixed `libusb` compilation with `autotools`
- Fixed broken Multi-COMMAND-execute_process()-pipe in `Findlibusb.cmake`, causing race conditions
- Updated shell script `gen_binaries_win.sh`
- Removed duplicate `libusb` build routine
- Removed old libusb_settings.h (leftover)

(Closes #1472) (Closes #1478) (Closes #1497) (Closes #1501)


Nightwalker-87:
Could someone please review this PR and test the changes in a native Windows environment?

klonyyy:
Hey @Nightwalker-87, I can go through the changes and test the compilation on my Windows machine - apart from compiling and running the stlink lib do you have any other tests you'd like me to perform?

Nightwalker-87:
@klonyyy Thank you for your help. I have tested the cross-complining intensively on Debian, but my possibilities on running and actively using the toolset is very limited due to VM use. This it would be great to have people there to test it more intensively on a native Windows system.
The focus should be set on compilation, proper placing of all dependencies and files and the focus on error messages and notifications - in short the issues people came across in the last release v1.8.0. I hope these are all resolved by now. Please also let me know, if there are any inconsistencies in this PR.

I am aware of that there are quite a few changes to review though...
If you have any specific questions, please let me know.

klonyyy:
No problem - just pulled the branch, compiled on Win11 with MinGW without issues. 

Tested:
1. #1472 - the output shows up correctly
2. #1478 - the flashing works on my STM32G474CC - no info about missing chipids
3. #1497 #1501 - couldn't find way to reproduce I'm guessing these were general topics for failed compilation under Windows which now works.

I went through the changes, and I didn't see obvious issues, however I don't yet have a full architecture picture of the whole project.

klonyyy (approved):

Nightwalker-87:
@klonyyy This is really good news for all of us, as it proves that all spent efforts have finally lead to the intended result.

2) indicates that the path settings for these files in the respective subdirectory are now correct, otherwise you would have seen a message that the respective file for the device you have could not be found.
3) Don't bother about this. As indicated, these PRs were the first three parts (interim steps) of the whole solution. They remain listed for documentation purposes in the upcoming release notes and their relation to the mentioned issues.

Nightwalker-87:
Waiting for at least a second reviewer here...

Nightwalker-87:
Waiting for at least a second reviewer here...
ping @slyshykO

slyshykO:
@Nightwalker-87, sorry for the delay and my detachment from the project.
Can you tell me if we still need to support building with MSVC, cause this branch doesn't support it? 
libusb wasn't downloaded.

src/stlink-lib/remote.h
```diff
@@ -45,6 +46,10 @@
 #include <stlink.h>
 #include <stlink_backend.h>
 
+#include "remote.h"
```
slyshykO, line 49: We really need to include "remote.h" inside "remote.h" ?

src/stlink-lib/remote.h
```diff
@@ -23,14 +23,15 @@
...
+
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
 
 #if defined(_WIN32)
-#include <windows.h>
+#include <win32_socket.h>
```
slyshykO, line 34: we should not include win32_socket.h here.
It leads to that in flash.c, line 291, we call win32_write_socket instead of regular write.

src/stlink-lib/remote.h
Nightwalker-87: Well spotted - of course not...

src/stlink-lib/remote.h
Nightwalker-87: Have you also tested this without it?
I assume we need

```
#if defined(_WIN32)
#include <win32_socket.h>
#endif
```

in `remote.c` then instead.

Nightwalker-87:
> @Nightwalker-87, sorry for the delay and my detachment from the project. Can you tell me if we still need to support building with MSVC, cause this branch doesn't support it? libusb wasn't downloaded.

I haven't tested MSVC and focussed on the original approach which was present before this was added.

@slyshykO Add the following code block in `Findlibusb.cmake` to see if this is applicable:

```
# Windows (native MSVC or MinGW without cross-compiling)
elseif(MSVC OR (WIN32 AND NOT EXISTS "/etc/debian_version"))
    # Fix for missing ssize_t on Windows (required by libusb)
    add_compile_definitions(_SSIZE_T_DEFINED ssize_t=int64_t)

    # Try to locate an existing Windows installation of libusb
    find_path(LIBUSB_INCLUDE_DIR
        NAMES libusb.h
        HINTS "C:/Program Files/libusb-1.0/include" "C:/Program Files (x86)/libusb-1.0/include"
        PATH_SUFFIXES libusb-1.0
    )

    find_library(LIBUSB_LIBRARY
        NAMES usb-1.0 libusb-1.0
        HINTS "C:/Program Files/libusb-1.0" "C:/Program Files (x86)/libusb-1.0"
    )

    # Nothing found locally -> fetch and build libusb via the community-maintained "libusb-cmake" wrapper,
    # since upstream libusb ships no native CMake build system for MSVC
    if(NOT LIBUSB_INCLUDE_DIR OR NOT LIBUSB_LIBRARY)
      message(STATUS "libusb-1.0 not found locally. Downloading and building from source via FetchContent...")

      set(LIBUSB_BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
      set(LIBUSB_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
      set(LIBUSB_BUILD_TESTING OFF CACHE BOOL "" FORCE)
      set(LIBUSB_INSTALL_TARGETS OFF CACHE BOOL "" FORCE)

      FetchContent_Declare(
        libusb_cmake
        GIT_REPOSITORY "https://github.com/libusb/libusb-cmake.git"
        GIT_TAG "v1.0.30-0"
      )
      FetchContent_MakeAvailable(libusb_cmake)

      # libusb-cmake exposes the target "libusb::usb-1.0" - alias it to "libusb::libusb" so downstream
      # CMakeLists.txt files in this project can keep linking against the same target name on every
      # platform (Unix, MinGW and now MSVC)
      if(NOT TARGET libusb::libusb)
        add_library(libusb::libusb ALIAS libusb::usb-1.0)
      endif()
 
      set(LIBUSB_FOUND TRUE)
    endif()
```

I need feedback and external testing to address this.

src/stlink-lib/remote.h
slyshykO: This patch works for me :

```patch
diff --git a/src/st-server/server.c b/src/st-server/server.c
index 03e892f..a898ac7 100644
--- a/src/st-server/server.c
+++ b/src/st-server/server.c
@@ -26,6 +26,7 @@
 
 #if defined(_WIN32)
 #include <windows.h>
+#include <win32_socket.h>
 #else
 #include <arpa/inet.h>
 #include <netinet/in.h>
diff --git a/src/stlink-lib/remote.c b/src/stlink-lib/remote.c
index 0810cff..8bf6e1b 100644
--- a/src/stlink-lib/remote.c
+++ b/src/stlink-lib/remote.c
@@ -29,6 +29,9 @@
  * probe; the server applies the supplied ap before each operation.
  */
 
+#if defined(_WIN32)
+#include <win32_socket.h>
+#endif
 #include "remote.h"
 
 #include "read_write.h"
diff --git a/src/stlink-lib/remote.h b/src/stlink-lib/remote.h
index 7522001..2c50054 100644
--- a/src/stlink-lib/remote.h
+++ b/src/stlink-lib/remote.h
@@ -31,7 +31,7 @@
 #include <errno.h>
 
 #if defined(_WIN32)
-#include <win32_socket.h>
+#include <windows.h>
 #else
 #include <unistd.h>
 #include <fcntl.h>
@@ -46,7 +46,6 @@
 #include <stlink.h>
 #include <stlink_backend.h>
 
-#include "remote.h"
 #include "read_write.h"
 #include "logging.h"
```

slyshykO:
> > @Nightwalker-87, sorry for the delay and my detachment from the project. Can you tell me if we still need to support building with MSVC, cause this branch doesn't support it? libusb wasn't downloaded.
> 
> I haven't tested MSVC and focussed on the original approach which was present before this was added.
> 
> @slyshykO Add the following code block in `Findlibusb.cmake` to see if this is applicable:
> 
> ```
> # Windows (native MSVC or MinGW without cross-compiling)
> elseif(MSVC OR (WIN32 AND NOT EXISTS "/etc/debian_version"))
>     # Fix for missing ssize_t on Windows (required by libusb)
>     add_compile_definitions(_SSIZE_T_DEFINED ssize_t=int64_t)
> 
>     # Try to locate an existing Windows installation of libusb
>     find_path(LIBUSB_INCLUDE_DIR
>         NAMES libusb.h
>         HINTS "C:/Program Files/libusb-1.0/include" "C:/Program Files (x86)/libusb-1.0/include"
>         PATH_SUFFIXES libusb-1.0
>     )
> 
>     find_library(LIBUSB_LIBRARY
>         NAMES usb-1.0 libusb-1.0
>         HINTS "C:/Program Files/libusb-1.0" "C:/Program Files (x86)/libusb-1.0"
>     )
> 
>     # Nothing found locally -> fetch and build libusb via the community-maintained "libusb-cmake" wrapper,
>     # since upstream libusb ships no native CMake build system for MSVC
>     if(NOT LIBUSB_INCLUDE_DIR OR NOT LIBUSB_LIBRARY)
>       message(STATUS "libusb-1.0 not found locally. Downloading and building from source via FetchContent...")
> 
>       set(LIBUSB_BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
>       set(LIBUSB_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
>       set(LIBUSB_BUILD_TESTING OFF CACHE BOOL "" FORCE)
>       set(LIBUSB_INSTALL_TARGETS OFF CACHE BOOL "" FORCE)
> 
>       FetchContent_Declare(
>         libusb_cmake
>         GIT_REPOSITORY "https://github.com/libusb/libusb-cmake.git"
>         GIT_TAG "v1.0.30-0"
>       )
>       FetchContent_MakeAvailable(libusb_cmake)
> 
>       # libusb-cmake exposes the target "libusb::usb-1.0" - alias it to "libusb::libusb" so downstream
>       # CMakeLists.txt files in this project can keep linking against the same target name on every
>       # platform (Unix, MinGW and now MSVC)
>       if(NOT TARGET libusb::libusb)
>         add_library(libusb::libusb ALIAS libusb::usb-1.0)
>       endif()
>  
>       set(LIBUSB_FOUND TRUE)
>     endif()
> ```
> 
> I need feedback and external testing to address this.

This didn't work. Along with libusb, we also need the pthread library, so adding only libusb didn't allow us to build the project with MSVC.
I successfully built the project with the support of vcpkg.
So, let's fix the include issues and push this PR.

a-michelis:
Hello all, I can test MSVC support and assist on it, but using the toolset is impossible for me due to lack of devices currently

src/stlink-lib/remote.h
Nightwalker-87: Thx. :+1: I will consider this in an upcoming commit.

Nightwalker-87:
> This didn't work. Along with `libusb`, we also need the `pthread` library, so adding only `libusb` didn't allow us to build the project with MSVC. I successfully built the project with the support of `vcpkg`. So, let's fix the include issues and push this PR.

Ok, if `vcpkg` is the cleaner way of getting this resolved, we should make use of it and also include this in this PR to ensure everything remains functional. I'd like to avoid Post-Release issues regarding installation in this context.

@a-michelis Thx for popping in. Please assist @slyshykO to allow for independent testing and updating the documentation for MSVC which you contributed earlier.

slyshykO (approved):

Nightwalker-87:
Proposal:

1) New file `vcpkg.json` in project root directory:

```
{
  "name": "stlink",
  "dependencies": [
    "libusb",
    "pthreads"
  ]
}
```

2) `cmake/modules/Findlibusb.cmake`

```
 # Windows (native MSVC or MinGW without cross-compiling)
 elseif(MSVC OR (WIN32 AND NOT EXISTS "/etc/debian_version"))
     # Fix for missing ssize_t on Windows (required by libusb)
     add_compile_definitions(_SSIZE_T_DEFINED ssize_t=int64_t)

     # Try to locate an existing Windows installation of libusb
     find_path(LIBUSB_INCLUDE_DIR
         NAMES libusb.h
         HINTS "C:/Program Files/libusb-1.0/include" "C:/Program Files (x86)/libusb-1.0/include"
         PATH_SUFFIXES libusb-1.0
     )

     find_library(LIBUSB_LIBRARY
         NAMES usb-1.0 libusb-1.0
         HINTS "C:/Program Files/libusb-1.0" "C:/Program Files (x86)/libusb-1.0"
     )
+
+    # Nothing found locally -> libusb does not ship a CMake config file (neither
+    # upstream nor via vcpkg), only a pkg-config (.pc) file, so fall back to the
+    # same pkg-config lookup already used below for other Unix-like systems.
+    # This also picks up a vcpkg-provided pkgconf/libusb installation when the
+    # vcpkg toolchain file is passed via -DCMAKE_TOOLCHAIN_FILE.
+    if(NOT LIBUSB_INCLUDE_DIR OR NOT LIBUSB_LIBRARY)
+        find_package(PkgConfig QUIET)
+        if(PKG_CONFIG_FOUND)
+            pkg_search_module(PC_LIBUSB QUIET libusb-1.0)
+        endif()
+
+        find_path(LIBUSB_INCLUDE_DIR
+            NAMES libusb.h
+            HINTS ${PC_LIBUSB_INCLUDE_DIRS}
+            PATH_SUFFIXES libusb-1.0
+        )
+
+        find_library(LIBUSB_LIBRARY
+            NAMES usb-1.0
+            HINTS ${PC_LIBUSB_LIBRARY_DIRS}
+        )
+    endif()
```

3) `doc/compiling.md`

```
 ### Installation

 1. Install `Build Tools for Visual Studio` from <https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022>
 2. Install `cmake` from <https://cmake.org/download/#latest> --> Binary distributions --> Windows x64 Installer<br />
    Ensure that you add cmake to the $PATH system variable when following the instructions by the setup assistant.
    Follow the installation instructions on the website.
 3. Fetch the project source files by running `git clone https://github.com/stlink-org/stlink.git` from the command-line (`cmd.exe`/`powershell.exe`)<br />
    or download and extract (`7zip`) the latest stlink `.zip` release from the Release page on GitHub.
+4. Install `vcpkg` (provides `libusb` and `pthreads` for MSVC, since neither ships a native MSVC build):
+   1. `git clone https://github.com/microsoft/vcpkg.git`
+   2. `cd vcpkg && .\bootstrap-vcpkg.bat`
+   3. `.\vcpkg install libusb pthreads --triplet x64-windows`

 ### Building

 1. Open the command-line (`cmd.exe`/`powershell.exe`) with administrator privileges
 2. Move to the `stlink` directory with `cd C:\$Path-to-your-stlink-folder$\`
 3. Create a new `build` subdirectory and move into it with `cd .\build`.
-4. Configure the project, using the following command: `cmake -G "Visual Studio 17 2022" .. -DCMAKE_BUILD_TYPE="Release"`
+4. Configure the project, using the following command:<br />
+   `cmake -G "Visual Studio 17 2022" .. -DCMAKE_BUILD_TYPE="Release" -DCMAKE_TOOLCHAIN_FILE="C:\$Path-to-your-vcpkg-folder$\scripts\buildsystems\vcpkg.cmake"`
 5. Build the project, using the following command: `cmake --build . --target ALL_BUILD`
```

What remains open/untested (I can't conclusively clarify this without a real MSVC environment):

* Whether `find_package(Threads REQUIRED)` in the main `CMakeLists.txt` is automatically covered by the vcpkg `pthreads` port, or whether an additional explicit `find_package(pthreads REQUIRED)` call is needed for that.
* Whether `pkg_search_module`, within the vcpkg toolchain context, correctly finds `PKG_CONFIG_PATH` pointing to `<vcpkg>\installed\x64-windows\lib\pkgconfig`, or whether `PKG_CONFIG_EXECUTABLE`/`PKG_CONFIG_PATH` also need to be set manually for that.

Could you investigate that?

slyshykO:
> Proposal:
> 
>     1. New file `vcpkg.json` in project root directory:
> 
> 
> ```
> {
>   "name": "stlink",
>   "dependencies": [
>     "libusb",
>     "pthreads"
>   ]
> }
> ```
> 
>     2. `cmake/modules/Findlibusb.cmake`
> 
> 
> ```
>  # Windows (native MSVC or MinGW without cross-compiling)
>  elseif(MSVC OR (WIN32 AND NOT EXISTS "/etc/debian_version"))
>      # Fix for missing ssize_t on Windows (required by libusb)
>      add_compile_definitions(_SSIZE_T_DEFINED ssize_t=int64_t)
> 
>      # Try to locate an existing Windows installation of libusb
>      find_path(LIBUSB_INCLUDE_DIR
>          NAMES libusb.h
>          HINTS "C:/Program Files/libusb-1.0/include" "C:/Program Files (x86)/libusb-1.0/include"
>          PATH_SUFFIXES libusb-1.0
>      )
> 
>      find_library(LIBUSB_LIBRARY
>          NAMES usb-1.0 libusb-1.0
>          HINTS "C:/Program Files/libusb-1.0" "C:/Program Files (x86)/libusb-1.0"
>      )
> +
> +    # Nothing found locally -> libusb does not ship a CMake config file (neither
> +    # upstream nor via vcpkg), only a pkg-config (.pc) file, so fall back to the
> +    # same pkg-config lookup already used below for other Unix-like systems.
> +    # This also picks up a vcpkg-provided pkgconf/libusb installation when the
> +    # vcpkg toolchain file is passed via -DCMAKE_TOOLCHAIN_FILE.
> +    if(NOT LIBUSB_INCLUDE_DIR OR NOT LIBUSB_LIBRARY)
> +        find_package(PkgConfig QUIET)
> +        if(PKG_CONFIG_FOUND)
> +            pkg_search_module(PC_LIBUSB QUIET libusb-1.0)
> +        endif()
> +
> +        find_path(LIBUSB_INCLUDE_DIR
> +            NAMES libusb.h
> +            HINTS ${PC_LIBUSB_INCLUDE_DIRS}
> +            PATH_SUFFIXES libusb-1.0
> +        )
> +
> +        find_library(LIBUSB_LIBRARY
> +            NAMES usb-1.0
> +            HINTS ${PC_LIBUSB_LIBRARY_DIRS}
> +        )
> +    endif()
> ```
> 
>     3. `doc/compiling.md`
> 
> 
> ```
>  ### Installation
> 
>  1. Install `Build Tools for Visual Studio` from <https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022>
>  2. Install `cmake` from <https://cmake.org/download/#latest> --> Binary distributions --> Windows x64 Installer<br />
>     Ensure that you add cmake to the $PATH system variable when following the instructions by the setup assistant.
>     Follow the installation instructions on the website.
>  3. Fetch the project source files by running `git clone https://github.com/stlink-org/stlink.git` from the command-line (`cmd.exe`/`powershell.exe`)<br />
>     or download and extract (`7zip`) the latest stlink `.zip` release from the Release page on GitHub.
> +4. Install `vcpkg` (provides `libusb` and `pthreads` for MSVC, since neither ships a native MSVC build):
> +   1. `git clone https://github.com/microsoft/vcpkg.git`
> +   2. `cd vcpkg && .\bootstrap-vcpkg.bat`
> +   3. `.\vcpkg install libusb pthreads --triplet x64-windows`
> 
>  ### Building
> 
>  1. Open the command-line (`cmd.exe`/`powershell.exe`) with administrator privileges
>  2. Move to the `stlink` directory with `cd C:\$Path-to-your-stlink-folder$\`
>  3. Create a new `build` subdirectory and move into it with `cd .\build`.
> -4. Configure the project, using the following command: `cmake -G "Visual Studio 17 2022" .. -DCMAKE_BUILD_TYPE="Release"`
> +4. Configure the project, using the following command:<br />
> +   `cmake -G "Visual Studio 17 2022" .. -DCMAKE_BUILD_TYPE="Release" -DCMAKE_TOOLCHAIN_FILE="C:\$Path-to-your-vcpkg-folder$\scripts\buildsystems\vcpkg.cmake"`
>  5. Build the project, using the following command: `cmake --build . --target ALL_BUILD`
> ```
> 
> What remains open/untested (I can't conclusively clarify this without a real MSVC environment):
> 
>     * Whether `find_package(Threads REQUIRED)` in the main `CMakeLists.txt` is automatically covered by the vcpkg `pthreads` port, or whether an additional explicit `find_package(pthreads REQUIRED)` call is needed for that.
> 
>     * Whether `pkg_search_module`, within the vcpkg toolchain context, correctly finds `PKG_CONFIG_PATH` pointing to `<vcpkg>\installed\x64-windows\lib\pkgconfig`, or whether `PKG_CONFIG_EXECUTABLE`/`PKG_CONFIG_PATH` also need to be set manually for that.
> 
> 
> Could you investigate that?

this is my fixes - https://github.com/slyshykO/stlink/commit/bbbd07b3aa8155cacf5c2cbb2d4646dec8a580ec#diff-a883cf46cbd8aa22e92cee2d199e16bf8c02d8d5edbadb9572f806175ca25ad8

Nightwalker-87:
As feared cross-compilation is not successful anymore. It now fails for x86-64.

```
CMake Error at cmake/modules/Findlibusb.cmake:114 (message):
  libusb make failed with code 2
Call Stack (most recent call first):
  src/stlink-env/CMakeLists.txt:53 (find_package)
```

a-michelis:
Tested `43db600` on native Windows with MSVC + vcpkg. Setup: Windows 11 Pro (26200), CMake 3.28.3, vcpkg 2026-07-27, VS 2022 Preview 17.14.36119.2 (MSVC 19.44) and VS 18 Build Tools 18.6.11806.211. No ST-LINK hardware here, so this is build/packaging only.

**MSVC + vcpkg works.** Configuring with `-G "Visual Studio 17 2022" -A x64` plus the vcpkg toolchain succeeds; vcpkg pulls libusb 1.0.30 and pthreads 3.0.0#14. Release and Debug both build with 0 errors and 0 warnings, and `st-info --probe` runs in both. @slyshykO's per-config libusb selection is correct: LIBUSB_LIBRARY_RELEASE and _DEBUG both resolve and each configuration links the matching lib, with the right pthreadVC3/pthreadVC3d DLL copied beside the exes.

---

**On the x86-64 cross-compilation failure at `Findlibusb.cmake:114`**

I think the cause is four lines above, at line 98:

```cmake
COMMAND ./configure --host=i686-w64-mingw${ARCH} --prefix=...
```

ARCH is 64 or 32 and is used in only this one place, so it expands to:

- ARCH=32 -> `i686-w64-mingw32` (valid)
- ARCH=64 -> `i686-w64-mingw64` (not a real triple)

The 64-bit triple is x86_64-w64-mingw32. The 32 in mingw32 refers to the Win32 API rather than the word size, so it should stay 32 while the prefix changes from i686 to x86_64. As written, the prefix is never switched.

That also explains why make fails rather than configure: with a `--host` whose `${host}-gcc` does not exist, autoconf falls back to the native gcc with a warning and exits 0, then make fails building libusb's Windows backend against native Linux headers. Error code 2 fits.

This predates the MSVC work (git log -L points at `dd2f03a` / `2be6fd7`), so it is probably not a regression from #1507. My guess is that splitting the chained COMMANDs into separate execute_process calls in `2be6fd7` is what finally lets the build reach make, exposing the bad triple. I have no Debian cross environment to confirm, but the one-line change to try is `--host=x86_64-w64-mingw32` for the 64-bit case.

---

**4 MSVC issues at `43db600`**

1.` gen_binaries_msvc.bat` fails when VS18 is installed but CMake is older than 4.2. The `vswhere [18.0,19.0)` probe matched my VS 18 Build Tools, so the script selected Visual Studio 18 2026 and CMake 3.28.3 aborted with "Could not create named generator". It does not fall back to the VS 17 generator that is installed and supported. The CMake 4.2 requirement is documented in prose but not enforced by the script.
2. `cmake --install` produces an installation that cannot run. Installed `bin/st-info.exe` imports `libusb-1.0.dll` and `pthreadVC3.dll`, and neither is installed, so the tools fail to start. `libusb.h` and `getopt.h` are also absent while the installed `libusb_settings.h` and `unistd.h` include them, which leaves `lib/stlink-static.lib` and the installed headers unusable downstream. v1.8.0 shipped `include/libusb-1.0/libusb.h` and `lib/libusb-1.0.lib`. Dropping the install step from `compiling.md` hides this, but the `INSTALL` target and cmake/packaging are still present.
3. Without the vcpkg toolchain, configure fails at `CMakeLists.txt:95` with "Could not find a package configuration file provided by PThreads4W" and no hint that vcpkg is needed. `find_package(PThreads4W QUIET)` plus an explicit `FATAL_ERROR` pointing at doc/compiling.md would be friendlier.
4. The `winsock v1`/`v2` conflict is latent rather than fixed. It works because libusb >= 1.0.30 dropped its `#include <winsock.h>` and the vcpkg port is exactly `1.0.30`. The MSVC `find_path`/`find_library` fallback can still select an older system libusb -- a previous stlink release's bundled copy did exactly that on my machine -- and then every TU where `libusb.h` precedes `win32_socket.h` fails with ~60 `C2011`/`C2375` redefinitions of `sockaddr`/`timeval`/`fd_set` and the socket functions. `c1763e0` addresses this per-TU by reordering includes, which has to be repeated for every new TU. Adding `-D_WINSOCKAPI_` to the `WIN32` definitions in `src/stlink-env/CMakeLists.txt` suppresses `winsock.h v1` globally and independently of include order, since `winsock2.h` guards on `WINSOCK2API` and defines `WINSOCKAPI` itself.

**On the two open questions**

- `find_package(Threads REQUIRED)` is not covered by the vcpkg pthreads port. @slyshykO's commit already resolves it: `FindThreads` accepts native Win32 threads while `usb.c` uses the pthread API. The port ships a `PThreads4W` CMake config and `find_package(PThreads4W REQUIRED)` works, resolving both configurations.
- `pkg_search_module` never runs on the vcpkg path: the new branch keys on `VCPKG_TARGET_TRIPLET` and reads `VCPKG_INSTALLED_DIR` directly, so `PkgConfig` is not invoked at all. The proposed `pkg-config` fallback is not needed here.

**A few words on the old build-from-source approach of `libusb`**

on Win32 there is no standard location for third-party libraries that aren't shipped through a platform mechanism, so build systems targeting Windows usually ship them directly. Anything else pushes the user into manually adding install directories to PATH. This repo already shows both halves of that: the find_path/find_library hints in Findlibusb.cmake invent a `C:/Program Files/libusb-1.0` convention rather than following one, and the libusb that find_library actually picked up on my machine was `C:/Program Files (x86)/stlink/include/libusb-1.0` - the copy bundled by our own v1.8.0 release. (That also caused the winsock clash in item 4, since the bundled copy is older than 1.0.30.)

To be clear, this is not an argument against vcpkg. vcpkg builds libusb and pthreads from source into the build tree and copies the DLLs next to the binaries, so it is per-project vendoring too - the difference is who curates the recipe, and outsourcing that to vcpkg is a reasonable trade - if we don't mind requiring users to install a separate package manager, essentially.

What I would like to keep from the older approach is the property that the shipped artifacts are self-contained. With statically vendored libusb, stlink.dll imported only `WSOCK32`/`WS2_32`/`KERNEL32` and the CRT, and the tools ran with nothing added to `PATH`. That property is what item 2 in the issues i'm bringing loses: the installed exes now import `libusb-1.0.dll` and `pthreadVC3.dll` and neither is installed. Whichever dependency source we settle on, I think `cmake --install` should keep producing a tree that runs as-is, and packaging under cmake/packaging depends on that too.

Nightwalker-87:
> This predates the MSVC work (git log -L points at `dd2f03a` / `2be6fd7`), so it is probably not a regression from #1507. My guess is that splitting the chained COMMANDs into separate execute_process calls in `2be6fd7` is what finally lets the build reach make, exposing the bad triple. I have no Debian cross environment to confirm, but the one-line change to try is `--host=x86_64-w64-mingw32` for the 64-bit case.

From what I can tell, it just worked again before one started to address the MSVC issue.

a-michelis:
> > This predates the MSVC work (git log -L points at `dd2f03a` / `2be6fd7`), so it is probably not a regression from #1507. My guess is that splitting the chained COMMANDs into separate execute_process calls in `2be6fd7` is what finally lets the build reach make, exposing the bad triple. I have no Debian cross environment to confirm, but the one-line change to try is `--host=x86_64-w64-mingw32` for the 64-bit case.
> 
> From what I can tell, it just worked again before one started to address the MSVC issue.

Inability to build due to arch errors is what prompted me to do the previous PR in the first place, - if my memory is intact, the zip archive downloaded by the older approach gave errors depending on the arch (some calculated zip path was off).

But maybe I was the one doing something wrong back then!

Nightwalker-87:
I have now removed any potential remnants of previous build approaches and restarted a clean build from scratch. Both variants now compile successfully.

https://pastebin.com/vbu5QSpq

a-michelis:
> I have now removed any potential remnants of previous build approaches and restarted a clean build from scratch. Both variants now compile successfully.
> 
> https://pastebin.com/vbu5QSpq

Worth noting though that your log shows the triple is still malformed and is just being papered over:

```
    32-bit: checking host system type... i686-w64-mingw32
            checking for i686-w64-mingw32-gcc... /usr/bin/i686-w64-mingw32-gcc

    64-bit: checking host system type... i686-w64-mingw64
            checking for i686-w64-mingw64-gcc... /usr/bin/x86_64-w64-mingw32-gcc
            checking for i686-w64-mingw64-g++... no
            configure: WARNING: using cross tools not prefixed with host triplet
```

Nightwalker-87:
@a-michelis: Can you re-investigate if there are any issues from your last comment which are still present now as it seems to work? I just want to make sure that what we are looking at currently is not an unstable state, based on any lucky coincidences.

Nightwalker-87:
> > I have now removed any potential remnants of previous build approaches and restarted a clean build from scratch. Both variants now compile successfully.
> > https://pastebin.com/vbu5QSpq
> 
> Worth noting though that your log shows the triple is still malformed and is just being papered over:
> 
> ```
>     32-bit: checking host system type... i686-w64-mingw32
>             checking for i686-w64-mingw32-gcc... /usr/bin/i686-w64-mingw32-gcc
> 
>     64-bit: checking host system type... i686-w64-mingw64
>             checking for i686-w64-mingw64-gcc... /usr/bin/x86_64-w64-mingw32-gcc
>             checking for i686-w64-mingw64-g++... no
>             configure: WARNING: using cross tools not prefixed with host triplet
> ```

This would subsequently result in:

```
# Windows-Build with MinGW via cross-compiling on Debian-Linux
elseif(MINGW AND EXISTS "/etc/debian_version")
    # Architecture: 64-bit or 32-bit?
    if (CMAKE_SIZEOF_VOID_P EQUAL 8)
        message(STATUS "=== Building for Windows (x86-64) ===")
        set(ARCH x86_64)
    else ()
        message(STATUS "=== Building for Windows (i686) ===")
        set(ARCH i686)
    endif ()
```

and

```
# Configure
execute_process(
    COMMAND ./configure --host=${ARCH}-w64-mingw32 --prefix=${libusb_BINARY_DIR}/install
                        --enable-static --disable-shared --disable-udev
    WORKING_DIRECTORY ${libusb_SOURCE_DIR}
    RESULT_VARIABLE CONFIGURE_RESULT
)
```

Nightwalker-87:
> **4 MSVC issues at `43db600`**
> 
> 1.` gen_binaries_msvc.bat` fails when VS18 is installed but CMake is older than 4.2. The `vswhere [18.0,19.0)` probe matched my VS 18 Build Tools, so the script selected Visual Studio 18 2026 and CMake 3.28.3 aborted with "Could not create named generator". It does not fall back to the VS 17 generator that is installed and supported. The CMake 4.2 requirement is documented in prose but not enforced by the script.
> 2. `cmake --install` produces an installation that cannot run. Installed `bin/st-info.exe` imports `libusb-1.0.dll` and `pthreadVC3.dll`, and neither is installed, so the tools fail to start. `libusb.h` and `getopt.h` are also absent while the installed `libusb_settings.h` and `unistd.h` include them, which leaves `lib/stlink-static.lib` and the installed headers unusable downstream. v1.8.0 shipped `include/libusb-1.0/libusb.h` and `lib/libusb-1.0.lib`. Dropping the install step from `compiling.md` hides this, but the `INSTALL` target and cmake/packaging are still present.
> 3. Without the vcpkg toolchain, configure fails at `CMakeLists.txt:95` with "Could not find a package configuration file provided by PThreads4W" and no hint that vcpkg is needed. `find_package(PThreads4W QUIET)` plus an explicit `FATAL_ERROR` pointing at doc/compiling.md would be friendlier.
> 4. The `winsock v1`/`v2` conflict is latent rather than fixed. It works because libusb >= 1.0.30 dropped its `#include <winsock.h>` and the vcpkg port is exactly `1.0.30`. The MSVC `find_path`/`find_library` fallback can still select an older system libusb -- a previous stlink release's bundled copy did exactly that on my machine -- and then every TU where `libusb.h` precedes `win32_socket.h` fails with ~60 `C2011`/`C2375` redefinitions of `sockaddr`/`timeval`/`fd_set` and the socket functions. `c1763e0` addresses this per-TU by reordering includes, which has to be repeated for every new TU. Adding `-D_WINSOCKAPI_` to the `WIN32` definitions in `src/stlink-env/CMakeLists.txt` suppresses `winsock.h v1` globally and independently of include order, since `winsock2.h` guards on `WINSOCK2API` and defines `WINSOCKAPI` itself.
 
> **A few words on the old build-from-source approach of `libusb`**
> 
> on Win32 there is no standard location for third-party libraries that aren't shipped through a platform mechanism, so build systems targeting Windows usually ship them directly. Anything else pushes the user into manually adding install directories to PATH. This repo already shows both halves of that: the find_path/find_library hints in Findlibusb.cmake invent a `C:/Program Files/libusb-1.0` convention rather than following one, and the libusb that find_library actually picked up on my machine was `C:/Program Files (x86)/stlink/include/libusb-1.0` - the copy bundled by our own v1.8.0 release. (That also caused the winsock clash in item 4, since the bundled copy is older than 1.0.30.)
> 
> To be clear, this is not an argument against vcpkg. vcpkg builds libusb and pthreads from source into the build tree and copies the DLLs next to the binaries, so it is per-project vendoring too - the difference is who curates the recipe, and outsourcing that to vcpkg is a reasonable trade - if we don't mind requiring users to install a separate package manager, essentially.
> 
> What I would like to keep from the older approach is the property that the shipped artifacts are self-contained. With statically vendored libusb, stlink.dll imported only `WSOCK32`/`WS2_32`/`KERNEL32` and the CRT, and the tools ran with nothing added to `PATH`. That property is what item 2 in the issues i'm bringing loses: the installed exes now import `libusb-1.0.dll` and `pthreadVC3.dll` and neither is installed. Whichever dependency source we settle on, I think `cmake --install` should keep producing a tree that runs as-is, and packaging under cmake/packaging depends on that too.

Please fix this properly without further complicating or changing the existing routines.
The intention was to simplify and streamline the compilation in terms of refactoring along with fixing persistent bugs.
**We need to avoid moving in the opposite direction again...**

Here are my thoughts on this:

**MSVC support policy**

1. No special-casing. Support is pinned to the current `libusb`/`CMake` compatibility baseline. Which `MSVC` versions are considered compatible follows Microsoft's own support lifecycle for `MSVC`, together with the current minimum system requirements of `CMake` and `libusb` (or version availability in `vcpkg`). If a given `MSVC` version doesn't fit that baseline, it's simply not supported - versions that fall out of support are dropped immediately, no fallback logic.
2. Winsock migration: Wherever `winsock.h` (legacy API) is currently included, replace it with `winsock2.h` — a straight API swap, nothing else changed at each call site for now. Target compatibility: Windows 10+.
3. MSVC + vcpkg only: The only supported combination is `MSVC` with the `vcpkg` toolchain. This is documented as a hard requirement, so no runtime detection or fallback checks are needed.
4. Should this approach result in notable extra efforts in maintenance for the upcoming months and years due to external circumstances, support will be cut down again.

Scope note: the `MinGW` cross-compile build (the actual purpose of this PR) stays exactly as self-contained as before — statically linked, single executable, nothing added to `PATH`. `MSVC` users accept vcpkg's vendoring model as part of choosing that toolchain (e.g. via `VCPKG_APPLOCAL_DEPS` to place DLLs next to the binaries).

a-michelis:
> This would subsequently result in:
> 
> ```
> # Windows-Build with MinGW via cross-compiling on Debian-Linux
> elseif(MINGW AND EXISTS "/etc/debian_version")
>     # Architecture: 64-bit or 32-bit?
>     if (CMAKE_SIZEOF_VOID_P EQUAL 8)
>         message(STATUS "=== Building for Windows (x86-64) ===")
>         set(ARCH x86_64)
>     else ()
>         message(STATUS "=== Building for Windows (i686) ===")
>         set(ARCH i686)
>     endif ()
> ```
> 
> and
> 
> ```
> # Configure
> execute_process(
>     COMMAND ./configure --host=${ARCH}-w64-mingw32 --prefix=${libusb_BINARY_DIR}/install
>                         --enable-static --disable-shared --disable-udev
>     WORKING_DIRECTORY ${libusb_SOURCE_DIR}
>     RESULT_VARIABLE CONFIGURE_RESULT
> )
> ```

Tested your patch as written, in a MinGW cross env (Ubuntu 24.04, mingw-w64) at `43db600`.

Both variants resolve correctly and both build clean:

```
64-bit: checking host system type... x86_64-w64-mingw32
        checking for x86_64-w64-mingw32-gcc... /usr/bin/x86_64-w64-mingw32-gcc
        checking for x86_64-w64-mingw32-g++... x86_64-w64-mingw32-g++

32-bit: checking host system type... i686-w64-mingw32
        checking for i686-w64-mingw32-gcc... /usr/bin/i686-w64-mingw32-gcc
        checking for i686-w64-mingw32-g++... i686-w64-mingw32-g++
```

Artifacts: `PE32+ executable x86-64` and `PE32 executable Intel 80386`.

It also clears the native g++ fallback that the baseline had (`checking for i686-w64-mingw64-g++... no`, then `checking for g++... g++`, which then got paired with the mingw `ld`).

One note, take it or leave it: deriving the arch from `CMAKE_SIZEOF_VOID_P` and hardcoding `-w64-mingw32` can still drift from `TOOLCHAIN_PREFIX`, which is what actually selects the compiler. `--host=${TOOLCHAIN_PREFIX}` cannot diverge and adds no new requirement, since `set_toolchain.cmake` already derives `CMAKE_C_COMPILER` from it (if unset, the compiler becomes `-gcc` and configure fails at detection anyway). Raising it only because the original bug was precisely `--host` disagreeing with the compiler.

That variant, also tested on both architectures with the results above:

```diff
         # Configure
         execute_process(
-            COMMAND ./configure --host=i686-w64-mingw${ARCH} --prefix=${libusb_BINARY_DIR}/install
+            COMMAND ./configure --host=${TOOLCHAIN_PREFIX} --prefix=${libusb_BINARY_DIR}/install
                                 --enable-static --disable-shared --disable-udev
             WORKING_DIRECTORY ${libusb_SOURCE_DIR}
```

`ARCH` then has no remaining reader, so the block above collapses into a status line that names the target more precisely than a bare 32/64 (cosmetic, not part of what I tested):

```diff
 # Windows-Build with MinGW via cross-compiling on Debian-Linux
 elseif(MINGW AND EXISTS "/etc/debian_version")
-    # Architecture: 64-bit or 32-bit?
-    if (CMAKE_SIZEOF_VOID_P EQUAL 8)
-        message(STATUS "=== Building for Windows (x86-64) ===")
-        set(ARCH 64)
-    else ()
-        message(STATUS "=== Building for Windows (i686) ===")
-        set(ARCH 32)
-    endif ()
+    message(STATUS "=== Building for Windows (${TOOLCHAIN_PREFIX}) ===")
 
     # Download and build libusb via FetchContent
     if(NOT LIBUSB_FOUND)
```

---

**Note**: I'll be sharing my thoughts on MSVC tonight, once i get back from work

a-michelis:
I finally had some time to compose myself, so here's the short version of where I stand in regards to the MSVC policy and Windows in general.

I think uniformity is better arrived at through abstraction than through rulesets. And I do not think the problem here is MSVC. Windows support as a whole is patched rather than designed: hand-written POSIX shims, a socket layer we maintain ourselves, Windows release artifacts produced on Linux. MSVC is only where that shows up most clearly, and dropping it would remove none of it. What I would want is for Windows to be decided properly, in code as much as in the build configuration, one way or the other. Half-supported is an expensive state maintenance-wise. 

To be fair and honest: I'm extremely grateful that Windows support exists nonetheless - it would be ungrateful, naive and idiotic of me to expect the maintainers to have an arsenal of build machines, knowledge of all OSes and 25+ hours availability per day (let alone budget/compensation). I just want what we all want here: to make the project better and easier to advance. And what I'm sharing here is not an expectation I'm having from you, it's a discussion on how we could work on our shortcomings. Looking at my own contribution/involvement pattern, I am not consistent enough to be the person who maintains Windows support here, and less so now that I am working on a C++ port of the project. But I'm willing to attempt to commit to a consistent involvement interval, so I can own what I say and build what I think, instead of pointing fingers and preaching methodologies.

As a closing note, I believe the MSVC discussion should continue in a separate issue or PR, as the scope of this one is not MSVC support at its core.

---

Update: I sat on it for a while. Started working *proper* Windows support (that affects both MSVC and MinGW). I did some reading on WinUSB (which both msvc and mingw use as a backend to their libusb implementation) and tried to understand the `pthread` surface better. It's not necessarily to be accepted, it just goes to show my vision of the project as it stands. 

Will dig in the review in a bit.

Nightwalker-87:
@a-michelis Thx for your thoughts on this.
To me this would very well align with the suggested policy above as it would for now resolve all 4 points you mentioned earlier regarding your MSVC findings and likely also with limited efforts. Additionally this would round this topic up here in the sense of pushing potential further discussion behind the upcoming release, as you suggested. To me this sounds like a reasonable approach that would give something back to all perspectives mentioned so far.

Nightwalker-87:
I will include your `${TOOLCHAIN_PREFIX}` suggestion in the next commit.

Nightwalker-87:
BTW: We still need to reactivate and update the GH CI `c-cpp.yml` for MSVC.

src/stlink-env/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+endif()
+
+# ====================================================================================================
+
+### Propagate system-specific additional definitions, sources, headers and libraries to parent scope
+set(STLINK_SOURCES ${STLINK_SOURCES} PARENT_SCOPE)
+set(STLINK_HEADERS ${STLINK_HEADERS} PARENT_SCOPE)
+set(STLINK_DEFINITIONS ${STLINK_DEFINITIONS} PARENT_SCOPE)
+set(STLINK_COMPILE_OPTIONS ${STLINK_COMPILE_OPTIONS} PARENT_SCOPE)
+set(STLINK_INCLUDE_DIRS ${STLINK_INCLUDE_DIRS} PARENT_SCOPE)
```
a-michelis, line 125: Are we sure we don't also want `SSP_LIB` in here (with `PARENT_SCOPE` flag) ? I see root CMakeLists.txt using it while not setting it

CMakeLists.txt
```diff
@@ -452,12 +372,13 @@ endif()
 # Additional build tasks
 ###
 
-# MCU configuration files
+# MCU chip configuration files
 if (WIN32)
-set(CMAKE_CHIPS_DIR ./config/chips) ### TODO: no absolute path (!)
+    set(CMAKE_CHIPS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/config/chips)
```
a-michelis, line 377: I can see what this fixes. `gen_binaries_msvc.bat` never runs `cmake --install` or `cpack`, and `doc/compiling.md` now tells you to run the executable from the repository root, so an absolute source path makes that flow work where `./config/chips` only worked if you happened to be standing in the right directory.

The problem is that this also becomes a compile-time macro (:382), and on windows that value is now the source tree, which means: Each time the toolset wants to initialize chip ids database, it'd search directly on the source code's path (which is not there in the releases)

Additionally, I think this makes :384 install chip files onto themselves (?)

CMakeLists.txt
```diff
@@ -79,218 +80,105 @@ include(${CMAKE_MODULE_PATH}/get_version.cmake) # Determine project version
...
+include_directories(src/stlink-lib)
+include_directories(inc)                        # contains top-level header files
 
+# Project version file
+include_directories(${PROJECT_BINARY_DIR}/inc)  # contains version.h
+configure_file("${PROJECT_SOURCE_DIR}/inc/version.h.in"
+               "${CMAKE_BINARY_DIR}/inc/version.h")
 
+# Define source- and headerfiles for stlink library
+file(GLOB STLINK_LIB_HEADERS
```
a-michelis, line 152: This glob decides what we install, and it doesn't cover `src/stlink-env`, which is where `libusb_settings.h` lives now.

The problem is that the old copy at `inc/libusb_settings.h` is still matched, which means: we compile one copy and ship the other. On `testing` both were globbed and `src/stlink-lib` came second, so it overwrote the `inc` one and they happened to match.

The two are identical everywhere except FreeBSD, so nothing shows. There, the copy we compile accepts the older libusb that FreeBSD ships and the copy we install doesn't, so anyone building their own tool against our headers gets `#error unsupported libusb version` (which worked before)

Additionally, I think `inc/libusb_settings.h` can go altogether, nothing includes it anymore

CMakeLists.txt
```diff
@@ -391,45 +280,76 @@ install(TARGETS ${STLINK_LIB_STATIC} ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
...
+install(TARGETS st-trace  RUNTIME_DEPENDENCY_SET stlink_deps DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(TARGETS st-server RUNTIME_DEPENDENCY_SET stlink_deps DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(TARGETS st-flash  RUNTIME_DEPENDENCY_SET stlink_deps DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(TARGETS st-util   RUNTIME_DEPENDENCY_SET stlink_deps DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+if (MSVC)
+    # vcpkg links libusb/pthreads as DLLs on Windows; without this, an
+    # installed exe imports libusb-1.0.dll/pthreadVC3.dll that were never
+    # installed and fails to start (see PR discussion, issue #2).
+    install(RUNTIME_DEPENDENCY_SET stlink_deps
```
a-michelis, line 347: The excludes look right to me too: `api-ms-` and `ext-ms-` catch the UCRT forwarders before resolution and the System32 one catches `VCRUNTIME140.dll`, so we don't drag the redistributable along.

One thing to check: resolution here has no `DIRECTORIES`, so it only finds a DLL already sitting next to the exe, which is vcpkg's applocal copy. That's an MSBuild behaviour, so a Ninja configure wouldn't get it, and `cmake --install` fails outright rather than skipping (no `UNRESOLVED_DEPENDENCIES_VAR`). I haven't tried Ninja here so I may be wrong about it.

Additionally, `stlink-shared` (:219) doesn't carry the set, so the dll's own dependencies aren't covered. Works by accident right now, since the tools link the static lib and need the same files in the same place

src/stlink-env/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+    if (STLINK_HAVE_DIRENT_H)
+        list(APPEND STLINK_DEFINITIONS -DSTLINK_HAVE_DIRENT_H)
+    else ()
+        message(FATAL_ERROR "dirent.h not found! init_chipids can not be compiled for Unix/Linux.")
+    endif()
+else () # WIN32
+    CHECK_INCLUDE_FILE(windows.h STLINK_HAVE_WINDOWS_H)
+    message(STATUS "Windows platform detected. Using Windows API for init_chipids.")
+    list(APPEND STLINK_DEFINITIONS -DSTLINK_HAVE_WINDOWS_H)
+    # use GetModuleFileNameA and FindFirstFileA
```
a-michelis, line 115: Note to consider in the future:

This says `GetModuleFileNameA`, but only `FindFirstFileA` exists, at `src/stlink-lib/chipid.c` line 270. `GetModuleFileNameA` isn't anywhere in the tree. It reads as if the executable-relative lookup is already implemented, which is part of why an absolute chips path at `CMakeLists.txt` line 377 looks safe when it isn't. I'd either drop the half that isn't true, or implement it, and implementing it is probably its own PR.

CMakeLists.txt
```diff
@@ -5,6 +5,7 @@
 cmake_minimum_required(VERSION 3.20)
```
a-michelis, line 5: Unrelated to this PR, but the floor here is already wrong: CMAKE_C_STANDARD 17 (:18) needs cmake 3.21, and on 3.20 it fails at project(). RUNTIME_DEPENDENCY_SET (:337) needs 3.21 too, so this could just say 3.21.

a-michelis (commented):
One thing that isn't about a specific line: the MSVC job in `.github/workflows/c-cpp.yml` is still commented out at lines 188 to 190, so what actually runs is `linux_64_bit`, `macos_build` and `mingw_cross_compile`. This PR adds `vcpkg.json`, `gen_binaries_msvc.bat`, a `PThreads4W` dependency and a rewritten MSVC section in `doc/compiling.md`, and nothing in CI touches any of it.

a-michelis:
> BTW: We still need to reactivate and update the GH CI `c-cpp.yml` for MSVC.

Oh, GH didn't show me this comment, so my review summary is outdated 😝

a-michelis:
Two things I ran into while testing this on MSVC. Both are branches on my fork rather than PRs, so take them or leave them.

The first one is that 61f16f0 doesn't build here with MSVC at all: `flash_loader.c` fails with 58 errors, all of them `winsock2.h` redefining what `winsock.h` already defined.  `libusb.h` includes `<windows.h>` and then `<winsock.h>` (unconditionally) on Windows, while `sys_time.h` brings in `<winsock2.h>`, and the two can't live in the same translation unit. `flash_loader.c` is the only one that reaches `flash_loader.h` before `helper.h`, so it's the only one that gets them in the losing order. Including `winsock2.h` ahead of `libusb.h` sets `_WINSOCKAPI_` and makes libusb's `winsock.h` a no-op whichever way round they arrive.

Chasing that also explains my review comment about the glob better than I did at the time. We carry two copies of `libusb_settings.h`, and the two compilers don't pick the same one: gcc compiles the `src/stlink-env` copy, MSVC compiles the `inc` one. The branch deletes `inc/libusb_settings.h` and adds `src/stlink-env/*.h` to the install glob, which also makes the installed header match the compiled one.

The second one wraps the three pthread calls in `usb.c` behind a small platform wrapper, so that neither `PThreads4W` on MSVC nor winpthread on MinGW is needed: one header and one implementation per platform, picked in cmake. `vcpkg.json` drops to `libusb` alone, and the MinGW binaries stop importing `libwinpthread-1.dll`, which nothing was packaging anyway. It sits on top of the winsock branch.

Both are tested on Linux with gcc, clean under `-Wall -Wextra`, and on MSVC with vcpkg where the tree now builds with no errors or warnings and `st-info.exe` runs.

https://github.com/stlink-org/stlink/compare/61f16f0...a-michelis:libstlink:8b408d6

https://github.com/stlink-org/stlink/compare/61f16f0...a-michelis:libstlink:6628b8a

Happy to open either as a PR if that's easier for you, or for you to just cherry-pick.

CMakeLists.txt
slyshykO: I think we should always search chip configs in the relative path, not only in **CMAKE_CHIPS_DIR**

a-michelis:
Unrelated to this PR, but hot: `st-info --probe` doesn't reliably find more than 1 adapter.

With a V2 and a V3 attached on 61f16f0, 20 sequential runs gave me 0,1,2,0,2,0,0,2,2,1,2,2,1,0,0,2,2,2,2,0. 10 of 20 reported both, and 6 reported none at all. libusb logs `Access is denied` for whichever device lost that run.

The cause looks like 6622e2c, which made probing parallel. `stlink_probe_usb_devs` identifies every device and reads every serial, then hands each worker only the serial string and lets it rediscover the device through `stlink_open_usb`. So every worker walks the whole device list opening adapters to read their serials, including the ones it isn't looking for, and 2 workers end up fighting over the same device. When a worker is denied the device it actually wants, it skips it and returns nothing, which is why some runs find 0 rather than 1.

I checked that it's the parallelism and not something local to my setup: same commit, same hardware, same binary, with the only change being to call `stlink_probe_worker` inline instead of through a thread. That gives 20/20.

The fix probably isn't to drop the parallelism, it's to stop the workers rediscovering what the caller already knows. `stlink_probe_usb_devs` is holding every device by the time it spawns anything, so handing that to the worker instead of a serial string removes the scan, and the contention with it.

Happy to move this to its own issue if you'd rather keep the thread clean.

CMakeLists.txt
a-michelis: In windows, that might actually help! but in *nix that separate bin/lib/inc/share and have standardized locations for each, this might introduce unnecessary delays (ie scanning an enormous bin directory)

slyshykO: If we build a packet, yes, Linux has standardised folders for apps, libs, and data. But many users want to build the latest master to get support for newly added features and want to put it in a user folder or in /opt

a-michelis: Hmmm... We have 3 ways of "obtaining" an executable.

- `cmake --build . --target all`
- `cmake --build . --target install` + configuration with `--install-prefix`
- downloading the release zip.

Aggreed that the most probable way for the release zip to work is with relative paths ( `.`, `../share/stlink/chips`)
`install` will probably have it working as well. building (for debugging purposes, for example), tho needs to be special case. I don't know if `add_definition` takes `<build_interface>` and `<install_interface>` (i suspect it doesn't) and guess we wouldn't want to ship the build path hardcoded to a release. maybe debug vs release would do the trick... 

Will dig into it!

CMakeLists.txt
Nightwalker-87: Fixed.

Nightwalker-87:
> Both are tested on Linux with gcc, clean under `-Wall -Wextra`, and on MSVC with vcpkg where the tree now builds with no errors or warnings and `st-info.exe` runs.
> 
> [61f16f0...a-michelis:libstlink:8b408d6](https://github.com/stlink-org/stlink/compare/61f16f0...a-michelis:libstlink:8b408d6)
> 
> [61f16f0...a-michelis:libstlink:6628b8a](https://github.com/stlink-org/stlink/compare/61f16f0...a-michelis:libstlink:6628b8a)
> 
> Happy to open either as a PR if that's easier for you, or for you to just cherry-pick.

@a-michelis I have included the two changes locally with minor variations.

CMakeLists.txt
Nightwalker-87: The variant in `inc/libusb_settings.h` was actually the one with the intended content, it should have moved to `stlink-env/libusb_settings.h` already by now and it needs to be included, of course. This was my mistake.
The including has already been fixed in the two suggested patches.

src/stlink-env/CMakeLists.txt
Nightwalker-87: Deleted `# use GetModuleFileNameA and FindFirstFileA` completely.

Nightwalker-87:
Could we please agree on not posting any further unrelated findings into this PR?

I suggest to collect such locally for now and contribute these to a new ticket in a well structured way once the new release has been published. I'd prefer to not have (several) new tickets around as long as one tries to finalise v1.8.1.

The absolute maximum I see for this PR is to resolve the open conversations and a final check afterwards to ensure there are no missing leftovers from the intended fixes. This PR becomes to bloated.

src/stlink-env/CMakeLists.txt
Nightwalker-87: Added `set(SSP_LIB ${SSP_LIB} PARENT_SCOPE)` to the list.

CMakeLists.txt
Nightwalker-87:
```
if (MSVC)
    # vcpkg links libusb as a DLL on Windows; without this, an installed exe imports libusb-1.0.dll
    # that was never installed and fails to start.
    #
    # Resolution does not rely solely on VCPKG_APPLOCAL_DEPS having already copied the DLL next
    # to the exe at build time (that in turn relies on applocal.ps1 finding dumpbin/objdump
    # on PATH, which is not guaranteed outside a VS developer shell, e.g. plain Ninja builds);
    # the vcpkg triplet's own bin directory is searched directly as a second source.
    # Anything still unresolved is reported instead of aborting the install.

    install(RUNTIME_DEPENDENCY_SET stlink_deps
        PRE_EXCLUDE_REGEXES "api-ms-" "ext-ms-"
        POST_EXCLUDE_REGEXES ".*[Ss][Yy][Ss][Tt][Ee][Mm]32.*"
        DIRECTORIES "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}$<$<CONFIG:Debug>:/debug>/bin"
        UNRESOLVED_DEPENDENCIES_VAR stlink_unresolved_deps
        DESTINATION ${CMAKE_INSTALL_BINDIR}
    )
    install(CODE "
        if(stlink_unresolved_deps)
            message(WARNING \"cmake --install: unresolved runtime dependencies, not installed: \${stlink_unresolved_deps}\")
        endif()
    ")
endif()
```

a-michelis:
I enabled the MSVC job in CI on my fork so that 6ce04af gets checked on every platform at once, and it turned up a few things worth seeing before this merges. The Windows side now builds and packages, which means the packages themselves can be inspected.

https://github.com/a-michelis/libstlink/actions/runs/35185825863

The good part first: the runtime dependency fix works. `libusb-1.0.dll` lands in `bin/` in both the install tree and the zip, so an installed `st-info.exe` starts on a machine that has never seen vcpkg. The thread wrapper works too - the MinGW binaries import only `KERNEL32`, `msvcrt`, `WS2_32` and `WSOCK32`, so `libwinpthread-1.dll` is gone. MSVC builds clean on x64 and x86, tests included.

The one I would want fixed before any release is the chips directory. Both packages carry the build machine's own path:

```
stlink-1.8.0-win32/a/libstlink/libstlink/config/chips/...
stlink-1.8.0-i686-w64-mingw32/home/runner/work/libstlink/libstlink/config/chips/...
```

There is no copy anywhere else in either archive, and `STLINK_CHIPS_DIR` is compiled in as that same path, so the chip files ship at a location the binary will never look at. Anyone unzipping this gets "Can't find any chip description file" for every target.

Two smaller ones from the same run. 

- The x64 package is named `stlink-1.8.0-win32.zip`, because `CPACK_PACKAGE_FILE_NAME` hardcodes `win32` for native Windows - so x64 and x86 produce the same filename with different contents, while the cross-build branch right below gets this right with `${TOOLCHAIN_PREFIX}`. 
- The install now places `usb.h` and `unistd.h` in the consumer include directory, where they shadow the Windows SDK's `usb.h` and MinGW's own `unistd.h` for anyone who puts that directory on their include path.
- On libusb, `Findlibusb.cmake` states that anything before 1.0.30 is no longer supported on MSVC, which is clear enough, but `libusb_settings.h` still gates at `0x01000108`, which is 1.0.24. I ran into that with 1.0.29 here: it passes the version check and then fails with 232 winsock errors, instead of the message that file is trying to give.

---

**Updates:**

I made an effort on the chips dir. As @slyshykO said, it'd probably be better to relative-search for the chips dir.

So, I split the implementation to windows and linux (like with pthreads) and gave the function a 3-way search:

- If we're building in debug, the first thing we do is to search in the source directory for the chips (og approach of Win32, but now for all OSes). If not found, we go on to the next.
- If `STLINK_CHIPS_DIR` is defined as an environment variable, we search this before we go on.
- We (finally) search `${ExecutablePath}/../share/stlink/config/chips` and `${ExecutablePath}/chips`, before we determine that the chips are not found.

The path argument becomes optional, and when it is given it is checked instead of any of the above. If the caller named a directory, they probably have something there they want to test, so we don't quietly fall back to the usual places - we read that directory or we error out.

Nothing is compiled in any more except the source path, and that one only exists in a debug build, so a release binary carries no build machine paths at all.

CI: Green, Branch at https://github.com/a-michelis/libstlink/tree/fix/chips_discovery

Nightwalker-87:
@a-michelis Thx. Could you push the enabled MSVC CI and the chip file path fix via cherry-pick (two separate commits) directly to this PR?
I would then ask for a review of the current fixes and we may open a new PR for the leftovers mentioned above.

CMakeLists.txt
Nightwalker-87: @a-michelis Is this ok now?

a-michelis:
Available in #1508 -I wanted someone to first check the commits instead of plainly adding them myself

CMakeLists.txt
Nightwalker-87: Transferred to #1508.

Nightwalker-87:
@slyshykO @klonyyy @a-michelis Please review this PR again regarding the included topics.
The remaining open points will be addressed in #1508.

CMakeLists.txt
a-michelis: Yes, LGTM!

a-michelis (approved):
All good from my side - everything I raised here is addressed, and the chips one is covered by #1508. CI that was in my previous summary is also covered by #1508 

Just a minor nudge about the `libusb` versioning, since it's on a line this PR touches: `libusb_settings.h` sets `MINIMAL_API_VERSION` to `0x01000108`, which is 1.0.24, while `Findlibusb.cmake` says anything before 1.0.30 is no longer supported on MSVC. I ran into that with 1.0.29 here - it passes the version check and then fails with 232 winsock errors, instead of the message that file is trying to give. Could go into #1508, or be left alone for now - I have WinUSB work lined up that would take libusb out of Windows builds, which changes the picture here. That'll come as its own PR.

(Btw, i accidentally edited your comment, and reverted it back afterwards)

slyshykO (approved):

PR #1507, opened 2026-09-13
Fixed MSVC compatibility: Use vcpkg with MSVC

slyshykO:


src/stlink-env/CMakeLists.txt
```diff
@@ -51,7 +51,6 @@ endif()
 ### libusb
 
 find_package(libusb REQUIRED)
-list(APPEND STLINK_DEFINITIONS ${LIBUSB_DEFINITIONS}) # define ssize_t: only affects WIN32, MSVC & MINGW
```
Nightwalker-87, line 54: This seems to break `set(STLINK_DEFINITIONS ${STLINK_DEFINITIONS} PARENT_SCOPE)`.

CMakeLists.txt
```diff
@@ -90,19 +90,21 @@ cmake_host_system_information(RESULT OS_NAME QUERY OS_NAME)
 message(STATUS "Checking for OS_NAME: ${OS_NAME}")
 
 # Threading support
-find_package(Threads REQUIRED)
-set(THREADS_PREFER_PTHREAD_FLAG ON) # pthreads on Unix, winpthread on WIN32, MSVC & MINGW
```
Nightwalker-87, line 94: Please preserve the inline-comment
_pthreads on Unix, winpthread on WIN32, <del>MSVC</del> & MINGW_
at the new line below.

cmake/modules/Findlibusb.cmake
```diff
@@ -41,9 +61,6 @@ elseif(MSVC OR (WIN32 AND NOT EXISTS "/etc/debian_version"))
 
 # Windows-Build with MinGW via cross-compiling on Debian-Linux
 elseif(MINGW AND EXISTS "/etc/debian_version")
-    # Fix for ssize_t on Windows
```
Nightwalker-87, line 44: Not sure if this affects MinGW compilation on Windows or cross-compiling...

src/stlink-env/CMakeLists.txt
slyshykO: STLINK_DEFINITIONS is still exported unchanged. This line only appended LIBUSB_DEFINITIONS, which our find module never defines. The ssize_t compatibility definition now lives in win32/unistd.h. Before/after checks confirmed identical parent-scope definitions.

Nightwalker-87 (changes requested):
Please check the respective comments.

src/stlink-env/CMakeLists.txt
Nightwalker-87: Sorry, this was my mistake. Should have read properly.

cmake/modules/Findlibusb.cmake
Nightwalker-87: We can try it after merging this into the `fix/win_binaries` branch.

slyshykO: > Not sure if this affects MinGW compilation on Windows or cross-compiling...

I've checked compilation with msys/gcc and in WSL Debian. All compile.

CMakeLists.txt
slyshykO: add in new commit

Nightwalker-87: Wait - `WIN32` without `MSVC` falls into the else case. The same applies for `MINGW`.

slyshykO: Actually, mingw uses winpthread but it masks it for cmake and other builds systems

Nightwalker-87: I understand. - To be honest, this has never really been pointed out over the years, though `MinGW` is in use for quite a while now within this project.

Nightwalker-87 (approved):

PR #1508, opened 2026-09-17
Fixes for MSVC compatibility & CI for MSVC

a-michelis:
I don't have write access to the repo, so I couldn't push to `fix/win_binaries` directly.

Please rebase-and-merge rather than squash, so the two commits stay separate as you asked.

---

**Enable the Windows MSVC job in CI**

- Dropped the choco install of Visual Studio, and the pinned generator name
- Added the vcpkg toolchain file
- Debug configured with `CMAKE_BUILD_TYPE` so CTest is included
- Steps and layout follow the Linux and macOS jobs, install and package included

**Resolve the chip description directory at run time**

- Install destination made relative
- Lookup order: caller's directory (exclusive), source tree in a debug build only, `STLINK_CHIPS_DIR`, then `<exe>/../share/stlink/config/chips` and `<exe>/chips`
- Source path is the only thing compiled in, and only in a debug build
- Executable location and directory walking moved to `src/stlink-env/fs/`, one implementation per platform
- `process_chipfile` unchanged, still in `chipid.c`

Tested on Linux, MSVC and a MinGW cross build run on Windows; green on all 13 jobs on my fork.


Nightwalker-87:
> I don't have write access to the repo, so I couldn't push to `fix/win_binaries` directly.

You have, but nevertheless we will leave it here now.

Nightwalker-87:
@a-michelis I have reviewed the current state and will also do test-compiling on my machine tonight.
You can add a bugfix for this https://github.com/stlink-org/stlink/pull/1504#issuecomment-5677937652 here as well if you like _and_ if it's not too much work. As it reads you already spotted the source of the problem here.

Have I missed anything else left that turns out to be affective and/or urgent?
Otherwise one should focus on the G0/G4 issue...

a-michelis:
Thanks, @Nightwalker-87. I tried to make the changes as minimal/less-invasive as possible so it's still readable.

> You can add a bugfix for this https://github.com/stlink-org/stlink/pull/1504#issuecomment-5677937652 here as well

I can, but bear 2 things in mind:
1. The bug is probably not Windows-only, it affects all platform. 
2. My proposal (which we get to decide here) is either :
   - Remove parallelism on usb operations (as USB libs don't like them). This does *fix* the problem, but removes the previously merged PR
   - or remove the number of operations by passing everything the initial scan provided to the workers. This makes this operation go from "hopefully works 50% of the time" to "It may fail sometimes, but it's mostly OK".  

Will dive into the problem tomorrow to see if there are thread-safe alternatives i can leverage so we keep parallelism.

Nightwalker-87:
If this is likely a platform independent issue, we should better not include it here.
Reading your recent comment gave me the impression this was windows-specific.
Keep it in mind though, it should be fixed anyway sooner or later.

Nightwalker-87 (approved):
Native compiling and cross-compiling an Debian conclude successfully. 👍

a-michelis:
I'll be branching out from this branch to start a reiteration over WinUSB -like what i did with this PR- , if that's OK. 
I try to cram as much throughput as possible now that i have a little bit of time.

slyshykO (approved):

